PR #5, opened 2018-06-26
Hide address-based option in social errors

nheagy:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/8388

simulator screen shot - iphone x - 2018-06-26 at 12 12 09

When logging in with `restrictToWPCom` turned on, and the uses selects to login via Google, we now hide the option to login via address when there is an error. This will improve error recovery for these users.

# To test:
- Login to the app using a self-hosted site
- On the Me tab, tap to login to wpcom
- Chose to login via Google
- Force an error (I modify `googleLoginTapped()` in `LoginEmailViewController`)
- Ensure that the option to retry via site address is gone, and that the other two options still work
- Ensure that the three options all appear and work when login is not restricted to wpcom


nheagy:
Thanks @jleandroperez!

PR #6, opened 2018-07-03
Fix asset reference

loremattei:
This PR updates an misaligned asset reference. 

See issue #9692 on WordPress-iOS repo.


aerych:
Heya @loremattei ! 
Looks good. To test I double checked the asset name, then tested this commit as the pod source in the main app against `release/10.3`.  I was able to view the fancy alert view for the site address help screen and I see the image as expected. 
:shipit: !

PR #9, opened 2018-07-04
Backporting PR's merged against develop

jleandroperez:
### Details:
We had several PR's mapped against `master` instead of `develop` (which should have never happened).

Let's... fix this, and never speak of it, ever again.

cc @loremattei


jleandroperez:
Thank you sir!

PR #7, opened 2018-07-03
NUXButton configurable using UIAppearance

mindgraffiti:
This was branched from `new-woo-apis`.

⚠️ The secondary button's default configurations have not been tested against the original button designs for WPiOS. Since it's branched from `new-woo-apis` it isn't compatible with WPiOS just yet and was breaking when I attempted to test the secondary button.

This PR makes the NUXButton configurable using UIAppearance. To run a full test, check out the WCiOS PR instructions.

PR #10, opened 2018-07-05
NUXButton configurable using UIAppearance

jleandroperez:
Resubmitting #7 

Not sure how the former one got closed (apparently, happened automatically after i've deleted `new-woo-apis`).

I'm pasting, below, the former PR's description.

---

```
⚠️ The secondary button's default configurations have not been tested against the original button designs for WPiOS. Since it's branched from `new-woo-apis` it isn't compatible with WPiOS just yet and was breaking when I attempted to test the secondary button.

This PR makes the NUXButton configurable using UIAppearance. To run a full test, check out the WCiOS PR instructions. 
```


WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,285 @@
...
+        super.layoutSubviews()
+
+        if activityIndicator.isAnimating {
+            titleLabel?.frame = CGRect.zero
 
-    /// Configure the appearance of the button.
+            var frm = activityIndicator.frame
+            frm.origin.x = (frame.width - frm.width) / 2.0
+            frm.origin.y = (frame.height - frm.height) / 2.0
+            activityIndicator.frame = frm
```
jleandroperez:
```
activityIndicator.frame = frm.integral
```

**Reason:** floating positions don't really exist (the screen requires integer coordinates). Anything that has a comma value results in a blurry view.

cc @mindgraffiti

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,285 @@
...
+        static let highlightBorderColor = normalBorderColor
+    }
+
+    /// Style: Secondary
+    ///
+    struct Secondary {
+        static let normalBackgroundColor = UIColor.white
+        static let normalBorderColor = WPStyleGuide.greyLighten20()
+        static let highlightBackgroundColor = WPStyleGuide.greyLighten20()
+        static let highlightBorderColor = highlightBackgroundColor
```
jleandroperez: @mindgraffiti shouldn't all of the values here come from the new *Style* struct? (Primary.*Color / Secondary.*Color)

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,285 @@
...
+        static let backgroundColor = UIColor.white
+        static let borderColor = WPStyleGuide.greyLighten30()
+    }
+
+    /// Style: Title
+    ///
+    struct Title {
+        static let primaryColor = UIColor.white
+        static let secondaryColor = WPStyleGuide.darkGrey()
+        static let disabledColor = WPStyleGuide.greyLighten30()
```
jleandroperez: Same comment as above: Shouldn't these come from the new Style struct?

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -165,6 +164,32 @@ public struct WordPressAuthenticatorConfiguration {
...
+// MARK: - WordPress Authenticator Styles
+//
+public struct WordPressAuthenticatorStyle {
+    /// Alternative logins link color
+    ///
+    let linkColor: UIColor
+
+    /// Alternative logins highlight color
+    ///
+    let highlightColor: UIColor
```
jleandroperez: Please set `linkColor` and `highlightColor` as *public*, otherwise they won't be accessible from other frameworks.

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -165,6 +164,32 @@ public struct WordPressAuthenticatorConfiguration {
...
+    /// Designated initializer
+    ///
+    public init(linkColor: UIColor, highlightColor: UIColor) {
+        self.linkColor = linkColor
+        self.highlightColor = highlightColor
+    }
+}
+
+public extension WordPressAuthenticatorStyle {
+    static var defaultStyle: WordPressAuthenticatorStyle {
```
jleandroperez: @mindgraffiti please set to *public*

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -210,18 +239,19 @@ public struct WordPressAuthenticatorConfiguration {
...
-    private init(configuration: WordPressAuthenticatorConfiguration) {
+    private init(configuration: WordPressAuthenticatorConfiguration, style: WordPressAuthenticatorStyle) {
         self.configuration = configuration
+        self.style = style
     }
 
     /// Initializes the WordPressAuthenticator with the specified Configuration.
     ///
-    public static func initialize(configuration: WordPressAuthenticatorConfiguration) {
+    public static func initialize(configuration: WordPressAuthenticatorConfiguration, style: WordPressAuthenticatorStyle) {
```
jleandroperez: Can we have `style: WordPressAuthenticatorStyle  = .defaultStyle` ?

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,285 @@
 import UIKit
 import WordPressShared
+import WordPressUI
+
+/// A protocol for an element that can display a UIActivityIndicatorView
+@objc public protocol ActivityIndicatorButton {
```
jleandroperez: @mindgraffiti since the protocol `ActivityIndicatorButton` isn't really used anywhere (there are no ivars / vars anywhere with that specific Type), can we nuke it?

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,285 @@
...
 
 /// A stylized button used by Login controllers. It also can display a `UIActivityIndicatorView`.
-@objc open class NUXButton: NUXSubmitButton {
-    // MARK: - Configuration
-    fileprivate let horizontalInset: CGFloat = 20
-    fileprivate let verticalInset: CGFloat = 12
-    fileprivate let maxFontSize: CGFloat = 22
+@objc open class NUXButton: UIButton, ActivityIndicatorButton {
+    @objc var isAnimating: Bool {
+        get {
```
jleandroperez: @mindgraffiti you can drop the `get  { }` block. If a property is readOnly, you can just do something like this:

```
var isAnimating: Bool {
   return activityIndicator.isAnimating
}
```

Podfile.lock
```diff
@@ -1,20 +1,5 @@
 PODS:
   - 1PasswordExtension (1.8.5)
-  - AFNetworking (3.2.1):
```
nheagy: Any idea why the Podfile changes so much without Podfile changes?

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,285 @@
...
+
+/// A protocol for an element that can display a UIActivityIndicatorView
+@objc public protocol ActivityIndicatorButton {
+    func showActivityIndicator(_ show: Bool)
+}
 
 /// A stylized button used by Login controllers. It also can display a `UIActivityIndicatorView`.
-@objc open class NUXButton: NUXSubmitButton {
-    // MARK: - Configuration
-    fileprivate let horizontalInset: CGFloat = 20
```
nheagy: Where'd these config values go?

Podfile.lock
mindgraffiti: Yes, that's due to all the mucking about I did with making some of the pods local. I think I need to pull the original `Podfile` and `Podfile.lock` and overwrite the changes made in my PR.

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: They moved to here: https://github.com/wordpress-mobile/WordPressUI-iOS/blob/develop/WordPressUI/Extensions/UIImage%2BAssets.swift#L87

The NUXButton was using the same sizing and configurations as `FancyButton`, so I moved the definitions out of `FancyButton` and into the `UIImage+Assets` extension. This PR has more details https://github.com/wordpress-mobile/WordPressUI-iOS/pull/8

mindgraffiti:
> I think we should holdoff on merging this until the WPiOS implementation is available. Let me know if you think otherwise.

Agreed! I want to thoroughly test this now that the `new-woo-apis` branch has been integrated. Was looking for some initial impressions on the work so far, to make sure I'm on the right track.

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: This is confusing. `x`, `y`, and `frame` all accept `CGFloat` values. Why do they make it like that if they really want `int`s?

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
mindgraffiti: ✔️

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
mindgraffiti: ✔️

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
jleandroperez: @mindgraffiti actually, my explanation was wrong. Lemme rephrase: positions are expressed in terms of Points, not Pixels. The problem is:

- **Non Retina Devices:** PositionX = 0.5 maps to `half a pixel in the X axis`
- **@2x Devices:** PositionX = 0.5 maps to `one pixel`
- **@3x Devices:** Position X = 0.5 maps to `one pixel and a half
 
Any "Pixel Position" that's not integral, when rendered, gets antialias applied. Which is why you usually wanna avoid non integral positions. (Exception is: when you're actually calculating things, so it maps to integral pixels).

[A little bit more about this here](https://stackoverflow.com/questions/14190807/why-are-uiview-frames-made-up-of-floats).

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: ✔️

Podfile.lock
jleandroperez: We've been working on nuking AFNetworking (with Sergio) during last Hack Week, from WordPressKit. This change is just (or should be!!) a side effect of running `pod install`.

No need to bring back the old Podfile.lock, this is a *beautiful* thing actually!!! (one less dependency).

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -6,165 +6,6 @@ import WordPressShared
...
-        self.wpcomScheme = wpcomScheme
-        self.wpcomTermsOfServiceURL = wpcomTermsOfServiceURL
-        self.googleLoginClientId =  googleLoginClientId
-        self.googleLoginServerClientId = googleLoginServerClientId
-        self.googleLoginScheme = googleLoginScheme
-        self.userAgent = userAgent
-    }
-}
-
-
```
jleandroperez: ❤️ this change!!

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,157 @@
...
 
-        let disabledImage = UIImage.beveledDisabledButtonImage
-        let titleColorDisabled = WPStyleGuide.greyLighten30()
-
         setBackgroundImage(normalImage, for: .normal)
-        setBackgroundImage(highlightImage, for: .highlighted)
+        setBackgroundImage(highlightedImage, for: .highlighted)
         setBackgroundImage(disabledImage, for: .disabled)
 
+        activityIndicator.activityIndicatorViewStyle = .gray
```
jleandroperez: Perhaps this should go into `configureActivityIndicator()` (since it's not really a Background?)

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,157 @@
...
+        super.layoutSubviews()
+
+        if activityIndicator.isAnimating {
+            titleLabel?.frame = CGRect.zero
+
+            var frm = activityIndicator.frame
+            frm.origin.x = (frame.width - frm.width) / 2.0
+            frm.origin.y = (frame.height - frm.height) / 2.0
+            activityIndicator.frame = frm.integral
+        }
```
jleandroperez: Looks good!! but now that i realize about it: why not just Autolayout? (`centerXAnchor` / `centerYAnchor`)

WordPressAuthenticator/Extensions/WPStyleGuide+Login.swift
```diff
@@ -7,6 +7,7 @@ final class SubheadlineButton: UIButton {
         super.traitCollectionDidChange(previousTraitCollection)
         if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
             titleLabel?.font = WPStyleGuide.mediumWeightFont(forStyle: .subheadline)
+            titleLabel?.textColor = WPStyleGuide.wordPressBlue()
```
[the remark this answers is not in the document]
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: I'm drawing the line here. Need to get this PR closed. :)

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,149 @@
...
-        contentEdgeInsets = UIEdgeInsets(top: verticalInset, left: horizontalInset, bottom: verticalInset, right: horizontalInset)
+    open func showActivityIndicator(_ show: Bool) {
+        if show {
+            activityIndicator.startAnimating()
+        } else {
+            activityIndicator.stopAnimating()
+        }
+        configureBackgrounds()
+        configureTitleColors()
+        activityIndicator.activityIndicatorViewStyle = .gray
```
jleandroperez: You probably wanna move these three lines (configureBackgrounds, configureTitleColors and configureActivityIndicator) elsewhere. Perhaps  to  `awakeFromNib`.

Reason: Each time you call this, *configureBackgrounds*  will re-add the activityIndicator as a subview. This should probably print a warning or so, but it's something we should run just once.

Podfile.lock
nheagy: Ah, got it. 👍

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,149 @@
...
+    }
+
+    open override func awakeFromNib() {
+        super.awakeFromNib()
+        configureAppearance()
+    }
 
+    /// Setup: shorter reference for style
+    ///
+    private let style = WordPressAuthenticator.shared.style
```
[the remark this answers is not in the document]
mindgraffiti: Error: `Static var 'shared' is not a member type of 'WordPressAuthenticator'`

WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -1,54 +1,149 @@
...
+extension NUXButton {
+    override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
+            didChangePreferredContentSize()
+        }
     }
 
-    override open func configureBorderColor() {
+    struct Metrics {
```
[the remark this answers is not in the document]
mindgraffiti: ✔️

WordPressAuthenticator/NUX/NUXButton.swift
mindgraffiti: ✔️

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
nheagy: Same

WordPressAuthenticator/NUX/NUXButton.swift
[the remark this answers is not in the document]
jleandroperez: Ahhhhhhhh right, it's not a type!!!. My apologies!!. Only alternative is having this as a calculated property. But "meh", really no difference in there.

Sorry about the fuzz!!

PR #36, opened 2018-12-04
Add the ability to override text displayed to the user

### LINKED ISSUE #35: Expose login text property
https://github.com/woocommerce/woocommerce-ios/issues/472

It currently says:

- [ ] expose the property
- [ ] keep the WordPress string as the default
- [ ] confirm changes do not interfere with WordPress
- [ ] confirm changes allow overrides in WooCommerce

mindgraffiti:
Fixes #35. Ref. https://github.com/woocommerce/woocommerce-ios/issues/472.

**Background**
Some of the login text that displays to the user has instructions that are WordPress focused. It can be confusing to a user when the host app is WooCommerce instead. 

**Summary**
In this PR, we're exposing the instructions label and allowing the text to be overridden on init. The default display text remains the same for WordPress users. The `WordPressAuthenticatorDisplayText` struct is introduced as a way to manage any text that displays directly to the user. 

Better names welcomed! I tried: UserMessages, UIText, and TextOverrides. None of them quite fit.

**To Test (Pick one) - WCiOS Option**
1. Check out this PR and run it: https://github.com/woocommerce/woocommerce-ios/pull/487#event-2005399244

**To Test - WPiOS Option**
1. Using WPiOS, point the podfile to this branch: `pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS', :branch => 'feature/35-expose-login-text'`
2. `pod install` followed by `pod update`
3. Run the app and go through the email login flow, with an account that has Jetpack. The instructions should remain unchanged and say ```Log in to the WordPress.com account you used to connect Jetpack.``` 
4. Go to the site login flow. the instructions should remain unchanged and say ```Enter the address of your WordPress site you'd like to connect.``` 

Screenshots of WPiOS after installing this pod update:

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -59,7 +59,9 @@ import WordPressUI
 
     /// Designated Initializer
     ///
-    private init(configuration: WordPressAuthenticatorConfiguration, style: WordPressAuthenticatorStyle, text: WordPressAuthenticatorDisplayText) {
+    private init(configuration: WordPressAuthenticatorConfiguration,
+                 style: WordPressAuthenticatorStyle,
+                 text: WordPressAuthenticatorDisplayText) {
         self.configuration = configuration
         self.style = style
         self.text = text
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -38,7 +38,7 @@ import WordPressUI
 
     /// Authenticator's Display Texts.
     ///
-    public let text: WordPressAuthenticatorDisplayText
+    public let displayStrings: WordPressAuthenticatorDisplayStrings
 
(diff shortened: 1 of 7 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -55,19 +59,24 @@ import WordPressUI
 
     /// Designated Initializer
     ///
-    private init(configuration: WordPressAuthenticatorConfiguration, style: WordPressAuthenticatorStyle) {
+    private init(configuration: WordPressAuthenticatorConfiguration,
+                 style: WordPressAuthenticatorStyle,
+                 text: WordPressAuthenticatorDisplayText) {
```
jleandroperez: Nipticky comment here!!

I'm not sure if `text` is the best property name. Although its uncountable, feels like a "single string" for a "single spot". How about....

- **displayStrings**
- **localizedStrings**

This one caught my eye, mostly, because UILabel's `.text` property!

mindgraffiti:
Thank you!

PR #42, opened 2019-01-11
Issues/9618 login with username

aerych:
Refs: https://github.com/wordpress-mobile/WordPress-iOS/issues/9618

This PR adds custom handling for the `email_login_not_allowed` error that is sometimes occurs when logging into WordPress.com when a user's email address meets the criteria for being "suspicious".  In these cases we take the user to the LoginSelfHostedViewController, but configured to log into WordPress.com as the site. This allows the user to proceed with their username/password credentials rather than being stuck on the email address screen without a clear path forward. 

With these changes we:
- check for the error in question
- assign the error to be displayed
- show the LoginSelfHostedViewController via a newly added segue
- show the error after presenting the new view controller to provide context for what to do next

Example at C02Q9GACF/p1547220737194300-slack-ios-osx-dev

Pinging both @mindgraffiti and @nheagy for sanity checks since this is a shared library.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.1.8"
+  s.version       = "1.1.8-beta.1"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.1.7"
+  s.version       = "1.1.8"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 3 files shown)


aerych:
Thanks @nheagy and @mindgraffiti !!

hypest:
👋, I'm trying to investigate and fix [the issue](https://github.com/wordpress-mobile/WordPress-Android/issues/7922) for the Android side and I'm having trouble convincing the server to send the `email_login_not_allowed` error type.

I wonder, were you able to verify that the server does indeed send this error type while developing this PR? Any steps to recreate the situation?

What I've tried so far was to make the backend (in my sandbox) fail the `suspicious_email_login()` call but the response I get on the client (WPAndroid) is this one:
```
{
	"error": "invalid_request",
	"error_description": "Please log in using your WordPress.com username instead of your email address."
}
```

Notice that the error is `invalid_request`, not `email_login_not_allowed`.

Any clues appreciated :) cc @aerych

hypest:
I tried out WPiOS (v12.0) against my sandbox and verified that if I rig the sandbox to always fail suspicious_email_login() (returning true) then [the same response](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/42#issuecomment-481664749) comes through Charles (Proxy) to the app. So, at this point, I'm not sure whether the issue is really fixed by this PR 🤔.

mindgraffiti:
The wp.com changes have taken place: p1555065019002100-slack-C04B0MQUM. (Thank you @hypest!)

Can confirm, on the WCiOS side, I can receive the error: `email_login_not_allowed` and the authenticator reacts appropriately. Demo:

PR #44, opened 2019-01-16
Add automated testing

jkmassel:
Uses CircleCI to validate that everything is working correctly.

Also fixes issues causing the builds to fail

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,9 @@ jobs:
           name: Bundle install
           command: bundle install --path=vendor/bundle
       - run:
+          name: Fetch CocoaPods Specs
+          command: curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
+      - run:
           name: Pod Install
           command: bundle exec pod install
           environment:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+      xcode: "10.1.0"
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - wordpress-authenticator-ios-{{ checksum "Gemfile.lock" }}-{{ checksum "Podfile.lock" }}
+      - run:
+          name: Bundle install
+          command: bundle install --path=vendor/bundle
+      - run:
```
jtreanor: CircleCI recommend using this step prior to running `pod install` (notice that `pod install` takes 8 minutes currently):

```
- run:
          name: Fetch CocoaPods Specs
          command: |
            curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
```

As we expand more across CircleCI I think it may may sense to share steps using [CircleCI Orbs](https://circleci.com/orbs/) so we don't need to duplicate things like this.

PR #48, opened 2019-01-25
CircleCI: Simplify config by using shared Orbs for testing the project and validating the podspec

jtreanor:
This updates the CircleCI Config to use the Orbs defined in https://github.com/wordpress-mobile/circleci-orbs. It now also tests the project as well as validating the podspec.

We will be able to re-use this simple config across the other library projects in `wordpress-mobile`.

Related to https://github.com/wordpress-mobile/WordPress-iOS/pull/10892.

It also adds `cocoapods-repo-update` to the `Podfile` to ensure that specs are always up to date.


jtreanor:
@jkmassel Good point! I'll change that 😄

jklausa:
What's the difference between `lib lint` an `spec lint`?

jkmassel:
`lib lint` checks to ensure that the library is valid on the local machine.

`spec lint` downloads the most recent pod version it can find and validates _that_.

What that means is that if you do `spec lint`, you're not testing the version you're working on, you're testing the last release. Which in this case, isn't what you want :)

jklausa:
Thanks for the explanation!

PR #49, opened 2019-01-29
Updated podspec and Podfile to reflect WordPressKit revision

stevebaranski:
### Description
Propagates the fix for [this issue](
https://github.com/wordpress-mobile/WordPressKit-iOS/pull/80) in `WordPressKit` to `WordPressAuthenticator`, en route to WPiOS.

### Testing
Checkout the branch and confirm that existing tests pass.

PR #50, opened 2019-01-29
Corrected podspec version number

stevebaranski:
In the [preceding PR](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/49/), I omitted the beta suffix from the version in the podspec

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.1.9-beta"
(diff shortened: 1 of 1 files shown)


stevebaranski:
The CI error observed is as follows:
```
None of your spec sources contain a spec satisfying the dependency: `WordPressKit (~> 2.0-beta)`.
```

This version has just been pushed to trunk. Unfortunately, however, it doesn't look the CocoaPods repo is being updated. Perhaps it is related to [this change](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/0f34cd4ce9dd7b4b6e4408e57b7b1a514069982b#diff-1d37e48f9ceff6d8030570cd36286a61)?

PR #51, opened 2019-02-04
CircleCI: Update specs repo before validating podspec

jtreanor:
This is a small CircleCI config update that ensures the specs repo is up to date before validating the podspec. Without this change, CI could fail to find recently updated pods.

@stevebaranski This fixes the failure you encountered in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/50

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,5 @@
 source 'https://rubygems.org' do
   gem 'cocoapods', '1.5.3'
-  gem 'cocoapods-check'
   gem 'cocoapods-repo-update'
   gem 'xcpretty'
 end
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,6 +1,5 @@
 source 'https://rubygems.org' do
   gem 'cocoapods', '1.5.3'
-  gem 'cocoapods-check'
```
jtreanor: This is removed because its not used by the latest Orb.

stevebaranski:
Thank you @jtreanor!

PR #61, opened 2019-03-15
Merge back to develop changes that are in master

aerych:
It looks like a release was merged to master before it was merged to develop.
This PR merges those changes back to develop.

I've opened this one just from the GitHub UI so pinging both @SergioEstevao and @loremattei in case there is anything wonky with the way the web UI does things. 😬


aerych:
Thanks @SergioEstevao !

loremattei:
Thank you for taking care of this @aerych!

PR #67, opened 2019-03-27
Fix compatibility with recent versions of CocoaPods (> 1.5.3)

jtreanor:
This change fixes compatibility with recent versions of CocoaPods. Once merged it will be possible to update CocoaPods in all of our projects.

The issue we are seeing is caused by [`GoogleSignInRepacked`](https://github.com/wordpress-mobile/GoogleSignInRepacked). It seems that it is not behaving as a true dynamic framework and some linking changes by CocoaPods mean it doesn't work anymore. The issue is described [here](https://github.com/CocoaPods/CocoaPods/issues/8612#issuecomment-477219241).

The solution is to use the official `GoogleSignIn` pod and add `s.static_framework = true` to the podspec.

Changing to a static framework means that everything gets added to the main bundle. This can lead to conflicts once it is installed in an app. I have made a few changes to mitigate this:

- The podspec now uses `resource_bundles` to create `WordPressAuthenticatorResources.bundle` which contains the `.xcassets` file and the `json` animation files. Storyboards and xibs are still loaded from the main bundle.
- This meant that usages of `Bundle(for: WordPressAuthenticator.self)` needed to be updated for this new bundle.
- The storyboards also needed to be updated to not look for a module named `WordPress` for classes. This was likely left over from extracting the storyboards from WordPress-iOS.

To test:

Since the risk here is that resources won't be correctly loaded, its probably best to test this from within WordPress:

1. Check out this branch. Check out `develop` of WordPress-iOS.
2. Uncomment the line in your WordPress-iOS `Podfile` that says `pod 'WordPressAuthenticator', :path => '../WordPressAuthenticator-iOS'` and comment the line above.
3. Run `rake dependencies`
4. Build and run the app.
5. Run through a login flow and make sure nothing looks out of place (missing icons, animations etc.).

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -18,8 +18,17 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git", :tag => s.version.to_s }
   s.source_files  = 'WordPressAuthenticator/**/*.{h,m,swift}'
   s.private_header_files = "WordPressAuthenticator/Private/*.h"
-  s.resources     = [ 'WordPressAuthenticator/**/*.{xcassets,storyboard,xib,json}' ]
+  s.resource_bundles = {
+    'WordPressAuthenticatorResources': [
+      'WordPressAuthenticator/Resources/Assets.xcassets',
+      'WordPressAuthenticator/Resources/Animations/*.json'
+    ]
+  }
+  s.resources     = [
+    'WordPressAuthenticator/**/*.{storyboard,xib}'
+  ]
   s.requires_arc  = true
+  s.static_framework = true # This is needed because GoogleSignIn vendors a static framework
   s.header_dir    = 'WordPressAuthenticator'
 
   s.pod_target_xcconfig = { 'ENABLE_BITCODE' => 'NO' }
@@ -32,7 +41,7 @@ Pod::Spec.new do |s|
   s.dependency 'SVProgressHUD', '2.2.5'
 
   s.dependency 'Gridicons', '~> 0.15'
-  s.dependency 'GoogleSignInRepacked', '4.1.2'
+  s.dependency 'GoogleSignIn', '4.1.2'
   s.dependency 'WordPressUI', '~> 1.0'
   s.dependency 'WordPressKit', '~> 3.1'
   s.dependency 'WordPressShared', '~> 1.4'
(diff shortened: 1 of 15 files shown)


WordPressAuthenticator.podspec
```diff
@@ -32,7 +41,7 @@ Pod::Spec.new do |s|
   s.dependency 'SVProgressHUD', '2.2.5'
 
   s.dependency 'Gridicons', '~> 0.15'
-  s.dependency 'GoogleSignInRepacked', '4.1.2'
+  s.dependency 'GoogleSignIn', '4.1.2'
```
jtreanor: We can probably update this too since this is an old version but better to do one thing at a time 😄

jtreanor:
It's worth pointing out what the alternative solutions would be here:

1. Fix the issue in CocoaPods. While we could certainly do this and see if it would be accepted. The new behavior is arguably correct since `GoogleSignInRepacked` doesn't behave as a true dynamic framework.
2. Fix  `GoogleSignInRepacked`. We could take another look t how we repackage `GoogleSignIn`. There is another project that does this and it works with CP 1.6.0: https://github.com/crspybits/SMGoogleSignIn. We could use a similar approach. I think it's better to get rid of the repackaging step altogether though.

astralbodies:
I'd also test to make sure packaging a release works properly and passes the App Store tests. I remember something vaguely when @jleandroperez was working on this solution regarding that.

jleandroperez:
I can't recall exactly in which spot things literally exploded (and forced us to repack as a dynamic framework).

Archiving a binary (and doing a test upload to the AppStore), along with running the development binary directly in the device should suffice. As far as i remember the issue itself was a CocoaPods limitation. Since the Repacked framework was built... Mayb 2018?, perhaps this CocoaPods issue is already solved.

Please feel free to invoke me anytime, i'd be more than happy to help!!! YAY Opensource!!

jtreanor:
Good suggestion @astralbodies. I'll try that.

Thanks for the insight @jleandroperez! The CocoaPods limitation that transitive dependencies can't be static still exists, but the newish `s.static_framework = true` feature seems to allow this.

I'll dig and update here with what I find out.

jleandroperez:
Awesome!! Getting rid of the repacked binary would definitely be a good move. Too much blood went into that repack script!!!

jtreanor:
Ok, so I have built an AppStore archive of WPiOS that points to this branch and used "Validate App" from Organiser. It passed all checks:

image

I have also tested a debug build WPiOS on a physical device, including doing a full login flow and everything looks good.

jtreanor:
Thanks @nheagy!

jleandroperez:
@jtreanor I think `pod install` would throw the error right away. Glad to see this fix go thru, thanks everyone!!

:hug:

PR #71, opened 2019-04-03
WCiOS: Support Site URL login

mindgraffiti:
This PR allows WooCommerce iOS users start their login with a Site URL, then move to a WP.com username / password screen if the site is verified. Ref. https://github.com/woocommerce/woocommerce-ios/issues/263

Testing instructions WCiOS as the host app: https://github.com/woocommerce/woocommerce-ios/pull/835
Testing instructions WPiOS as the host app: https://github.com/wordpress-mobile/WordPress-iOS/pull/11407


WordPressAuthenticator/Credentials/WordPressCredentials.swift
```diff
@@ -9,5 +9,5 @@ public enum WordPressCredentials {
 
     /// WordPress.com Site Credentials.
     ///
-    case wpcom(authToken: String, isJetpackLogin: Bool, multifactor: Bool)
+    case wpcom(authToken: String, isJetpackLogin: Bool, multifactor: Bool, siteURL: String)
```
aerych: It pains me that we can assign default values to enums.  

It might be nice to have the default `dotcomAddress` declared here vs in `LoginFieldsMeta` since its a credentials thing, and it might be nice if we could mimic the behavior of default values with an internal builder class.   It would look something like (just psuedo code)
```
public enum WordPressCredentials {
    public struct Builder {
         public wpcom(authToken: String, isJetpackLogin: Bool, multifactor: Bool, siteURL: String?) {
                let site = siteURL ?? "http://wordpress.com"
                return WordPressCredentials.wpcom(authToken, isJetpackLogin, multifactor, site)
          }
    }
   [...]
}
```
And then wherever wpcom credentials were used: 
```
WordPressCredentials.Builder.wpcom(...) 
```

Just a thought.  This or something like it could save all the different places where the site is being checked before deciding whether to use the default or not.  Up to you!

Podfile.lock
```diff
@@ -54,7 +54,7 @@ PODS:
   - Specta (1.0.7)
   - SVProgressHUD (2.2.5)
   - UIDeviceIdentifier (1.1.4)
-  - WordPressKit (3.1.1):
+  - WordPressKit (3.2.1):
```
yaelirub: When I do pod update after checking the branch I see a warning:
[!] GTMOAuth2 has been deprecated in favor of GTMAppAuth
Is that expected?

WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
```diff
@@ -185,32 +188,78 @@ class LoginSiteAddressViewController: LoginViewController, NUXKeyboardResponder
         let successBlock: (WordPressComSiteInfo) -> Void = { [weak self] siteInfo in
             self?.loginFields.meta.siteInfo = siteInfo
             if WordPressAuthenticator.shared.delegate?.allowWPComLogin == false {
+                // Hey, you have to log out of your existing WP.com account before logging into another one.
                 self?.promptUserToLogoutBeforeConnectingWPComSite()
                 self?.configureViewLoading(false)
             } else {
-                self?.showSelfHostedUsernamePassword()
+                self?.configureViewLoading(false)
+                guard let strongSelf = self else {
```
yaelirub: We can do `guard let self = self` in Swift 4.2

WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
```diff
@@ -185,32 +188,78 @@ class LoginSiteAddressViewController: LoginViewController, NUXKeyboardResponder
...
+                }
+
+                strongSelf.presentNextControllerIfPossible(siteInfo: nil)
+            })
         }
     }
 
+    func presentNextControllerIfPossible(siteInfo: WordPressComSiteInfo?) {
+        WordPressAuthenticator.shared.delegate?.shouldPresentUsernamePasswordController(for: siteInfo, onCompletion: { (error, isSelfHosted) in
+            guard let originalError = error as NSError? else {
```
yaelirub: I don't think the casting to NSError here is needed.

WordPressAuthenticator/Signin/LoginUsernamePasswordViewController.swift
```diff
@@ -0,0 +1,313 @@
+import UIKit
+import WordPressShared
+
+/// Part two of the self-hosted sign in flow. A valid site address should be acquired
+/// before presenting this view controller.
+///
+class LoginUsernamePasswordViewController: LoginViewController, NUXKeyboardResponder {
```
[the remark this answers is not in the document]
mindgraffiti: So the `LoginUsernamePasswordViewController` is a copy of `LoginSelfHostedViewController`. If you open up the `Login.storyboard`, you'll see that it has some minor modifications compared to the self-hosted VC. The purpose of the `LoginUsernamePasswordViewController` is to use it for a WooCommerce flow. We are currently doing Sign-In research, and this flow will probably get deleted after another sprint or two. This was the best way I could think of to create the custom WooCommerce flow, not disrupt WordPress flows, and make it easy to delete / remove once the new research is finished and a new sign-in flow is implemented for Woo.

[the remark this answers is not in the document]
mindgraffiti: I must have un-done the .proj file at some point. Re-added the `LoginUsernamePasswordViewController` to the .proj file.

Podfile.lock
mindgraffiti: That's a great question! Yes that's supposed to happen. @jtreanor has been doing extensive work to get Authenticator up-to-date so it's compatible with CocoaPods version 1.6.1. So now we are free to update the `GoogleSignIn` pod (which is where that warning is originating from). I am not going to do that in this PR, but I can file a task so the warning will not get lost.

WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
mindgraffiti: Nice! I didn't know that.

WordPressAuthenticator/Signin/LoginUsernamePasswordViewController.swift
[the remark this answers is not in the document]
aerych: > and make it easy to delete / remove once the new research is finished and a new sign-in flow is implemented for Woo

Sounds good!  If we decide to keep it as a variant we should probably revisit the implementation so its not a copy/paste. Our future-selves can deal with that.

mindgraffiti:
Ready for a second look.

aerych:
I took another peek with the latest changes and after testing in WPiOS and WCiOS its looking good to me. :shipit:

mindgraffiti:
Thank you @aerych and @yaelirub!

Issue #38, opened 2018-12-11
A console warning is observed regarding a referenced image asset


stevebaranski:
When using a `WordPressAuthenticator` component in WordPress-iOS, I'm seeing the following warning in the console:
```
2018-12-11 09:21:31.650079-0800 WordPress[91387:1155878] Could not load the "beveled-blue-button-down" image referenced from a nib in the bundle with identifier "org.cocoapods.WordPressAuthenticator"
```

stevebaranski:
I'm afraid that this is still an issue in/on WPiOS. I am able to reproduce when attempting to login; navigating to the below screen evinces the aforementioned console log entry.
<img width="584" alt="wpa-38" src="https://user-images.githubusercontent.com/221062/55815854-ec2d3480-5aa5-11e9-9120-4eed26be7bd7.png">

I believe the image is missing from `WordPressAuthenticator`. I see it referenced in [this storyboard](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/d665345e2b13181391e8399274db14d9680d2c20/WordPressAuthenticator/NUX/NUXButtonView.storyboard#L111), but I don't see it within the module.

It appears that this image is located within WPiOS itself, referenced in both [`AppImages.xcassets`](https://github.com/wordpress-mobile/WordPress-iOS/tree/4ac54edd8c4db2978a0122f97e137c755e627d4a/WordPress/Resources/AppImages.xcassets/beveled-blue-button-down.imageset) and in [`LoginEpilogue.storyboard`](https://github.com/wordpress-mobile/WordPress-iOS/blob/4ac54edd8c4db2978a0122f97e137c755e627d4a/WordPress/Classes/ViewRelated/NUX/LoginEpilogue.storyboard#L187).

If I recall, `WordPressAuthenticator` was extracted into a standalone framework from WPiOS. Perhaps this image just wasn't ported at that time? 

Given that it's still used in WPiOS, perhaps the simplest solution is to copy it from the WPiOS asset catalog to [this one](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/tree/develop/WordPressAuthenticator/Resources/Assets.xcassets) in this framework?

jtreanor:
Thanks for re-opening @stevebaranski!

[...]

PR #72, opened 2019-04-05
Add storyboards and xibs to the resource_bundle

jtreanor:
In https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/67, as part of converting to a static framework, I moved images and assets to `WordPressAuthenticatorResources.bundle`. `.storyboard` and `.xib` files remained in the main bundle.

This mostly worked fine for WordPress-iOS, but has proved to be problematic in some cases there and in all cases for WooCommerce-iOS. Image references in storyboards were not resoving, with errors like this:

```
2018-12-11 09:21:31.650079-0800 WordPress[91387:1155878] Could not load the "beveled-blue-button-down" image referenced from a nib in the bundle with identifier "org.cocoapods.WordPressAuthenticator"
``` 

Fixes https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/38.

The solution I have implemented is as follows:

- Move `.xib` and `.storyboard` files into the `resource_bundle`. This means images will be loaded from the same bundle and it will find them correctly.
- Rename `WordPressAuthenticatorResources.bundle` to `WordPressAuthenticator.bundle`. This is a surprising quirk of CocoaPods to ensure they are all in the same target. More info [here](https://muyexi.im/switch-resource-to-resource_bundle/).
- Related code changes to load the right bundle. It's simpler now.

To test:

- Point WordPress-iOS to this branch (I have made a branch [here](https://github.com/wordpress-mobile/WordPress-iOS/tree/try/wpauthenticator-resources) to save some effort).
    - Smoke test the login flow, making sure images are loaded correctly.
- Point WooCommerce-iOS to this branch (I have made a branch [here](https://github.com/woocommerce/woocommerce-ios/tree/try/wpauthenticator-resources) to save some effort).
    - Smoke test the login flow, making sure images are loaded correctly.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -19,14 +19,12 @@ Pod::Spec.new do |s|
   s.source_files  = 'WordPressAuthenticator/**/*.{h,m,swift}'
   s.private_header_files = "WordPressAuthenticator/Private/*.h"
   s.resource_bundles = {
-    'WordPressAuthenticatorResources': [
+    'WordPressAuthenticator': [
       'WordPressAuthenticator/Resources/Assets.xcassets',
-      'WordPressAuthenticator/Resources/Animations/*.json'
+      'WordPressAuthenticator/Resources/Animations/*.json',
+      'WordPressAuthenticator/**/*.{storyboard,xib}'
     ]
   }
-  s.resources     = [
-    'WordPressAuthenticator/**/*.{storyboard,xib}'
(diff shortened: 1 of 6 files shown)


jtreanor:
That's great, thanks @mindgraffiti!

 I have tested WordPress-iOS quite thoroughly with this and I'm satisfied that all is well now so I'm going to go ahead with the merge.

PR #78, opened 2019-04-11
Make WordPressAuthenticator.bundle public

jtreanor:
This makes the `bundle` var in `WordPressAuthenticator` public so that the bundle can be accessed from outside the framework.

This will allow https://github.com/wordpress-mobile/WordPress-iOS/issues/11444 to be fixed.


jtreanor:
I'm going to cherry pick this into a 1.3.1 so only these changes are included in WPiOS.

PR #85, opened 2019-04-23
Updating WPKit pod version.

ScoutHarris:
This updates WordPressKit to the latest version.

To test:
Build and test. Verify tests pass.

PR #87, opened 2019-04-23
Bumping WPAuth pod version.

ScoutHarris:
This updates the pod version (that should have been done with #85).


ScoutHarris:
Thanks @stevebaranski !

PR #89, opened 2019-05-01
Compatibility with both CocoaPods dynamic and static frameworks

jtreanor:
This makes the changes needed for the WordPressAuthenticator pod to be installed as a static or dynamic framework (with or without `use_frameworks!`). Currently, it does not work correctly when installed statically. Even though `static_framework` is set to true, these changes are needed to ensure it works without `use_frameworks!`.

The changes are as follows:

- Use `#import "WPKit-Swift.h` instead of `#import <WordPressKit/WordPressKit-Swift.h>` to avoid the issue described in https://github.com/CocoaPods/CocoaPods/issues/7594.

You can see that the tests still build and pass with the changes I've made.

Related to https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/203 and https://github.com/wordpress-mobile/WordPressKit-iOS/pull/151

### DIFF
--- a/WordPressAuthenticator.xcodeproj/project.pbxproj
+++ b/WordPressAuthenticator.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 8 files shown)


mindgraffiti:
Hey James!

I'm not sure what to test. Is this supposed to work with the existing code in the host apps' develop branches?

PR #95, opened 2019-05-10
Update `WordPressShared` pod version

jaclync:
* This PR updated `WordPressShared` dependency to the latest version `1.8.0-beta.1` (https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/204), and updated its own pod version to `1.5.0-beta.3`.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.5.0-beta.2"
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator.podspec
```diff
@@ -40,5 +40,5 @@ Pod::Spec.new do |s|
   s.dependency 'GoogleSignIn', '~> 4.4'
   s.dependency 'WordPressUI', '~> 1.0'
   s.dependency 'WordPressKit', '~> 4.1.0-beta'
```
danielebogo: This needs to be updated as well!

PR #93, opened 2019-05-08
Make WordPress.com URLs configurable through WordPressAuthenticatorConfiguration

jtreanor:
I recently made changes to WordPressKit to allow WP.com urls to be configured for testing purposes (https://github.com/wordpress-mobile/WordPressKit-iOS/pull/141). This is the corresponding changes to WordPressAuthenticator to take advantage of that.

I have added two new items to `WordPressAuthenticatorConfiguration`: `wpcomBaseURL` and `wpcomAPIBaseURL`. They default to the correct values so that existing installations will not break.

I have also updated the `.podspec` version since this includes API changes.

### Testing Details

- All tests pass and the podspec validates.
- (Optional) Check out my corresponding WPiOS branch (`try/configure-base-urls`) and do a smoke test of the login flow. Draft PR is here: https://github.com/wordpress-mobile/WordPress-iOS/pull/11655.

- [x] Please check here if your pull request includes additional test coverage.

PR #96, opened 2019-05-14
Use wpcomAPIBaseURL in all WordPressComRestApi instances

jtreanor:
As I have been developing networking mocking, I have discovered a few places that the WP.com API URL needs to be injected which I missed in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/93.

This PR makes sure every instance of `WordPressComRestApi` in WordPressAutheticator uses the injectable `wpcomAPIBaseURL`.

### DIFF
--- a/WordPressAuthenticator/Signup/SignupEmailViewController.swift
+++ b/WordPressAuthenticator/Signup/SignupEmailViewController.swift
@@ -140,7 +140,8 @@ class SignupEmailViewController: LoginViewController, NUXKeyboardResponder {
 
     private func checkEmailAvailability(completion:@escaping (Bool) -> ()) {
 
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Signup/SignupEmailViewController.swift
```diff
@@ -140,7 +140,7 @@ class SignupEmailViewController: LoginViewController, NUXKeyboardResponder {
 
     private func checkEmailAvailability(completion:@escaping (Bool) -> ()) {
 
-        let remote = AccountServiceRemoteREST(wordPressComRestApi: WordPressComRestApi())
+        let remote = AccountServiceRemoteREST(wordPressComRestApi: WordPressComRestApi(baseUrlString: WordPressAuthenticator.shared.configuration.wpcomAPIBaseURL))
```
jkmassel: I don't think we enforce it on this repo, but this line is rather long – could probably be broken up into smaller statements to be a bit more understandable?

Up to you :)

PR #98, opened 2019-05-24
Allowing to continue login flow for self hosted sites that have Jetpack.

yaelirub:
Self hosted sites that have jet pack return with a successful response from the fetch site so we don't show the "logout" prompt for those sites.

**Before**:
self_hosted_jetpack1

**After**:
self_hosted_jetpack

### DIFF
--- a/WordPressAuthenticator/Model/WordPressComSiteInfo.swift
+++ b/WordPressAuthenticator/Model/WordPressComSiteInfo.swift
@@ -17,7 +17,7 @@ public class WordPressComSiteInfo {
     ///
     public let url: String
 
-    /// Indicates if Jetpack is available, or not,
+    /// Indicates if Jetpack is available, or not.
     ///
     public let hasJetpack: Bool
 
@@ -25,6 +25,8 @@ public class WordPressComSiteInfo {
     ///
     public let icon: String
     
+    /// Indicates whether the site is WordPressDotCom, or not.
+    ///
     public let isWPCom: Bool
 
 
--- a/WordPressAuthenticator/Model/WordPressComSiteInfo.swift
+++ b/WordPressAuthenticator/Model/WordPressComSiteInfo.swift
@@ -24,6 +24,8 @@ public class WordPressComSiteInfo {
     /// URL of the Site's Blavatar.
     ///
     public let icon: String
+    
+    public let isWPCom: Bool
 
 
 
@@ -35,5 +37,7 @@ public class WordPressComSiteInfo {
         url         = remote["URL"] as? String          ?? ""
         hasJetpack  = remote["jetpack"] as? Bool        ?? false
         icon        = remote["icon.img"] as? String     ?? ""
+        isWPCom     = remote["isWordPressDotCom"] as? Bool ?? false
+        
     }
 }
--- a/WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
@@ -197,16 +197,12 @@ class LoginSiteAddressViewController: LoginViewController, NUXKeyboardResponder 
                 return
             }
             self.configureViewLoading(false)
-            if siteInfo.isWPCom {
-                if WordPressAuthenticator.shared.delegate?.allowWPComLogin == true {
-                    self.presentNextControllerIfPossible(siteInfo: siteInfo)
-                } else {
-                    // Hey, you have to log out of your existing WP.com account before logging into another one.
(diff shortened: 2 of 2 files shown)


yaelirub:
@aerych , @mindgraffiti , I took a shoot a this as [more users are experiencing difficulties logging in](https://github.com/wordpress-mobile/WordPress-iOS/issues/11526#issuecomment-495374057):  

Can you think of any issue with allowing the flow to continue with this added condition?

yaelirub:
@aerych , @mindgraffiti would you mind taking a look?

aerych:
Sorry Yael! This one got lost over the holiday shuffle. Taking a look.

WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
```diff
@@ -194,7 +194,7 @@ class LoginSiteAddressViewController: LoginViewController, NUXKeyboardResponder
         let baseSiteUrl = WordPressAuthenticator.baseSiteURL(string: loginFields.siteAddress)
         let successBlock: (WordPressComSiteInfo) -> Void = { [weak self] siteInfo in
             self?.loginFields.meta.siteInfo = siteInfo
-            if WordPressAuthenticator.shared.delegate?.allowWPComLogin == false {
+            if WordPressAuthenticator.shared.delegate?.allowWPComLogin == false && !siteInfo.hasJetpack {
```
aerych: So... there are a few things going on in `fetchSiteInfo` and its associated code paths that seem off to me. For example, the `else` block in the conditional following the success block is never executed. I suspect there might be more work to be done in this method.  That said....

I think this change might introduce a hidden bug -- one that is not immediately obvious nor experienced but _could_ be if something else changed.  The logic here says: "If the delegate does not allow a wpcom login and the site does not have Jetpack, then show the prompt".  This works for WordPress.com sites since they do not have Jetpack since the conditional evalutes to false.  This works for self-hosted sites that have Jetpack since the conditional evaluates to true.  This _would fail_ for self-hosted sites without Jetpack if it was ever evaluated for one.  Currently the code path for a self-hosted non-Jetpack site does not encounter this check since the API call results in the failure block being executed.   Subtle, eh? 

I wonder if a more robust fix would be to just call the unauthenticated endpoint.  The `connect/site-info` results include a filed for `isWordPressDotCom` which could be checked first.  Then if true the delegate could be consulted about whether to allow a dotcom login.  Its a larger change but maybe worth it to clarify the work being done (and potentially clean up the other code paths that go through here.)

Thoughts?

yaelirub: >This would fail for self-hosted sites without Jetpack if it was ever evaluated for one
Oh! I didn't realize that.

I'm going to give your proposed solution a try and see if it makes more sense. I love the idea of cleaning this method up a bit.

yaelirub:
@aerych , refactored `fetchInfo` according to your suggested solution and tested on self hosted, self hosted with JP, WP public and private and all seem to work. 
ready for another look.

WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
```diff
@@ -188,69 +188,36 @@ class LoginSiteAddressViewController: LoginViewController, NUXKeyboardResponder
...
-                        self?.configureViewLoading(false)
-                        guard let self = self else {
-                            return
-                        }
-
-                        self.presentNextControllerIfPossible(siteInfo: nil)
-                    })
+            self.configureViewLoading(false)
+            if siteInfo.isWPCom {
+                if WordPressAuthenticator.shared.delegate?.allowWPComLogin == true {
```
aerych: _Loving_ the change ❤️. 
One nitpick, maybe we could tweak this to do the work with fewer conditionals?  Maybe something like
```
            if siteInfo.isWPCom && WordPressAuthenticator.shared.delegate?.allowWPComLogin == false {
                // Hey, you have to log out of your existing WP.com account before logging into another one.
                self.promptUserToLogoutBeforeConnectingWPComSite()
                return;
            }
            self.presentNextControllerIfPossible(siteInfo: siteInfo)
```

WordPressAuthenticator/Model/WordPressComSiteInfo.swift
```diff
@@ -24,6 +24,8 @@ public class WordPressComSiteInfo {
     /// URL of the Site's Blavatar.
     ///
     public let icon: String
+    
+    public let isWPCom: Bool
```
aerych: For consistency could we add a comment describing the property?  Its pretty obvious but 🤷‍♂.

PR #101, opened 2019-06-10
Rename WordPressAuthenticator.bundle to WordPressAuthenticatorResources.bundle to avoid conflicts

jtreanor:
This change ensures that the crash that occurred in https://github.com/wordpress-mobile/WordPress-iOS/issues/11848 cannot happen for WordPressAuthenticator.

Apart from the small bundle rename from `WordPressAuthenticator.bundle` to `WordPressAuthenticatorResources.bundle`, all the storyboards and xibs have been updated to load from the correct module. This won't be needed once CocoaPods 1.8 is released (see https://github.com/CocoaPods/CocoaPods/pull/8890).

@jkmassel Since we are not encountering the crash in WPAuthenticator, we could not make this change and wait for the CocoaPods update to land. What do you think?


jtreanor:
Yep, I agree its best to remove the risk here.

PR #104, opened 2019-06-21
Remove unused call to WPStyleGuide.configureColors

nheagy:
This method will be removed from the Shared pod.

See https://github.com/wordpress-mobile/WordPress-iOS/pull/11967


nheagy:
Thanks @danielebogo!

PR #105, opened 2019-06-28
Update for new WPiOS colors

nheagy:
This PR makes changes necessary for WPiOS to adopt a new color scheme:
- change the prologue animations and colors
- add new, optional, style properties for prologue text and bg colors
- remove the default blue of the navigation bar bg color


WordPressAuthenticator/Authenticator/WordPressAuthenticatorStyles.swift
```diff
@@ -70,6 +78,8 @@ public struct WordPressAuthenticatorStyle {
         self.subheadlineColor = subheadlineColor
         self.viewControllerBackgroundColor = viewControllerBackgroundColor
         self.navBarImage = navBarImage
+        self.prologueBackgroundColor = prologueBackgroundColor ?? WPStyleGuide.wordPressBlue()
```
frosty: Should we just set these as default parameters?

WordPressAuthenticator.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.5.4-beta.2"
+  s.version       = "1.5.4-beta.3"
```
frosty: Do you think these changes warrant a bigger version bump?

PR #107, opened 2019-07-15
Make remaining colors configurable

nheagy:
To finish my efforts to use the new WPiOS color scheme during login, this PR makes the remaining colors from the old WPiOS color palette configurable. It also removes the default `WordPressAuthenticatorStyle`, and makes the `style` argument mandatory.

This will have compatibility implications for the other projects that use this pod.

### Examples:

image
image

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorStyles.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorStyles.swift
@@ -82,25 +82,3 @@ public struct WordPressAuthenticatorStyle {
         self.prologueTitleColor = prologueTitleColor
     }
 }
-
-public extension WordPressAuthenticatorStyle {
-    static var defaultStyle: WordPressAuthenticatorStyle {
-        return WordPressAuthenticatorStyle(primaryNormalBackgroundColor: WPStyleGuide.mediumBlue(),
-                                           primaryNormalBorderColor: WPStyleGuide.wordPressBlue(),
-                                           primaryHighlightBackgroundColor: WPStyleGuide.wordPressBlue(),
-                                           primaryHighlightBorderColor: WPStyleGuide.wordPressBlue(),
-                                           secondaryNormalBackgroundColor: UIColor.white,
-                                           secondaryNormalBorderColor: WPStyleGuide.greyLighten20(),
-                                           secondaryHighlightBackgroundColor: WPStyleGuide.greyLighten20(),
-                                           secondaryHighlightBorderColor: WPStyleGuide.greyLighten20(),
-                                           disabledBackgroundColor: UIColor.white,
-                                           disabledBorderColor: WPStyleGuide.greyLighten30(),
-                                           primaryTitleColor: UIColor.white,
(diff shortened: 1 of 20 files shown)


aerych:
Heya @nheagy !  Looking good from here.   I think CI is being cranky about this warning:
Screen Shot 2019-07-16 at 10 16 04 AM

Thoughts?

nheagy:
@aerych I hadn't noticed that. Pushed another commit to move `font` into the `guard`.

nheagy:
@aerych I pushed one more commit as I found some more colors to change in the signup epilogue. Care to give it a glance?

aerych:
Hard to not miss a few strays given the amount of code here.  :)   Its all good. My :shipit: stands.

nheagy:
Thanks for your reviews! Does one of you have time for a small re-review?

@jaclync based on your prod to fix `self.attributedPlaceholder` I feel into a rabbit hole and simplified this code quite a bit!

WordPressAuthenticator/UI/SiteInfoHeaderView.swift
```diff
@@ -113,9 +113,9 @@ private extension SiteInfoHeaderView {
             blavatarImageView.layer.borderWidth = 0
             blavatarImageView.tintColor = .clear
         } else {
-            blavatarImageView.layer.borderColor = WPStyleGuide.greyLighten20().cgColor
+            blavatarImageView.layer.borderColor = WordPressAuthenticator.shared.style.instructionColor.cgColor
             blavatarImageView.layer.borderWidth = 1
-            blavatarImageView.tintColor = WPStyleGuide.greyLighten10()
+            blavatarImageView.tintColor = WordPressAuthenticator.shared.style.instructionColor
```
jaclync: the old diffs seem to have different shades of grey for the border and tint, and the new diffs seem to have the same - just double checking this is an intentional change?

WordPressAuthenticator/Authenticator/WordPressAuthenticatorStyles.swift
```diff
@@ -75,32 +85,14 @@ public struct WordPressAuthenticatorStyle {
         self.primaryTitleColor = primaryTitleColor
         self.secondaryTitleColor = secondaryTitleColor
         self.disabledTitleColor = disabledTitleColor
+        self.textButton = textButton
+        self.textButtonHighlight = textButtonHighlight
```
jaclync: sorry didn't comment on this earlier, a non-blocking nit: maybe `buttonTextColor`/`buttonTextHighlightedColor` would sound more clear that these two are colors not buttons? also more consistent with the names of other arguments (same for the property names)

nheagy: I agree, these should be suffixed with `Color`. 

This whole struct has gotten a bit big for my tastes, but I'll leave a more thorough refactor for another time 😉

nheagy:
@jaclync great suggestions, now fixed!

nheagy:
Thanks @jaclync and @aerych!

PR #110, opened 2019-07-18
`WordPressComSiteInfo`: update setting `hasJetpack` from remote response

jaclync:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/1113

- WCiOS PR branch for testing: https://github.com/woocommerce/woocommerce-ios/pull/1118
- WPiOS PR branch for testing: https://github.com/wordpress-mobile/WordPress-iOS/pull/12170

## Background

In [this PR](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/98/files), we switched the API call to fetch site info during site address sign in flow from `fetchSiteInfo` to `fetchUnauthenticatedSiteInfoForAddress` in order to accommodate both public and private sites (`fetchSiteInfo` only works for public sites). However, the responses from these two calls are very different.

With `fetchSiteInfo` (what we called before for WP.com sites), the response is like:
```
{
    "ID": 161199960,
    "name": "fun testing",
    "description": "",
    "URL": "https://funtestingusa.wpcomstaging.com",
    "jetpack": true,
    ...
}
```

And with `fetchUnauthenticatedSiteInfoForAddress`, the response looks like:
```
{
    "urlAfterRedirects": "https://funtestingusa.wpcomstaging.com",
    "isWordPress": true,
    "hasJetpack": true,
    "isWordPressDotCom": true
    ...
}
```

Originally, the site info’s `hasJetpack` is decoded from the response’s `jetpack` field, but now this info lives in the response's `hasJetpack` field. Thus, site info’s `hasJetpack` field is always false.

This is not a problem in WPiOS because WPiOS [does not check any other fields in the site info ](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift#L184) before proceeding to the next screen vs. WCiOS [only continues when site info's `hasJetpack` field is true](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L138) in the authenticator delegate method. That’s why the site address is not considered valid and stuck on the site address screen in WCiOS.

## Changes

- Updated how we populate `hasJetpack` field in `WordPressComSiteInfo` from remote response's `jetpack` field to `hasJetpack` field
- Notes: there are more fields that are not populated correctly with the new API response (e.g. `name`, `url`, `icon`), but I'd like to keep this PR small to just fix the sign in issue in WCiOS

## Testing

In [WCiOS branch](https://github.com/woocommerce/woocommerce-ios/pull/1118)
- Sign in with site address --> should be able to complete the flow to enter a WC store

In [WPiOS branch](https://github.com/wordpress-mobile/WordPress-iOS/pull/12170)
- Sign in with site address --> should work as before

### DIFF
--- a/WordPressAuthenticator/Model/WordPressComSiteInfo.swift
+++ b/WordPressAuthenticator/Model/WordPressComSiteInfo.swift
@@ -37,7 +37,7 @@ public class WordPressComSiteInfo {
         name        = remote["name"] as? String         ?? ""
         tagline     = remote["description"] as? String  ?? ""
         url         = remote["URL"] as? String          ?? ""
-        hasJetpack  = remote["jetpack"] as? Bool        ?? false
+        hasJetpack  = remote["hasJetpack"] as? Bool     ?? false
         icon        = remote["icon.img"] as? String     ?? ""
         isWPCom     = remote["isWordPressDotCom"] as? Bool ?? false
         


jaclync:
I just tested on WPiOS, but signing in with site address didn't work for me on the latest `develop`, the [testing branch](https://github.com/wordpress-mobile/WordPress-iOS/compare/issue/wcios-1113-authenticator-sign-in-site-address-testing?expand=1), and also on the latest app store build. Do you see the same issue, or maybe I missed something?

Simulator Screen Shot - iPhone SE - 2019-07-18 at 18 22 36

ctarda:
@jaclync I just tested with WooCommerce and for this store:


I got this error back:


To be fair, that is the same result I get when trying to log in with our `develop`

Also, the colour palette has changed completely, this is what we have on develop right now:

jaclync:
@ctarda thanks for testing!

> I just tested with WooCommerce and for this store. To be fair, that is the same result I get when trying to log in with our develop

Hmm are you able to sign in with the same address in production from the App Store? Also, could you try again including `http://` in the address?

> Also, the colour palette has changed completely, this is what we have on develop right now

Yep if we update the authenticator pod to the latest with color changes (still in beta), then we'd update the colors for the new color variables (`buttonTextColor`, `buttonTextHighlightColor`, `placeholderColor` etc.). I set these colors to some native colors just for testing, since I didn't know the right colors right away 😛 If we want to update the authenticator pod to a newer stable version like 1.5.3 or 1.6.0, we don't have to make the color changes.

yaelirub:
@jaclync , I'm going to review and test soon but just a small comment on the description of this issue:
>we switched the API call to fetch site info during site address sign in flow from `fetchUnauthenticatedSiteInfoForAddress` to `fetchSiteInfo`

We actually switched from `fetchSiteInfo` to `fetchUnauthenticatedSiteInfoForAddress`

jaclync:
> We actually switched from fetchSiteInfo to fetchUnauthenticatedSiteInfoForAddress

@yaelirub my bad! Just updated the PR description :)

ctarda:
@jaclync sorry for the delay. I still can not login to store.ctarda.com, neither with nor without the scheme, but I can not do it in production either (same error)

I tested with a different site, without a subdomain and it worked.

jaclync:
thanks for reviewing & testing! 😃 merging now and closing the testing PRs right after

PR #111, opened 2019-08-13
Log In: show button view with login options.

ScoutHarris:
This adds a new Log In flow. Namely, if `showNewLoginFlow` is `true`, when Log In is selected a button view now appears with available options.

This can be tested with WPiOS PR https://github.com/wordpress-mobile/WordPress-iOS/pull/12305.

### DIFF
--- a/WordPressAuthenticator/Signin/LoginPrologueLoginMethodViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueLoginMethodViewController.swift
@@ -1,0 +1,73 @@
+import WordPressUI
+import WordPressShared
+
+class LoginPrologueLoginMethodViewController: NUXViewController {
+    /// Buttons at bottom of screen
+    private var buttonViewController: NUXButtonViewController?
+
+    /// Gesture recognizer for taps on the dialog if no buttons are present
+    fileprivate var dismissGestureRecognizer: UITapGestureRecognizer?
+
+    open var emailTapped: (() -> Void)?
+    open var googleTapped: (() -> Void)?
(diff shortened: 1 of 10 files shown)


WordPressAuthenticator/Signin/LoginPrologueLoginMethodViewController.swift
```diff
@@ -0,0 +1,73 @@
...
+    private func configureButtonVC() {
+        guard let buttonViewController = buttonViewController else {
+            return
+        }
+
+        let wordpressTitle = NSLocalizedString("Continue with WordPress.com", comment: "Button title. Tapping begins our normal log in process.")
+        let googleTitle = NSLocalizedString("Continue with Google", comment: "Button title. Tapping begins log in using Google.")
+        
+        buttonViewController.setupTopButton(title: wordpressTitle, isPrimary: false, accessibilityIdentifier: "Log in with Email Button") { [weak self] in
+            self?.dismiss(animated: true)
```
frosty: Should we be tracking a tap on this button too?

ScoutHarris: Sorry, I didn't think to make a note. I'm keeping a list of new tracks that are needed. I"ll add them a bit later.

ScoutHarris:
> Also, of note: the "Or login using your site address" text appears much smaller on iOS 13 – we'll want to fix that somehow.

Gotcha. Thanks for that.

ScoutHarris:
Site address font fixed in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/115.

PR #113, opened 2019-08-17
[SIWA] Log In: add 'Continue with Apple' button to login options

ScoutHarris:
If the following conditions are met, a 'Continue with Apple' button is added to the Log In options button view.
- Is Xcode 11.
- Is iOS 13.
- The `WordPressAuthenticatorConfiguration:enableSignInWithApple` flag is `true`.

apple_button

Since there is no button action just yet, tapping the SIWA button will result in:
- The button view being dismissed.
- Log messages in the console. Ex: `Login Prologue: Apple tapped.`

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12334

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.8.0-beta.2"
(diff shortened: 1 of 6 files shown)


astralbodies:
:shipit:

PR #117, opened 2019-08-22
[SIWA] Show new account epilogue

ScoutHarris:
This adds logic to show the account created epilogue after a WP account is created with an Apple ID. To note, the endpoint doesn't yet support Apple, so account creation will fail, thus the epilogue won't actually show yet.

This also passes Apple specific parameters to the social connect endpoint in WPKit, although it's not used yet.

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12360
Ref WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/182

### DIFF
--- a/WordPressAuthenticator/Signin/LoginViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginViewController.swift
@@ -262,20 +262,33 @@ extension LoginViewController {
         guard let serviceName = loginFields.meta.socialService, let serviceToken = loginFields.meta.socialServiceIDToken else {
             return
         }
-
-        linkSocialService(serviceName: serviceName, serviceToken: serviceToken, wpcomAuthToken: wpcomAuthToken)
+        
+        linkSocialService(serviceName: serviceName,
+                          serviceToken: serviceToken,
+                          wpcomAuthToken: wpcomAuthToken,
+                          appleEmail: loginFields.meta.appleUser?.email,
+                          appleFullName: loginFields.meta.appleUser?.fullName)
     }
 
     /// Links the current WordPress Account to a Social Service.
     ///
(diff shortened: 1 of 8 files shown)


WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -262,20 +262,33 @@ extension LoginViewController {
...
     /// Links the current WordPress Account to a Social Service.
     ///
-    func linkSocialService(serviceName: SocialServiceName, serviceToken: String, wpcomAuthToken: String) {
-        guard serviceName == .google else {
+    func linkSocialService(serviceName: SocialServiceName,
+                           serviceToken: String,
+                           wpcomAuthToken: String,
+                           appleEmail: String? = nil,
+                           appleFullName: String? = nil) {
+        guard serviceName == .google || serviceName == .apple else {
```
frosty: Because it's an enum, `serviceName` can't be anything other that these, can it? Is this guard needed?

WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -262,20 +262,33 @@ extension LoginViewController {
...
     }
 
     /// Links the current WordPress Account to a Social Service.
     ///
-    func linkSocialService(serviceName: SocialServiceName, serviceToken: String, wpcomAuthToken: String) {
-        guard serviceName == .google else {
+    func linkSocialService(serviceName: SocialServiceName,
+                           serviceToken: String,
+                           wpcomAuthToken: String,
+                           appleEmail: String? = nil,
```
frosty: Same comment as in WPKit, I'm not sure about including the apple parameters directly here. Perhaps we could add them to a parameters dictionary and pass it in?

ScoutHarris:
Hey @frosty . Changes made and ready for another review please.

PR #121, opened 2019-08-26
Support existing Apple accounts going right into the epilogue

astralbodies:
Adds an additional logical path for account coming in through Sign in with Apple. This also shows a spinner during the process of creating an account through the API.

1. New WordPress.com user.
2. Existing WordPress.com user matching e-mail address of first time SIWA user.
3. Existing SIWA/WordPress.com user (this PR adds this).

**To test:**

1. Sign in with Sign in with Apple for the first time.
2. Log out.
3. Sign back in using the same SIWA account.
4. Verify the login epilogue displays rather than the signup epilogue.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.8.0-beta.11"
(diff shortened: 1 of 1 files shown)


frosty:
@astralbodies I pushed up a fix to check for `user_exists` as an error instead of in the success response. It seems to work for me. Do you want to re-test?

PR #123, opened 2019-08-29
Change copy per Editorial and to reflect Android app

lorenzolewis:
To address wordpress-mobile/WordPress-iOS#12188

Test by logging out and then attempting to log in by using an invalid URL (such as %$#), error text should be "Please enter a complete website address, like example.com." and title text should be "Enter the address of the WordPress site you'd like to connect."


jkmassel:
Merged! Thanks @lorenzolewis!

PR #128, opened 2019-09-04
[SIWA] Sync account with host app for existing accounts

ScoutHarris:
Ref WPiOS issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/12434

The incorrect delegate method was being called with existing accounts. Now, if logging into an existing account, the `sync` method is called (instead of `createdWordPressComAccount`).

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12471


astralbodies:
@Tug is making endpoint changes to return that information to be able to be pulled from the response to push into the UI.

PR #141, opened 2019-10-01
[SIWA] Show email address on WP password entry view

ScoutHarris:
Ref WPiOS issue #https://github.com/wordpress-mobile/WordPress-iOS/issues/12556
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/183

This uses the changes in WPKit to capture the existing account's email address for display on the WP password entry view.

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12606

(ping @ctarda )

### DIFF
--- a/WordPressAuthenticator/Services/SignupService.swift
+++ b/WordPressAuthenticator/Services/SignupService.swift
@@ -71,10 +71,13 @@ class SignupService {
                                     success(createdAccount, false, username, bearer_token)
         }, failure: { error in
             if let error = (error as NSError?) {
-                
+
+                // If an account already exists, the account email should be returned in the Error response.
+                // Extract it and return it.
                 var existingEmail = ""
-                if let errorData = error.userInfo[WordPressComRestApi.ErrorKeyErrorData] as? [String: String],
-                    let email = errorData.first?.value {
+                if let errorData = error.userInfo[WordPressComRestApi.ErrorKeyErrorData] as? [String: String] {
+                    let emailDict = errorData.first { $0.key == "email" }
+                    let email = emailDict?.value ?? ""
                     existingEmail = email
                 }
 
--- a/WordPressAuthenticator/Services/SignupService.swift
+++ b/WordPressAuthenticator/Services/SignupService.swift
@@ -76,7 +76,7 @@ class SignupService {
                 // Extract it and return it.
                 var existingEmail = ""
                 if let errorData = error.userInfo[WordPressComRestApi.ErrorKeyErrorData] as? [String: String] {
-                    let emailDict = errorData.first { $0.key == "email" }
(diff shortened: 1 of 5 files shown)


ScoutHarris:
The test will continue to fail until the podspec is updated with an official WPKit version. I'll do that when I create a new WPKit release. However, this should not prevent testing this PR.

WordPressAuthenticator/Services/SignupService.swift
```diff
@@ -71,9 +71,16 @@ class SignupService {
                                     success(createdAccount, false, username, bearer_token)
         }, failure: { error in
             if let error = (error as NSError?) {
+                
+                var existingEmail = ""
+                if let errorData = error.userInfo[WordPressComRestApi.ErrorKeyErrorData] as? [String: String],
+                    let email = errorData.first?.value {
```
frosty: I don't know if this is potentially confusing, as we're assuming that an error from the endpoint is an email address and that it's valid data to display to the user as an email address 🤔 

Perhaps a code comment could help here to explain? And I wonder if we should check the error code before using the data, so we can be sure it's the right kind of error?

ScoutHarris: Hey @frosty .

I made the test a bit smarter. See what you think now. Thanks!

WordPressAuthenticator/Services/SignupService.swift
```diff
@@ -71,10 +71,13 @@ class SignupService {
...
             if let error = (error as NSError?) {
-                
+
+                // If an account already exists, the account email should be returned in the Error response.
+                // Extract it and return it.
                 var existingEmail = ""
-                if let errorData = error.userInfo[WordPressComRestApi.ErrorKeyErrorData] as? [String: String],
-                    let email = errorData.first?.value {
+                if let errorData = error.userInfo[WordPressComRestApi.ErrorKeyErrorData] as? [String: String] {
+                    let emailDict = errorData.first { $0.key == "email" }
```
frosty: I like this! Could we create a static value for `email` instead of putting the literal here? Perhaps alongside `WordPressComRestApi.ErrorKeyErrorData`.

ScoutHarris: @astralbodies @frosty - a new constant has been added to WPKit. Ready for another review please.

frosty:
I'm not sure why the build is failing here. It seems to work fine for me locally, and the WPiOS branch works as expected and is building with CI.

ScoutHarris:
> I'm not sure why the build is failing here. It seems to work fine for me locally, and the WPiOS branch works as expected and is building with CI.

Hey @frosty . 

I commented on that [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/141#issuecomment-537235383) .

frosty:
Sorry, missed that! All good then :)

PR #142, opened 2019-10-02
[Dark Mode] Update activity indicator color when button is disabled

### LINKED ISSUE #140: [Dark Mode] NUX Button loading
Activity indicator on secondary NUX buttons appears white-on-white in light mode

danielebogo:
Fixes #140 

This PR reload the activity indicator color when the button `isEnabled` property is true/false

## To Test
- Run WP iOS [#12610](https://github.com/wordpress-mobile/WordPress-iOS/pull/12610)

### DIFF
--- a/WordPressAuthenticator/NUX/NUXButton.swift
+++ b/WordPressAuthenticator/NUX/NUXButton.swift
@@ -20,14 +20,6 @@ import WordPressUI
         return indicator
     }()
 
-    open override var isEnabled: Bool {
-        didSet {
-            if #available(iOS 13, *) {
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator/NUX/NUXButton.swift
```diff
@@ -69,7 +82,10 @@ import WordPressUI
     open override func awakeFromNib() {
         super.awakeFromNib()
         configureAppearance()
-        activityIndicator.style = .gray
+        guard #available(iOS 13, *) else {
+            activityIndicator.style = .gray
```
frosty: Do we need this? Should we just initialise the indicator with a `.gray` appearance instead of white?

PR #147, opened 2019-10-09
Use CocoaPods CDN specs repo to reduce CI runtime from ~12 minutes to ~3 minutes

jtreanor:
We're coming closer to hitting our CircleCI limit each month so I'm making a few small changes to improve usage and hopefully improve the development cycle as well.

In 1.7.0, CocoaPods introduced a [CDN specs repo](https://blog.cocoapods.org/CocoaPods-1.7.2/) which could be used to replace the huge, slow git-based specs repo. In 1.8.0 CocoaPods started using this by default, so we will need to switch at some stage. I'm starting on this repo as it hopefully is small and manageable.

The only disadvantage of using the CDN specs repo for us is that there is a delay of about 4 minutes from when `pod trunk push` completes and a pod is available in the CDN specs repo. For WordPressAuthenticator, this means if you try to use a new version of WordPressShared within 4 minutes of it being published, `pod install` won't find it. You can see deploys to the CocoaPods Specs CDN [here](https://app.netlify.com/sites/cocoapods-cdn/deploys). Given this speeds things up by much more than 4 minutes, I feel that this tradeoff is worthwhile.

So here are the changes I made to implement this:

- Updated CocoaPods to 1.8.4.
- Updated the CircleCI Orb to get improved CDN specs support (https://github.com/wordpress-mobile/circleci-orbs/pull/39).
- Changed the `source` in the `Podfile` to use CDN specs.
- Removed `cocoapods-repo-update` since we never need to run `pod repo update` with the CDN specs.

I have also tested the `Publish to Trunk` CircleCI job [here](https://circleci.com/gh/wordpress-mobile/WordPressAuthenticator-iOS/998). Run time for that is reduced from ~13 minutes to ~3.5 minutes.

## To Test

- See that CircleCI is green and runs much more quickly.
- Checkout this branch and run `bundle install && bundle exec pod install`. Confirm it completes and there are no changes to `Podfile.lock`.


jtreanor:
> I wonder if we can add a line about the small delay in availability to the message we post on slack at the end of the pod trunk action.

Thats a very good idea! I can add that.

PR #152, opened 2019-10-29
Consistently track Google and Apple sign up/log in actions

ScoutHarris:
This updates tracking signup and login for social accounts:
- To be consistent between Google and Apple (i.e. they track the same actions). 
- Adding ["source": "google"] to Google tracks to differentiate between the two.

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/12803


ScoutHarris:
Thanks @frosty !

PR #156, opened 2019-11-20
Add signed_in Tracks event whenever we log in socially

frosty:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/12827. This PR adds `signedIn` tracks events everywhere we're already posting a `loginSocialSuccess` event. You can test this PR with the associated WPiOS PR here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12981

I updated 4 locations where I could find usages of `loginSocialSuccess` with no associated `signedIn`. I'm not 100% sure if these are correct, so please let me know what you think! Here's my thoughts on each one:

* **Login2FAController**: I'm a little unsure about this one, as the line above mentions Google login but I'm honestly not quite sure how to trigger this method. It doesn't appear to get called by doing a normal sign in with 2FA.
* **LoginEmailViewController**: I updated this one, but as far as I can tell this method is actually not used any more, as we no longer display the Google option on this screen. Since the addition of Sign in with Apple, we moved it as an option at the start of the process.
* **LoginViewController**: I also couldn't seem to get this one called. I'm not sure if it's used when connecting an Apple account for the first time? It doesn't seem to be called for my already-connected Google or Apple accounts.
* **SignupGoogleViewController**: This one is definitely called when I log in with my Google account :)

**To test**

* Review the code and see if you think these placements look okay
* As far as possible, exercise each event here. As I mentioned, I don't think the email one is used any more, and I could only get the SignupGoogleViewController method called (by signing in with my existing Google account).
* Any thoughts you have on this would be very welcome!

### DIFF
--- a/WordPressAuthenticator/Signin/Login2FAViewController.swift
+++ b/WordPressAuthenticator/Signin/Login2FAViewController.swift
@@ -178,6 +178,8 @@ class Login2FAViewController: LoginViewController, NUXKeyboardResponder, UITextF
 
         // Disconnect now that we're done with Google.
         GIDSignIn.sharedInstance().disconnect()
+
+        WordPressAuthenticator.track(.signedIn)
         WordPressAuthenticator.track(.loginSocialSuccess)
     }
 
--- a/WordPressAuthenticator/Signin/LoginEmailViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginEmailViewController.swift
@@ -559,6 +559,7 @@ extension LoginEmailViewController {
 
         // Disconnect now that we're done with Google.
         GIDSignIn.sharedInstance().disconnect()
+        WordPressAuthenticator.track(.signedIn, properties: ["source": "google"])
         WordPressAuthenticator.track(.loginSocialSuccess, properties: ["source": "google"])
     }
 
--- a/WordPressAuthenticator/Signup/SignupGoogleViewController.swift
+++ b/WordPressAuthenticator/Signup/SignupGoogleViewController.swift
@@ -141,6 +141,7 @@ private extension SignupGoogleViewController {
     /// Social Login Successful: Analytics + Pushing the Login Epilogue.
     ///
     func wasLoggedInInstead(with credentials: AuthenticatorCredentials) {
+        WordPressAuthenticator.track(.signedIn, properties: ["source": "google"])
         WordPressAuthenticator.track(.signupSocialToLogin, properties: ["source": "google"])
         WordPressAuthenticator.track(.loginSocialSuccess, properties: ["source": "google"])
 
--- a/WordPressAuthenticator/Signup/SignupGoogleViewController.swift
+++ b/WordPressAuthenticator/Signup/SignupGoogleViewController.swift
@@ -133,6 +133,7 @@ private extension SignupGoogleViewController {
         // This stat is part of a funnel that provides critical information.  Before
         // making ANY modification to this stat please refer to: p4qSXL-35X-p2
         WordPressAuthenticator.track(.createdAccount, properties: ["source": "google"])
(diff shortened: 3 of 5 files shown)


ScoutHarris:
Login2FAViewController - It looks like this particular method is used by social + 2fa account.
LoginEmailViewController - you are correct.
LoginViewController - again correct. `linkSocialService` gets called when you first use either social account type, but not after.

WordPressAuthenticator/Signin/Login2FAViewController.swift
```diff
@@ -178,6 +178,8 @@ class Login2FAViewController: LoginViewController, NUXKeyboardResponder, UITextF
 
         // Disconnect now that we're done with Google.
         GIDSignIn.sharedInstance().disconnect()
+
+        WordPressAuthenticator.track(.signedIn)
```
ScoutHarris: I think we can add a `["source": "google"]` here since this is only used by Google as far as I can tell. Thoughts?

frosty: I'm happy to do that if that's the case, I just couldn't work out if it was used by not-Google logins. There are no other references to Google in here other than line 180, and the `loginSocialSuccess` call below doesn't mention Google so I wasn't sure!

ScoutHarris: Well, I just tried with an Apple account and was finally able to repro https://github.com/wordpress-mobile/WordPress-iOS/issues/12477! 🤦‍♂ 

I'd be ok with leaving it generic for now. Currently, it is only google, but hard-coding that is probably dangerous and might be wrong when 12477 is fixed.

frosty: 👍

WordPressAuthenticator/Signup/SignupGoogleViewController.swift
```diff
@@ -141,6 +141,7 @@ private extension SignupGoogleViewController {
     /// Social Login Successful: Analytics + Pushing the Login Epilogue.
```
aerych: We should probably also bump `.signedIn` from `socialSignupWasSuccessful`.  We bump both account_created  and signed_in events when creating a WordPress account via email magic link.  We should do the same for social flows.  (Tested to double check.)

WordPressAuthenticator/Signin/LoginEmailViewController.swift
```diff
@@ -559,6 +559,7 @@ extension LoginEmailViewController {
 
         // Disconnect now that we're done with Google.
         GIDSignIn.sharedInstance().disconnect()
+        WordPressAuthenticator.track(.signedIn, properties: ["source": "google"])
```
aerych: Wasn't able to reach this code path.  I'm guessing its orphaned code now that the social options are available from the prologue vs the LoginEmailViewController.  @ScoutHarris, thoughts?

ScoutHarris: That code is sort of feature flagged via the `showNewLoginFlow` flag in `WordPressAuthenticatorConfiguration`. I would suggest removing that flag and associated code in a separate PR. I'm pretty sure there's some storyboard mojo that needs to be removed as well. Yes, it's totally my fault - I haven't cleaned that up post SIWA launch.

frosty: After further inspection, I think Woo is still using this code? They still show the Google login option on the email screen. So we can leave this for now I think.

frosty:
Sorry for the delay - @aerych I pushed up some changes to address your feedback.

WordPressAuthenticator/Signin/LoginEmailViewController.swift
aerych: Good catch!

PR #168, opened 2019-12-11
Fix memory layout issues with DDLogLevel

jkmassel:
Rather than using int as the type, we use DDLogLevel per https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Documentation/DynamicLogLevels.md.


jkmassel:
@mindgraffiti – thanks for the review!!

PR #170, opened 2019-12-13
[VoiceOver] Self-hosted Login Flow enhancements

shiki:
This fixes the issues described in https://github.com/wordpress-mobile/WordPress-iOS/issues/13086. 

This is a dependency of https://github.com/wordpress-mobile/WordPress-iOS/pull/13099. This PR needs to be reviewed first. 

## Changes

The changes mainly involve the `LoginSiteAddressViewController` and `LoginSelfHostedViewController`. 

### Announce Errors

For errors that happen after pressing _Next_, we sometimes use the [`errorLabel`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/1361a86fec25f7b69d88f74a7bb676b739749713/WordPressAuthenticator/NUX/NUXViewController.swift#L56) to display the error. The problem with this is that it is a UI element that just appears unannounced. A VoiceOver user may not know that it's there and would expect that a network request is still happening in the background. 

This fixes that by making VoiceOver focus on the error label. Which, in effect, makes VoiceOver speak the error message. 

### Text Field Labels

Before | After 
--------|-------
  |  

For these view controllers, we rely on the text field's placeholders instead of labels to describe what the text field is. [Placeholders are a bit of a concern for accessibility](https://www.nngroup.com/articles/form-design-placeholders/). For this, I was mostly concerned about the VoiceOver user navigating back to the text field again and **not knowing** what it is. 

This changes the text fields so they will have a static label like “Username“ or “Password“. 

### Show Password Toggle Button

Shown | Hidden 
--------|-------
   image  |       image

The Show Password button didn't have a label. This is a toggle button so it's a bit tricky. Some apps would change the `accessibilityLabel` when the image changes. WPAndroid uses a static label and makes TalkBack say “selected“ or “not selected“. 

For this, I opted to have a static `accessibilityLabel` but change the `accessibilityValue` to the **current state** of the toggle button. 

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/cafee3d971f3212631d8dd063980925db8ab9480/WordPressAuthenticator/NUX/WPWalkthroughTextField.m#L278-L289

This is the same strategy used in https://github.com/wordpress-mobile/WordPress-iOS/pull/13057 which seems to work well. I believe it's a bad idea to change the `accessibilityLabel` as that would indicate that there are **2 different buttons**. 

### Others

- Added accessibility label for the 1Password button
- Fixed https://github.com/wordpress-mobile/WordPress-iOS/issues/12188 in 
0c3a93f
- Change the forgot password button's trait to be a `.link` since it opens Safari

## Testing

Please use the WPiOS branch at https://github.com/wordpress-mobile/WordPress-iOS/pull/13099. 

1. Log out
2. Turn VoiceOver on
3. Navigate to Login → Enter Site Address
4. Confirm that the site address text field has a label
5. Enter an invalid URL and press Next. An error will be shown in red. Confirm that VoiceOver speaks the error. 
6. Enter a valid self-hosted URL and press Next to navigate to the next page. 
7. Confirm that:
    a. The username and password text fields have labels
    b. The 1Password button has a label
    c. _Lost your password_ is announced as a link
8. Enter some values in the username and password fields. 
9. Navigate between the username and password fields. Confirm that their labels are still spoken by VoiceOver. 
10. Navigate VoiceOver to the Show Password button
11. Toggle the button a couple of times. Confirm that VoiceOver would always speak the current status like “Show password, Shown“ or “Show password, Hidden“. 

## Reviewer Checklist

- [ ] Review this PR
- [ ] Approve https://github.com/wordpress-mobile/WordPress-iOS/pull/13099 if you approved this PR 🙂 

## Submitter Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.
- [x] If it's feasible, I have added unit tests. 
- [ ] Create a release version after merging

--- 

@mindgraffiti Would you have some time to review this? 🙂

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.10.5-beta.3"
+  s.version       = "1.10.5-beta.4"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
@@ -25,6 +25,6 @@ public extension WordPressAuthenticatorDisplayStrings {
     static var defaultStrings: WordPressAuthenticatorDisplayStrings {
         return WordPressAuthenticatorDisplayStrings(emailLoginInstructions: NSLocalizedString("Log in to your WordPress.com account with your email address.", comment: "Instruction text on the login's email address screen."),
                                                     jetpackLoginInstructions: NSLocalizedString("Log in to the WordPress.com account you used to connect Jetpack.", comment: "Instruction text on the login's email address screen."),
-                                                    siteLoginInstructions: NSLocalizedString("Enter the address of your WordPress site you'd like to connect.", comment: "Instruction text on the login's site addresss screen."))
+                                                    siteLoginInstructions: NSLocalizedString("Enter the address of the WordPress site you'd like to connect.", comment: "Instruction text on the login's site addresss screen."))
     }
 }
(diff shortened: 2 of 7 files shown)


shiki:
Thank you for the review, @mindgraffiti! 

Re: WPCom login, it is tracked in a different task. But I appreciate you digging more into it. I haven't done that yet. I linked your research here https://github.com/wordpress-mobile/WordPress-iOS/issues/13089#issuecomment-566194483. 

> Also, TIL Authy does not support VoiceOver in the app ☹️

😞 I found some of the apps I use are also not good.

PR #178, opened 2020-01-23
V. 1.10.7-beta.1 - Magic Email Clients

Gio2018:
### Magic Email Clients
- Add AppSelector class, that creates an action sheet that contains a list of apps that can be called by tapping on its buttons. The specific use case are (magic) installed email clients
  - EmailClients.plist contains the URL Schemes of the currently supported email clients
  - Currently supported email clients:
    - Apple Mail (default)
    - Airmail
    - Fastmail
    - Gmail
    - Microsoft Outlook
    - Spark
    - Yahoo Mail

- Add LinkMailPresenter, that presents the action sheet on a UIViewController. If no other client is available, Apple Mail will be opened directly without presenting the action sheet
- Add the protocol URLHandler as a generic URL handler, injected into AppSelector. UIApplication conforms to URLHandler (and it's the default to handle URL Schemes)

Related WordPress-iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/13275


WordPressAuthenticator/Email Client Picker/AppSelector.swift
```diff
@@ -0,0 +1,122 @@
...
+        case .appleMail:
+            return NSLocalizedString("Mail (Default)", comment: "Label for Apple Mail as default email client")
+        case .gmail:
+            return NSLocalizedString("Gmail", comment: "Gmail")
+        case .airmail:
+            return NSLocalizedString("Airmail", comment: "Airmail")
+        case .msOutlook:
+            return NSLocalizedString("Microsoft Outlook", comment: "Microsoft Outlook")
+        case .spark:
+            return NSLocalizedString("Spark", comment: "Spark")
```
frosty: Do you think most of these need to be localised? I may be wrong, but I think e.g. Outlook and Gmail are still called the same thing in other languages? If we do keep these as localized strings, we may want to provide a little more context in the comments – for example, `Spark` on its own doesn't indicate that it's the name of an email client. Perhaps `Option to select the Spark email client when logging in with magic links`?

WordPressAuthenticator/Email Client Picker/LinkMailPresenter.swift
```diff
@@ -0,0 +1,39 @@
...
+                showAlertToCheckEmail(on: viewController)
+            }
+            return
+        }
+        viewController.present(picker.alertController, animated: true)
+    }
+
+    private func showAlertToCheckEmail(on viewController: UIViewController) {
+        let title = NSLocalizedString("Please check your email", comment: "Alert title for check your email during logIn/signUp.")
+        let message = NSLocalizedString("Please open your email app and look for an email from WordPress.com.", comment: "Message to ask the user to check their email and look for a WordPress.com email.")
```
frosty: I know these were existing strings, but perhaps we should match them to the messages we show on the web:

Title: "Check your email!"
Message: "We just emailed a link to EMAIL_ADDRESS. Please check your mail app and tap the link to log in."

If it's too difficult to include the email address here, we could just change that to "We just emailed you a link to log in. Please check your mail app and tap the link."

WordPressAuthenticator/Email Client Picker/AppSelector.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// initializes the picker with a dictionary. Initialization will fail if an empty/invalid app list is passed
+    /// - Parameters:
+    ///   - appList: collection of apps to be added to the selector
+    ///   - defaultAction: default action, if not nil, will be the first element of the list
+    ///   - urlHandler: object that handles app URL schemes; defaults to UIApplication.shared
+    init?(with appList: [String: String],
+          defaultAction: UIAlertAction? = nil,
+          urlHandler: URLHandler = UIApplication.shared) {
+        /// inline method that builds a list of app calls to be inserted in the action sheet
+        func buildAppCalls(from appList: [String: String]) -> [UIAlertAction]? {
```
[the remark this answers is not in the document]
Gio2018: If I use an instance method, the compiler would complain that I'm using self before initializing. It could be a class method but does not seem that appropriate, since it's only used at initialization time. The other option is to just put that code in the init.
Absolutely agree that it should be make instead of build, and makeAlertActions sounds more appropriate

WordPressAuthenticator/Email Client Picker/AppSelector.swift
Gio2018: I thought about that, I did it mostly for languages that have different character sets than English. I will provide more context.

WordPressAuthenticator/Email Client Picker/LinkMailPresenter.swift
Gio2018: we can use the email address. I think it's safe to assume that the address will always be a valid email address (otherwise users would not get a magic link and not land to this page). Do you see any issue with this assumption?

WordPressAuthenticator/Email Client Picker/AppSelector.swift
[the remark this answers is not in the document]
frosty: > If I use an instance method, the compiler would complain that I'm using self before initializing

Sorry, my bad – I always get tripped up by this! 🙃

WordPressAuthenticator/Email Client Picker/LinkMailPresenter.swift
frosty: I don't see an issue – it informs the user where we've attempted to send a link, so if it's not a valid email address for some reason it should help the user notice that.

Gio2018:
@frosty , @ScoutHarris , catching up on this after my vacation break! Please let me know if there are any other changes needed before approving it. Thanks!

WordPressAuthenticator/Email Client Picker/LinkMailPresenter.swift
```diff
@@ -26,8 +33,13 @@ class LinkMailPresenter {
     }
 
     private func showAlertToCheckEmail(on viewController: UIViewController) {
-        let title = NSLocalizedString("Please check your email", comment: "Alert title for check your email during logIn/signUp.")
-        let message = NSLocalizedString("Please open your email app and look for an email from WordPress.com.", comment: "Message to ask the user to check their email and look for a WordPress.com email.")
+        let title = NSLocalizedString("Check your email!",
+                                      comment: "Alert title for check your email during logIn/signUp.")
+        let message = NSLocalizedString("We just emailed a link to ",
```
frosty: I'd suggest using a localized string with format here, so that these strings can be combined into one – different languages may need to put the email address into a different order in the string, so it's easiest for translation if it's all together. Something like:

```
let message = String.localizedStringWithFormat(NSLocalizedString("We just emailed a link to %@. Please check your mail app and tap the link to log in.", comment: "blah"), emailAddress)
```

Gio2018: done

Issue #182, opened 2020-02-11
Unified Login & Signup


mindgraffiti:
~Duplicate the code, with storyboards and classes #187~

----

## Spike / Research
- [x] Spike: take 5 days and attempt to break apart the Login.storyboard. The purpose is to discover the feasibility of the task.

1. Remove the storyboard segues in Interface Builder and create programmatic ones.
2. Remove the IBActions in Interface Builder and create programmatic ones.
3. Remove delegates & datasources linked in Interface Builder and create programmatic ones.
4. ~Remove Scenes from Interface Builder and move to their own nibs~ (.xib)
5. Edit a ViewController and estimate how long it will take to make changes to the existing code.

----
## 0. Modify Epilogues in WPiOS.
- [x] Master issue to modify the Login Epilogue in WPiOS https://github.com/wordpress-mobile/WordPress-iOS/issues/13413
- [x] Master issue to modify the Signup Epilogue in WPiOS https://github.com/wordpress-mobile/WordPress-iOS/issues/13939

## 1. Remove known segues. Replace with navController.
- [x] remove .showLoginEmail #205
- [x] remove .showSignupMethod #225 
- [x] remove .showSigninV2 #232 
- [x] remove .showGoogle #219 
- [x] remove .showURLUsernamePassword #240 
- [x] remove .showWPUsernamePassword #245 
- [x] remove .showSelfHostedLogin  #242
- [x] remove .showWPComLogin #273
- [x] remove .startMagicLinkFlow #256 
- [x] remove .showMagicLink #254 
- [x] remove .showLinkMailView #211 
- [x] remove .show2FA #227 
- [x] remove .showDomains #260 
- [x] remove .showCreateSite #263 
- [x] remove .showSignupEmail #265 
- [x] remove .showUsernames #268
- [x] remove .showLoginMethod #270 
- [x] some segues were added manually (didn't use the enum) and need changes (these were discovered when testing a removed segue and a crash happened during the navigation attempt)

## 2. Unify the flows
- [ ] Google sign in #282
- [ ] Sign in with Apple
- [ ] WP.com login/signup
- [x] Site address #283 

## 3. Button bottom sheet UI changes
- [ ] Change "Log In" button text to "Log in with WordPress.com"
- [ ] Change that button's action to navigate to Get Started screen
- [ ] Remove the "3 button bottom sheet" from the "landing page"

PR #183, opened 2020-02-11
Remove .showEmailLogin segue

mindgraffiti:
Master issue #182 - this is a spike to test removing segues.

This PR removes all references to the segue `.showEmailLogin` and programmatically navigates the user to the `LoginEmailViewController` screen. There are 3 areas where the segue has been removed: WPiOS sign up, WPiOS sign in, and WCiOS sign in.

### To Test - WPiOS
1. Check out this branch
2. Check out this branch for WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/tree/try/break-up-storyboard and point it to your local pod by changing the podfile location to your local copy of WordPressAuthenticator-iOS
3. Build the Authenticator project
4. `rake dependencies` on WPiOS
5. Build the WPiOS project. You are now set up with local pod development! 
6. Navigate to and back from each of these flows: 

1) Login > Continue with WordPress.com - should land on the LoginEmailViewController page
2) Sign up for WordPress.com > Sign up with Email >  Enter an email address that has an existing WP.com account and tap Next - should land on the LoginEmailViewController page

### To Test - WCiOS
1. Check out this branch
2. In the WCiOS project, open the podfile, comment out Authenticator and replace with local pod development. It will look similar to this: `pod 'WordPressAuthenticator', :path => '/Users/thuycopeland/WordPressAuthenticator-iOS/'` 
3. Build the Authenticator project
4. `rake dependencies` on WCiOS
5. Build the WCiOS project. You are now set up with local pod development! 
6. Navigate to and back from this flow: Log in with Jetpack > should show the expected LoginEmailViewController

### To Test - Newspack
When checking out the `develop` branch on Newspack, I see the Site Address flow. I don't think this PR applies to Newspack. Pinging @aerych just in case.

### DIFF
--- a/WordPressAuthenticator/Extensions/UIStoryboard+Helpers.swift
+++ b/WordPressAuthenticator/Extensions/UIStoryboard+Helpers.swift
@@ -1,0 +1,19 @@
+import Foundation
+
+
+// MARK - Storyboard enum
+enum Storyboard: String {
+    case login
+
+    var instance: UIStoryboard {
+        return UIStoryboard(name: self.rawValue.capitalized, bundle: WordPressAuthenticator.bundle)
+    }
+
+    /// Returns a view controller from a Storyboard
+    /// assuming the identifier is the same as the class name.
+    ///
+    func instantiateViewController<T: NSObject>(ofClass classType: T.Type) -> T? {
+        let identifier = classType.classNameWithoutNamespaces
+        return instance.instantiateViewController(withIdentifier: identifier) as? T
+    }
+}
--- a/WordPressAuthenticator/Extensions/UIStoryboard+Helpers.swift
+++ b/WordPressAuthenticator/Extensions/UIStoryboard+Helpers.swift
@@ -4,6 +4,7 @@ import Foundation
 // MARK - Storyboard enum
 enum Storyboard: String {
     case login
+    case signup
 
     var instance: UIStoryboard {
         return UIStoryboard(name: self.rawValue.capitalized, bundle: WordPressAuthenticator.bundle)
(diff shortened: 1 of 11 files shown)


WordPressAuthenticator/Extensions/UIStoryboard+Helpers.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+
+// MARK - Storyboard enum
+enum Storyboard: String {
+    case login
+    case signup
+
+    var instance: UIStoryboard {
+        return UIStoryboard(name: self.rawValue.capitalized, bundle: WordPressAuthenticator.bundle)
```
ScoutHarris: `WordPressAuthenticator.swift` determines which bundle to use based on `If installed via CocoaPods`. This does not. Does it need to?

mindgraffiti: I'm confused. `WordPressAuthenticator.bundle` calls the [member class variable bundle](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/develop/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift#L439). The comment says it determines which bundle to use. Is calling the getter wrong? I don't understand bundles very well. Is this line supposed to be plain `bundle`?

ScoutHarris: 🤦‍♂ sorry. I misread what that was doing. Nothing to see here...

mindgraffiti: :blob-whew:

PR #193, opened 2020-03-04
Merges 1.10.7.1. back into develop.

diegoreymendez:
Just merges 1.10.7.1 back into develop.


diegoreymendez:
@jkmassel - As I have to propagate the fix we did yesterday to develop and to the 14.3, I'll just move ahead and merge.  The logic changes are the same we reviewed yesterday.

PR #200, opened 2020-03-12
Update version of NSURL+IDN to 0.4

Gio2018:
Also updates the shared scheme to Xcode 11

To test:

- run pod install and verify that the framework builds correctly


Gio2018:
Going to merge this one, as it's got an approval and there is a newer one that bumps the release (and updates one more dependency)

PR #212, opened 2020-03-18
Update to Gridicons 1.0

frosty:
This PR updates the authenticator library to use the new Gridicons 1.0 framework. I've updated the podspec and Podfile (which I'll further update once the pod is released), as well as switched over to the new style method calls.

**To Test**

Build and run the associated WordPress-iOS PR, and ensure that the auth flow loads and that some of the icons loaded here appear correctly: https://github.com/wordpress-mobile/WordPress-iOS/pull/13682

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -9,7 +9,7 @@ def wordpress_authenticator_pods
   ## Automattic libraries
   ## ====================
   ##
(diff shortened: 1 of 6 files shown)


frosty:
Thanks for the reminder!

PR #221, opened 2020-03-24
Bump the podspec to use the correct versioning

mindgraffiti:
Force-change the podfile just because.


mindgraffiti:
Thank you Stephenie!! 💟

PR #225, opened 2020-03-25
UL&S: Remove `.showSignupMethod`

### LINKED ISSUE #224: UL&S: Remove `.showSignupMethod`
Remove the `.showSignupMethod` storyboard segue from the storyboard and view controllers. Replace it with programmatic navigation.

Ref. #182

- [x] LoginEmailViewController
- [x] LoginPrologueViewController
- [x] enum .showSignupMethod
- [x] Login.storyboard - 2 references

mindgraffiti:
Fixes #224 
Ref. #182 

This PR removes all references to the segue `.showSignupMethod` and programmatically navigates the user to the `LoginPrologueSignupMethodViewController`. There are 2 areas where the segues have been removed:

1. WPiOS sign up
2. WPiOS sign in

### To Test - WPiOS
1. Check out this branch
2. `rake dependencies` and ensure there are no errors
3. Build and run (there should be no errors)
4. Visit the WPiOS PR to run the changes: https://github.com/wordpress-mobile/WordPress-iOS/pull/13733

### DIFF
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -142,7 +142,7 @@ class LoginPrologueViewController: LoginViewController {
         vc.dismissBlock = dismissBlock
         vc.transitioningDelegate = self
         vc.modalPresentationStyle = .custom
-        navigationController?.pushViewController(vc, animated: true)
+        navigationController?.present(vc, animated: true, completion: nil)
     }
 
     private func appleTapped() {
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -123,10 +123,18 @@ class LoginPrologueViewController: LoginViewController {
     }
 
     private func signupTapped() {
-        // This stat is part of a funnel that provides critical information.  Before
-        // making ANY modification to this stat please refer to: p4qSXL-35X-p2
+        // This stat is part of a funnel that provides critical information.
+        // Before making ANY modification to this stat please refer to: p4qSXL-35X-p2
         WordPressAuthenticator.track(.signupButtonTapped)
-        performSegue(withIdentifier: .showSignupMethod, sender: self)
+
+        guard let vc = LoginPrologueSignupMethodViewController.instantiate(from: .login) else {
+            DDLogError("Failed to navigate to LoginPrologueSignupMethodViewController")
+            return
+        }
+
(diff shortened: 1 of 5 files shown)


WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -129,10 +126,25 @@ class LoginPrologueViewController: LoginViewController {
...
+        vc.loginFields = self.loginFields
+        vc.dismissBlock = dismissBlock
+        vc.transitioningDelegate = self
+        vc.modalPresentationStyle = .custom
+
+        vc.emailTapped = { [weak self] in
+            self?.performSegue(withIdentifier: .showSigninV2, sender: self)
+        }
+
+        navigationController?.present(vc, animated: true, completion: nil)
```
mindgraffiti: 1. This pattern (or one similar to it) is going to be all over the codebase. It's starting to look like a code smell. Any opinions on what I should do differently? Or is it too early to consider that?

2. I'm not sure where the best place is, to put the emailTapped callback.

ScoutHarris: > This pattern (or one similar to it) is going to be all over the codebase

I'm not quite clear on what "this pattern" is - assigning all the properties to the VC, or just the `tapped` callbacks?

> Or is it too early to consider that?

In either case, I think this. Especially for the taps. Mainly because we still have separate `LoginPrologueSignupMethodViewController` and `LoginPrologueLoginMethodViewController`. In addition, during certain paths there are alternative paths accessible (iirc, they're for Woo, like accessing Google from Site Address), which makes these `tapped` callbacks more prevalent.

Once we start consolidating these paths, we'll have a "higher level" VC where these taps could live, or at least less VCs that need them.

Regarding the assigning all the properties to the VC, once we start consolidating it'll become clear if we can move these to a more central VC, and what that should be.

My 2 cents - right now just retain the current model, as ugly as it's going to be, mainly to reduce complexity at this stage, and we can certainly clean it up later.

Thoughts?

mindgraffiti: Pattern as in assigning all the properties to the VC. I'm guessing there isn't really much to be done about that.

> My 2 cents - right now just retain the current model, as ugly as it's going to be, mainly to reduce complexity at this stage, and we can certainly clean it up later.

Cool. That's how I feel about it too!

mindgraffiti:
✅Fixed: Continue with Google - tapping the button does nothing
✅Fixed: Continue with Apple - tapping the button does nothing

PR #243, opened 2020-04-16
Updating WPKit version

ScoutHarris:
With https://github.com/wordpress-mobile/WordPressKit-iOS/pull/241 the WPKit pod version was incremented. WPiOS failed to build because WPAuth depends on WPKit as well. So this simply updates WPKit pod version to make everyone happy!

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/13917

Note the `Validate Podspec` test will continue to fail until there is an official WPKit release. I'll update that before merging. But you should be able to test locally successfully.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -11,8 +11,8 @@ def wordpress_authenticator_pods
   ##
   pod 'Gridicons', '~> 1.0'
   pod 'WordPressUI', '~> 1.5.2'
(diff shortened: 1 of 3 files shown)


ScoutHarris:
> Don't forget to bump the podspec

Well I surely did forget didn't I? @mindgraffiti thanks for that!

PR #285, opened 2020-05-22
UL&S: Add a feature flag for unified site address login

mindgraffiti:
Add a feature flag that will enable the unified site address login flow.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
@@ -95,6 +95,6 @@ public struct WordPressAuthenticatorConfiguration {
         self.showLoginOptions = showLoginOptions
         self.enableSignInWithApple = enableSignInWithApple
         self.enableUnifiedAuth = enableUnifiedAuth
-        self.enableUnifiedSiteAddress = enableUnifiedSiteAddress
+        self.enableUnifiedSiteAddress = enableUnifiedAuth && enableUnifiedSiteAddress
     }
 }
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
```diff
@@ -55,9 +55,15 @@ public struct WordPressAuthenticatorConfiguration {
     ///
     let enableSignInWithApple: Bool
 
-    /// Flag indicating if the unified login/signup flow should be displayed.
+    /// Flag for the unified login/signup flows.
+    /// If disabled, none of the unified flows will display.
+    /// If enabled, allows selected unified flows to display.
```
ScoutHarris, line 60: Thanks for beefing this up!

WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
```diff
@@ -88,5 +95,6 @@ public struct WordPressAuthenticatorConfiguration {
         self.showLoginOptions = showLoginOptions
         self.enableSignInWithApple = enableSignInWithApple
         self.enableUnifiedAuth = enableUnifiedAuth
+        self.enableUnifiedSiteAddress = enableUnifiedSiteAddress
```
ScoutHarris, line 98: I don't know how/where you planned on checking both `enableUnifiedAuth` and `enableUnifiedSiteAddress`, but what do you think about making this:
`self.enableUnifiedSiteAddress = enableUnifiedAuth && enableUnifiedSiteAddress`

So `enableUnifiedSiteAddress` in only `true` if both are `true`?

mindgraffiti: Oh I like it! I'll change it to this.

Issue #282, opened 2020-05-19
ULS: Unify Google


ScoutHarris:
Unify Google Signup and Login flows.

Refs:
- ULS [Master Issue](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/182)
- ULS [Designs](https://wpmobilep2.wordpress.com/2020/04/24/unified-login-signup-for-wordpress-i3/)

Tasks:
- [ ] Create `GoogleAuthenticator` to contain:
  - [ ] Existing login logic.
  - [ ] Existing signup logic.
- [ ] Update logic to direct user to correct flow (without failing, as it does now).
- [ ] Show `Waiting for Google` view for both Login and Signup.
- [ ] Update `Waiting for Google` view to match new designs. (note there is no Zeplin design for this)

The Google flows utilize these views (to be considered before updating the Google view).
- `LoginWPComViewController` - for WP account login.
- `Login2FAViewController` - for 2FA code entry.
- `SignupEmailViewController` - when redirected from Login to Signup. It's not used currently, but will be in the new flow.

---
`Continue with Google` design from Zeplin:

![Continue with Google](https://user-images.githubusercontent.com/1816888/82370035-210ff880-99d5-11ea-9af6-2b4e57dedf0e.png)

---
Current `Waiting for Google` view:

![Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-19 at 13 38 17](https://user-images.githubusercontent.com/1816888/82370640-1ace4c00-99d6-11ea-8e6f-8da9595406a8.png)

diegoreymendez:
Marked tracks as complete for Google Sign In.

PR #287, opened 2020-05-27
Refactor Google signup logic

ScoutHarris:
Ref #282 

This moves the Google Signup logic out of `SignupGoogleViewController` and into `GoogleAuthenticator`. `SignupGoogleViewController` is notified of Google's progress via `GoogleAuthenticatorSignupDelegate` methods.

Regarding the `GoogleAuthenticator` delegates:

Because `AuthenticatorCredentials` is a Swift `struct`, it isn't visible to ObjC. Therefore, I couldn't use one delegate protocol with optional methods. So I created two protocols, containing their own required methods. `GoogleAuthenticatorDelegate` has become `GoogleAuthenticatorLoginDelegate` and `GoogleAuthenticatorSignupDelegate`.

And an honorable mention:

The `SVProgressHUD` that is displayed after selecting a Google account always said `Completing Signup`. That was misleading (and a bit scary) when signup was redirected to login. So I've changed it to the more generic `Processing Account`.

Can tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14196

### DIFF
--- a/WordPressAuthenticator/Signin/GoogleAuthenticator.swift
+++ b/WordPressAuthenticator/Signin/GoogleAuthenticator.swift
@@ -29,12 +29,14 @@ protocol GoogleAuthenticatorSignupDelegate {
     // Google account signup was successful.
     func googleFinishedSignup(credentials: AuthenticatorCredentials, loginFields: LoginFields)
 
-    // Google account login was successful.
+    // Google account signup redirected to login was successful.
     func googleLoggedInInstead(credentials: AuthenticatorCredentials, loginFields: LoginFields)
 
     // Google account signup failed.
     func googleSignupFailed(error: Error, loginFields: LoginFields)
 
+    // Google account signup cancelled by user.
+    func googleSignupCancelled()
 }
 
 class GoogleAuthenticator: NSObject {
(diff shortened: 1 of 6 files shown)


WordPressAuthenticator/Signin/GoogleAuthenticator.swift
```diff
@@ -78,14 +118,20 @@ private extension GoogleAuthenticator {
...
+        var trackProperties = properties
+        trackProperties["source"] = "google"
+        WordPressAuthenticator.track(event, properties: trackProperties)
+    }
+    
     enum LocalizedText {
         static let googleConnected = NSLocalizedString("Connected But…", comment: "Title shown when a user logs in with Google but no matching WordPress.com account is found")
         static let googleConnectedError = NSLocalizedString("The Google account \"%@\" doesn't match any account on WordPress.com", comment: "Description shown when a user logs in with Google but no matching WordPress.com account is found")
         static let googleUnableToConnect = NSLocalizedString("Unable To Connect", comment: "Shown when a user logs in with Google but it subsequently fails to work as login to WordPress.com")
+        static let processing = NSLocalizedString("Processing Account", comment: "Shown while the app waits for the account process to complete.")
```
mindgraffiti, line 133: Like the new phrasing. Calmer UX compared to before :)

ScoutHarris:
Thanks @mindgraffiti !

PR #290, opened 2020-05-29
UL&S: Unified Site Address

mindgraffiti:
This PR uses the SiteAddress feature flag that was enabled previously. Log in > "entering your site address" will navigate to the new `SiteAddressViewController`. The `SiteAddressViewController` inherits from the `AuthBaseViewController`, which is a `UIViewController` that contains an instance of a `UITableView`.


mindgraffiti:
I know which way to go now! Thanks @ScoutHarris 🌟

Issue #283, opened 2020-05-20
UL&S: Site Address flow


mindgraffiti:
#### Base UI

Create the re-usable views that will be needed in the Site Address flow.

- [x] Create a flag to enable just the unified Site Address flow #285 
- [x] Create a new blank viewController, `SiteAddressViewController` and navigate to it #293
- [x] Inside the viewController, add a tableview. Set autolayout leading / trailing margins to zero #294 
- [x] Spike: @mattmiklic requests the continue button to attach to the keyboard for regular size classes and scroll for compact size classes. Research feasibility, estimations, and impact compared to pinning the button to the bottom of the screen. #296
- [x] Create the first uitableviewcell - InstructionTableViewCell #305 
- [x] Add basic TextFieldTableViewCell #306 
- [x] Add basic TextLinkButtonTableViewCell #308  
(and renamed InstructionTableViewCell to TextLabelTableViewCell) 
- [x] Fix dynamic text bug in TextLinkButtonTableViewCell #310 
- [x] Add basic ErrorLabelTableViewCell #312 
- [x] Add support for iPad layouts. https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/334

#### Pages
- [x] SiteAddressViewController #312 
- [x] SiteCredentialsViewController #322

mindgraffiti:
Completed after #334 was merged.

mindgraffiti:
Re-opening because this needs Tracks

diegoreymendez:
I updated the description above to include a link to the tracks issue.

PR #294, opened 2020-06-01
UL&S: Site Address tableView and continue button

mindgraffiti:
Ref. #283 
Ref. tk

This PR removes the temporary label, adds the tableView, and adds the continue button.

#### To test
1. Check out this branch
2. `bundle exec pod install`
3. build (with no errors)
4. Check out the referenced PR to test changes: tk

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/SiteAddress.storyboard
+++ b/WordPressAuthenticator/Unified Auth/View Related/SiteAddress.storyboard
@@ -37,6 +37,10 @@
                         </constraints>
                         <viewLayoutGuide key="safeArea" id="ihD-pY-rg9"/>
                     </view>
+                    <connections>
+                        <outlet property="continueButton" destination="ClH-Cn-49d" id="MQY-BY-P8D"/>
+                        <outlet property="tableView" destination="KLl-Uz-wEP" id="ntt-cX-m20"/>
+                    </connections>
                 </viewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="Ipm-G3-kY7" userLabel="First Responder" customClass="UIResponder" sceneMemberID="firstResponder"/>
             </objects>
--- a/WordPressAuthenticator/Unified Auth/View Related/SiteAddress.storyboard
+++ b/WordPressAuthenticator/Unified Auth/View Related/SiteAddress.storyboard
@@ -16,15 +16,25 @@
                         <rect key="frame" x="0.0" y="0.0" width="375" height="667"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
```diff
@@ -12,19 +12,34 @@ public struct WordPressAuthenticatorDisplayStrings {
...
     ///
-    public init(emailLoginInstructions: String, jetpackLoginInstructions: String, siteLoginInstructions: String) {
+    public init(emailLoginInstructions: String,
+                jetpackLoginInstructions: String,
+                siteLoginInstructions: String,
+                siteAddressPrimaryButton: String) {
         self.emailLoginInstructions = emailLoginInstructions
         self.jetpackLoginInstructions = jetpackLoginInstructions
         self.siteLoginInstructions = siteLoginInstructions
+        self.siteAddressPrimaryButton = siteAddressPrimaryButton
```
ScoutHarris, line 28: This seems to indicate we're going to have a string for every button on every view. That seems a bit excessive. Does the button text need to be customized? And to this level of granularity? It looks like host apps will typically use the same text for the same actions. 

If we do need to customize the button text, I'm wondering if maybe we can have more generic strings, like `continue`, `done`, `openMail` , i.e. customize the text for the action instead of the view.

What do you think?

mindgraffiti: I'm overthinking it. More generic names for strings sounds better to me!

mindgraffiti:
@Gio2018 `WordPressauthenticatorDisplayStrings` needs to be public because other host apps set their own strings. [Here's an example in WCiOS](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L51). Unfortunately when I created the `WordPressauthenticatorDisplayStrings` struct, I made it opinionated by setting WPiOS strings as the default. I don't know how to un-do that and build it better, but I'm definitely open to suggestions.

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
```diff
@@ -12,19 +12,34 @@ public struct WordPressAuthenticatorDisplayStrings {
...
+        return WordPressAuthenticatorDisplayStrings(
+            emailLoginInstructions: NSLocalizedString("Log in to your WordPress.com account with your email address.",
+                                                      comment: "Instruction text on the login's email address screen."),
+            jetpackLoginInstructions: NSLocalizedString("Log in to the WordPress.com account you used to connect Jetpack.",
+                                                        comment: "Instruction text on the login's email address screen."),
+            siteLoginInstructions: NSLocalizedString("Enter the address of the WordPress site you'd like to connect.",
+                                                     comment: "Instruction text on the login's site addresss screen."),
+            continueButtonTitle: NSLocalizedString("Continue",
+                                                    comment: "The primary call-to-action button text on the unified site address screen.")
```
ScoutHarris, line 42: This comment should probably be generic.

Gio2018:
> @Gio2018 `WordPressauthenticatorDisplayStrings` needs to be public because other host apps set their own strings.

Ok then, I was missing something 🙂

ScoutHarris:
Oh, I just noticed the podspec version wasn't update. Probably should be `1.18.0-beta.3`.

mindgraffiti:
Thank you for the reviews @Gio2018 and @ScoutHarris!

PR #295, opened 2020-06-02
ULS: Unify Google auth flows

ScoutHarris:
To accommodate unified Google auth flow:

- There is now a `GoogleAuthenticatorDelegate` that encompasses all delegate methods needed for the unified flow. 
  - This is only used by `GoogleAuthViewController` (the VC for the unified flow).
  - `GoogleAuthenticatorLoginDelegate` and `GoogleAuthenticatorSignupDelegate` remain to support the old separate flows (`LoginPrologueViewController`, `LoginEmailViewController`, `SignupGoogleViewController`) .
- Bonus: The `Processing Account` HUD is now shown in the login flow (previously it was only shown in signup).

With this change, the Google flows are unified. Since `GoogleAuthViewController` is presented from both login and signup, they both use the same flow. 🎉 

There is one path that is not unified yet. 

After a Google account is selected, `GoogleAuthenticator` will first attempt to login. If a WP account does not exist, it does not redirect to signup as there is no logic currently to do so. (That's the point of this project after all, innit? 😄 ) Instead, this view is displayed. This will be addressed in a future PR. Stay tuned!

login_with_new_acct

### DIFF
--- a/WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
+++ b/WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
@@ -195,8 +195,10 @@ extension GoogleAuthenticator: GIDSignInDelegate {
                     track(.signupSocialButtonFailure, properties: properties)
                 }
 
-                // Notify the signupDelegate so the Google Signup view can be dismissed.
+                // Notify the delegates so the Google Auth view can be dismissed.
                 signupDelegate?.googleSignupCancelled()
+                delegate?.googleAuthCancelled()
+                
                 return
         }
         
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
```diff
@@ -47,6 +84,7 @@ class GoogleAuthenticator: NSObject {
     private override init() {}
     var loginDelegate: GoogleAuthenticatorLoginDelegate?
     var signupDelegate: GoogleAuthenticatorSignupDelegate?
+    var delegate: GoogleAuthenticatorDelegate?
```
Gio2018, line 87: This property should be weak (and the related protocol, class bound), otherwise it looks like an instance of `GoogleAuthViewController` is retained after the login completes. Here's a screenshot of the memory graph showing that.

Screen Shot 2020-06-03 at 12 04 41 PM

Note that when I tweaked that var to weak, this did not happen. It maybe worth looking in the two existing delegates as well

PR #299, opened 2020-06-04
UL&S: SiteAddress: anchor Continue button to keyboard

mindgraffiti:
Ref. #283 
Ref. pbArwn-Al-p2

This PR takes the things learned from #296 and applies them to the current Continue button. Note the textfield is a temporary prototype cell so that the presentation of the keyboard and the animation for the continue button can be observed. 

### To test
1. `bundle exec pod install`
2. build (with no errors)
3. Test PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14248


mindgraffiti:
Removing reviewers. Just noticed my autolayout constraints for the Continue button aren't right on all screens.

PR #300, opened 2020-06-05
ULS: Create Google account

ScoutHarris:
Ref: #282 
Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14258

When `Next` is tapped on the Google Signup Confirmation view, account creation is now attempted. If successful, the Signup Epilogue will be displayed.

On the off chance a user goes off somewhere else and creates a WP account with the selected Google account _while this view is still displayed_, the account will be logged in instead and the Login Epilogue displayed. _Unless_ the user has:
- Enabled 2FA on the WP account.
- Disconnected Google on the WP account.

In these cases, an error alert will be displayed. These should be edge cases, so not really worth the effort in attempting to handle them.

Side note: the way I implemented `GoogleAuthenticatorDelegate` didn't quite work out as planned, which resulted in several no-op delegate methods. I'll look at refactoring that separately.

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Google/GoogleSignupConfirmationViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Google/GoogleSignupConfirmationViewController.swift
@@ -46,13 +46,62 @@ private extension GoogleSignupConfirmationViewController {
         submitButton?.accessibilityIdentifier = "Google Signup Email Next Button"
     }
     
-    // MARK: - Button Action Handling
+    // MARK: - Button Handling
 
     @IBAction func handleSubmit() {
-        displayError(message: "Cheers!")
-        // TODO: create account.
-        // TODO: re-enable this when handling added.
-        // configureSubmitButton(animating: true)
+        configureSubmitButton(animating: true)
+        GoogleAuthenticator.sharedInstance.delegate = self
+        GoogleAuthenticator.sharedInstance.createGoogleAccount(loginFields: loginFields)
+    }
+
+}
+
+// MARK: - GoogleAuthenticatorDelegate
+
+extension GoogleSignupConfirmationViewController: GoogleAuthenticatorDelegate {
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Unified Auth/View Related/Google/GoogleSignupConfirmationViewController.swift
```diff
@@ -46,13 +46,62 @@ private extension GoogleSignupConfirmationViewController {
...
+        // Display generic inline error.
+        displayError(message: NSLocalizedString("Google sign up failed.", comment: "Message shown on screen after the Google sign up process failed."))
+        // Display the API error in a Fancy Alert.
+        displayError(error as NSError, sourceTag: .wpComSignup)
+    }
+    
+    // MARK: - Login
+
+    func googleFinishedLogin(credentials: AuthenticatorCredentials, loginFields: LoginFields) {
+        // Here for protocol compliance.
```
Gio2018, line 88: Would it be applicable to break the protocol into two (Login and Signup perhaps), in order to avoid empty methods? Maybe it's not as I don't have the full context. Certainly can be done in a separate PR, if ever.

ScoutHarris: Yea, like I said in my note:

> the way I implemented GoogleAuthenticatorDelegate didn't quite work out as planned, which resulted in several no-op delegate methods. I'll look at refactoring that separately.

I had initially thought `GoogleAuthenticatorLoginDelegate` and `GoogleAuthenticatorSignupDelegate` would only be needed for the original separate flows, but the "unified" `GoogleAuthenticatorDelegate` methods ended up being separated across VCs anyway. So I'll probably fallback to having just the 2 (login and signup). I'm not sure yet, which is why I'm going to to it separately.

PR #304, opened 2020-06-15
Relax Lottie version constraint to allow host app updates

jkmassel:
Allow the host app to use any version of Lottie compatible with this library without needing to update this repo.


jkmassel:
Updates seem to break the animations, so I'm closing this for now.

PR #306, opened 2020-06-17
UL&S: Add TextFieldTableViewCell to SiteAddressViewController

mindgraffiti:
Ref. #283 

This PR adds a basic textfield inside a tableview cell. The cell should look like the Zeplin designs. This PR also abstracts out the row presentation.

### To test
1. `bundle exec pod install`
2. Build and run
3. Check out the testing PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14340

Screen Shot 2020-06-17 at 11 29 01 AM

**iPhone 11 Pro screenshots**
<a href="https://user-images.githubusercontent.com/1062444/85311821-9e6ad500-b47b-11ea-9c16-e289ca1d524c.png"></a>

<a href="https://user-images.githubusercontent.com/1062444/85311854-aaef2d80-b47b-11ea-832b-1e7770c125f4.png"></a>

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
@@ -13,6 +13,7 @@ class TextFieldTableViewCell: UITableViewCell {
     override func awakeFromNib() {
         super.awakeFromNib()
 
-        borderView.backgroundColor = WordPressAuthenticator.shared.unifiedStyle?.borderColor
+        let borderColor = WordPressAuthenticator.shared.unifiedStyle?.borderColor ?? WordPressAuthenticator.shared.style.primaryNormalBorderColor
+        borderView.backgroundColor = borderColor
     }
 }
--- a/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
@@ -12,7 +12,8 @@ class TextFieldTableViewCell: UITableViewCell {
 
     override func awakeFromNib() {
         super.awakeFromNib()
-        // Initialization code
+
+        borderView.backgroundColor = WordPressAuthenticator.shared.unifiedStyle?.borderColor
     }
 
     override func setSelected(_ selected: Bool, animated: Bool) {
--- a/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
@@ -7,7 +7,12 @@ class TextFieldTableViewCell: UITableViewCell {
 
     public static let reuseIdentifier = "TextFieldTableViewCell"
 
+    private var hairlineBorderWidth: CGFloat {
+        return 1.0 / UIScreen.main.scale
+    }
+
     @IBOutlet var borderView: UIView!
+    @IBOutlet var borderWidth: NSLayoutConstraint!
     @IBOutlet var textField: UITextField!
 
     override func awakeFromNib() {
@@ -15,5 +20,6 @@ class TextFieldTableViewCell: UITableViewCell {
 
         let borderColor = WordPressAuthenticator.shared.unifiedStyle?.borderColor ?? WordPressAuthenticator.shared.style.primaryNormalBorderColor
         borderView.backgroundColor = borderColor
+        borderWidth.constant = hairlineBorderWidth
     }
 }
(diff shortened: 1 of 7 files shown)


mindgraffiti:
@mattmiklic this one needs design review.

1. The border line in Zeplin shows a 1 pixel view with rgba(60 60 60, 0.29), which doesn't match the muriel color set for dividers and tablecell separators. Is the one I have above (.divider) correct?
2. The spacing between table cells in Zeplin isn't selectable. I can see there is extra space between the label and the textfield but I don't know how much. What's rendered above is a minimum height of 44 points for the 2 cells. How much top / bottom margin should be between the 2 items?

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -97,3 +135,23 @@ extension SiteAddressViewController: NUXKeyboardResponder {
         keyboardWillHide(notification)
     }
 }
+
+
+// MARK: - Constants
+extension SiteAddressViewController {
+    /// Rows listed in the order they were created
+    ///
+    enum Row {
```
mindgraffiti, line 144: There is only 1 section for each screen so I left Sections out of the enum. I can add it here if it helps.

ScoutHarris:
> The border line in Zeplin shows a 1 pixel view with rgba(60 60 60, 0.29), which doesn't match the muriel color set for dividers and tablecell separators. Is the one I have above (.divider) correct?

FWIW, in WPiOS we use `hairlineBorderWidth`, which is:
```
    static var hairlineBorderWidth: CGFloat {
        return 1.0 / UIScreen.main.scale
    }
```

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
ScoutHarris, line 144: This should be `private`. Also, you could just make the extension `private` and move all the methods in here instead of having to explicitly mark everything `private`.

ScoutHarris: I don't think that's necessary.

WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
```diff
@@ -0,0 +1,18 @@
...
+
+    public static let reuseIdentifier = "TextFieldTableViewCell"
+
+    @IBOutlet var borderView: UIView!
+    @IBOutlet var textField: UITextField!
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+
+        borderView.backgroundColor = WordPressAuthenticator.shared.unifiedStyle?.borderColor
```
ScoutHarris, line 16: What happens if `unifiedStyle` is undefined (since it optional)? Maybe it should fallback to `WordPressAuthenticator.shared.style.primaryNormalBorderColor`?

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -65,18 +75,46 @@ final class SiteAddressViewController: LoginViewController {
...
+    ///
+    func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
+        switch cell {
+        case let cell as InstructionTableViewCell:
+            configureInstruction(cell)
+        case let cell as TextFieldTableViewCell:
+            configureTextField(cell)
+        default:
+            // TODO: return a blank cell when this is ready for review
+            fatalError("Error: Did you forget to configure a custom cell?")
```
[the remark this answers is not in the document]
mindgraffiti: 🤦‍♀️  my bad

mindgraffiti:
Moving design review to the WPiOS repo where there is a "Needs design review" label.

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -97,3 +82,79 @@ extension SiteAddressViewController: NUXKeyboardResponder {
...
+    /// Configure cells
+    ///
+    func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
+        switch cell {
+        case let cell as InstructionTableViewCell:
+            configureInstruction(cell)
+        case let cell as TextFieldTableViewCell:
+            configureTextField(cell)
+        default:
+            DDLogError("Error: Unidentified tableViewCell type found.")
```
mindgraffiti, line 127: I can't return UITableViewCell() here because the methods don't ask for one, but the default case can't be empty, so I settled on logging an error message.

ScoutHarris:
I just have one request, when you can, not necessarily for this PR. Can we auto-select the text field when the view appears? So the user can just start typing without having to tap the field? Personally, I find this a nice touch. I'd like to do this for all the views going forward if we could.

mindgraffiti:
>Can we auto-select the text field when the view appears?

Yep! I will incorporate some cleanup to my table cell classes and refine the textfield so it's the first responder in the next PR. Now that I can see a pattern I can tell there are refinements that need made.

PR #307, opened 2020-06-17
ULS: navigation bar large titles

ScoutHarris:
Ref: #182 

This adds the ability to show Large Titles in the navigation bar.
- By default, `prefersLargeTitles` is `false`.
- Once a VC calls `setLargeTitleDisplayMode`, large titles are enabled and the display mode set.
- The navigation bar displays the title set in the VC.
- The titles are defined in and obtained from `WordPressAuthenticatorDisplayStrings`.

To note, the titles are not styled at all, and thus will be displayed using the default style for now.

Since we don't have a full unified flow yet, I've added some example usage to the email login flow for testing purposes. I'll remove that before merging.

Can be tested with WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/14344

### DIFF
--- a/WordPressAuthenticator/Signin/LoginLinkRequestViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginLinkRequestViewController.swift
@@ -24,10 +24,6 @@ class LoginLinkRequestViewController: LoginViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        // TODO: remove before merging
-        navigationItem.title = WordPressAuthenticator.shared.displayStrings.logInTitle
-        setLargeTitleDisplayMode(.automatic)
-
         localizeControls()
         configureUsePasswordButton()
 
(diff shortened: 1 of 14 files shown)


WordPressAuthenticator/Signin/LoginLinkRequestViewController.swift
```diff
@@ -24,6 +24,10 @@ class LoginLinkRequestViewController: LoginViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        // TODO: remove before merging
+        navigationItem.title = WordPressAuthenticator.shared.displayStrings.logInTitle
```
mindgraffiti, line 28: What happens to the large title if an app puts in a blank string or nil?

ScoutHarris: If large titles are enabled but no title is set, then the large title area is displayed with no title:

no_title

PR #308, opened 2020-06-23
UL&S: Add TextLinkTableViewCell to SiteAddressViewController

mindgraffiti:
Ref. #283 

This PR refines the code style to be consistent among the new table cells and creates another base UI piece, TextLinkTableViewCell, which is a plain button with text. If only one property is needed on a cell, the variable is made public and the value can be assigned. For more than one property on a cell, a `configuration` method is created so the different parameters can be passed in.

Concerns: the naming isn't consistent. The thing is a hint button according to the user, but a plain button in interface builder. `TextLinkButtonTableViewCell`? `HintButtonTableViewCell`? 

### To test
1. `bundle exec pod install`
2. Build and run
3. Test on PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14372


WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -1,28 +1,32 @@
...
+    /// Private properties
+    ///
     @IBOutlet private weak var tableView: UITableView!
     @IBOutlet var bottomContentConstraint: NSLayoutConstraint?
 
     // Required property declaration for `NUXKeyboardResponder` but unused here.
     var verticalCenterConstraint: NSLayoutConstraint?
 
-    var displayStrings: WordPressAuthenticatorDisplayStrings {
+    private var displayStrings: WordPressAuthenticatorDisplayStrings {
```
Gio2018, line 17: this won't really be private as the singleton isn't. I wonder if we could just reference directly to the property of the singleton since it's used in just a couple of places?

ScoutHarris:
> Concerns: the naming isn't consistent. The thing is a hint button according to the user, but a plain button in interface builder. TextLinkButtonTableViewCell? HintButtonTableViewCell?

My 2 cents - `TextLinkButtonTableViewCell` is more generic and usage agnostic. Yes, we're using it for hints, but who's to say it can't/won't be used for something else? I think naming it what it is, not what it's used for, is less confusing.

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
ScoutHarris: Just another option - if you're goal is to just have a shorter name, you could use `typealias`. 

`typealias DisplayStrings = WordPressAuthenticator.shared.displayStrings`

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -85,11 +89,11 @@ extension SiteAddressViewController: NUXKeyboardResponder {
     }
 
     func showKeyboard() {
-        tableView.firstSubview(ofType: UITextField.self)?.becomeFirstResponder()
+        firstTextField?.becomeFirstResponder()
```
ScoutHarris, line 92: Thank you!

ScoutHarris:
> If only one property is needed on a cell, the variable is made public and the value can be assigned. For more than one property on a cell, a configuration method is created so the different parameters can be passed in.

I'm a bit torn about this as it's inconsistent and may not be clear. But I'm willing to wait and see how it shakes out.

ScoutHarris:
I just noticed - the text doesn't respect text size changes. Which made me remember we don't use button labels to display text in this manner because button label text sizes don't change. I think we're going to need a different solution that involves text outside a button.

Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-06-23 at 17 24 46

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
mindgraffiti: I forgot about using a typealias. Thanks!

mindgraffiti: `typealias` doesn't work with singletons. Error: `Static property 'shared' is not a member type of 'WordPressAuthenticator'`. I'll just reference the properties directly.

mindgraffiti:
@ScoutHarris ready for another review. For dynamic text on the button in the TextLinkButtonTableViewCell, I first tried using a text label in a row, then implementing the `didSelectRowAt:` delegate method. At first none of the touches were firing selected cells, so I added `.cancelsTouchesInViews` to the NUXBaseViewController, where a background view tap gesture was set. Using this approach created a lag in response times for the iPhone SE (2nd gen) sims. I'm not certain why.

Anyway, TIL that buttons can have dynamic text too. It's not a control that lives in InterfaceBuilder. It has to be set programmatically. It's working now. I also touched up the controls for VoiceOver, though I don't plan on asking reviewers to test that until the whole screen is ready to go.

<a href="https://user-images.githubusercontent.com/1062444/85613702-c430df00-b61f-11ea-9571-f775728a0dfc.png"></a>

<a href="https://user-images.githubusercontent.com/1062444/85613733-cbf08380-b61f-11ea-9f8a-fe693fd96854.png"></a>

Found a bug in NUXButton. When the dynamic text shrinks back down, the NUXButtons do not respond to the change. I'll file a bug for it in a bit.

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -123,31 +123,57 @@ private extension SiteAddressViewController {
     ///
     func loadRows() {
         rows = [.instructions, .siteAddress]
+        let displayHintButtons = WordPressAuthenticator.shared.configuration.displayHintButtons
+        if displayHintButtons {
```
ScoutHarris, line 127: Couldn't this just be `if WordPressAuthenticator.shared.configuration.displayHintButtons {`?

WordPressAuthenticator/Extensions/CollectionType+AuthHelpers.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+
+/// Collection Helpers for Authenticator.
+///
+extension Collection {
+    /// Returns the element at the specified index if it is within bounds, otherwise nil.
+    subscript (safe index: Index) -> Element? {
```
ScoutHarris, line 8: Maybe I missed something, but is this used by anything in this PR?

ScoutHarris:
Was confused about why `CollectionType+AuthHelpers.swift` was added. I'll finish the review once that question is answered.

WordPressAuthenticator/Extensions/CollectionType+AuthHelpers.swift
mindgraffiti: Artifact from trying out a label and setting didSelectRowAt. Will delete!

ScoutHarris:
I only see one issue. With large text on an iPhone (ex: 11 Pro Max), the hint text is truncated. I would expect it to wrap. 

iphone_large_text

I figure you might have the same issue with the error label, so I'd be ok with resolving this in a follow up PR. It's up to you, but I think we can consider this GTG assuming that will be fixed later.

mindgraffiti:
Find yo' address 😂 . Thanks for the review! I will make the button text wrap in the next PR.

PR #310, opened 2020-06-25
UL&S: bug fix: make the TextLinkButtonTableViewCell button word wrap, not truncate

mindgraffiti:
Ref #283 

Per #308, there is a bug where the button title text was truncating. This PR fixes the truncation.

PR to test fix: https://github.com/wordpress-mobile/WordPress-iOS/pull/14385

Before:
iphone_large_text

After:
<a href="https://user-images.githubusercontent.com/1062444/85781913-d11b0480-b6eb-11ea-9d3d-194a0c1d459b.png"></a>

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextLinkButtonTableViewCell.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextLinkButtonTableViewCell.swift
@@ -18,6 +18,14 @@ class TextLinkButtonTableViewCell: UITableViewCell {
 
     public var actionHandler: (() -> Void)?
 
+	override func awakeFromNib() {
+		super.awakeFromNib()
+
+		button.titleLabel?.font = UIFont.preferredFont(forTextStyle: .callout)
+		button.titleLabel?.adjustsFontForContentSizeCategory = true
+		button.titleLabel?.lineBreakMode = .byWordWrapping
+	}
+
     public func configureButton(text: String?) {
         button.setTitle(text, for: .normal)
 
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextLinkButtonTableViewCell.swift
```diff
@@ -18,15 +18,20 @@ class TextLinkButtonTableViewCell: UITableViewCell {
 
     public var actionHandler: (() -> Void)?
 
+	override func awakeFromNib() {
+		super.awakeFromNib()
+
+		button.titleLabel?.font = UIFont.preferredFont(forTextStyle: .callout)
+		button.titleLabel?.adjustsFontForContentSizeCategory = true
+		button.titleLabel?.lineBreakMode = .byWordWrapping
```
Gio2018, line 26: Just a thought here: since you're using a `xib`, it would be nice to set these at design time, although, for a `UIButton` is less obvious, since you don't have the 'adjustsFontForContentSizeCategory` readily available. You might still set the font and line break mode in there. You could also set `adjustsFontForContentSizeCategory` in there using `@ibdesignable` (which I don't think is worth) or trying to set that value as a user defined runtime attribute (but it would still be a runtime calculation). Again it's just a thought, there's nothing wrong in leaving as it is.

mindgraffiti: Okie dokie, I moved setting the font and lineBreakMode to the Interface Builder. `ajustsFontForContentSizeCategory` is a variable in the `UIContentSizeCategoryAdjusting` protocol. Protocol variables don't seem to work with @IBInspectable or @IBDesignable, so I left it as-is.

PR #312, opened 2020-06-26
UL&S: Error Label for SiteAddressViewController

mindgraffiti:
Ref. #283 
Testing PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14432

In this PR, error-checking is restored. The Error Label behavior inherited from the LoginViewController is connected to the UI found in the SiteAddressViewController, and error-checking happens after tapping the Continue button.

I did not include two of the old error handling states. These are:
1. A method frequently used, `refreshSiteAddressError(immediate:)`. It was effectively dead code. The `showError` bool will always return `false` because the method `!loginFields.validateSiteForSignin()` always returns `false`. I did get the method working correctly again, only to find out that the user experience with this method working correctly is terribly annoying. It validates the URL as you type, so an error message immediately displays.
2. The `handleTextFieldDidChange(_ sender:)` method no longer immediately wipes out displayed errors. The old behavior immediately deletes an error message after changing a single letter or deleting a single letter in the textfield. I tried incorporating this behavior and timeboxed it to 4 hours. Since the error label now lives in a tableViewCell, it reloads the table data and dismisses the keyboard after a single change to the URL, after an error displayed. That behavior was annoying, and I didn't get the correct state, UI display, nor the correct behavior after 4 hours, so I didn't commit that.


WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
```diff
@@ -66,6 +70,22 @@ private extension TextFieldTableViewCell {
 }
 
 
+// MARK: - UITextFieldDelegate conformance
+extension TextFieldTableViewCell: UITextFieldDelegate {
+    func textFieldDidChangeSelection(_ textField: UITextField) {
```
Gio2018, line 75: Just a thought here: `TextFieldTableViewCell` is the `UITextFieldDelegate`, but the delegate methods are actually executing closures that are defined in `SiteAddressViewController`. What if `SiteAddressViewController` was instead set as the 'UITextFieldDelegate` and implements these methods directly?

mindgraffiti: I was trying not to bloat the ViewController, but I agree. It's easier to understand and easier to use if the textfield delegate methods are the SiteAddressViewController's responsibility.

WordPressAuthenticator/Unified Auth/View Related/Reusable Views/TextFieldTableViewCell.swift
```diff
@@ -26,6 +25,10 @@ final class TextFieldTableViewCell: UITableViewCell {
         setCommonTextFieldStyles()
     }
 
+	/// Configures the textfield for URL, username, or entering a password.
+	/// - Parameter style: changes the textfield behavior and appearance.
+	/// - Parameter placeholder: the placeholder text, if any
+	///
```
ScoutHarris, line 31: 👍 !

WordPressAuthenticator/Unified Auth/View Related/SiteAddressViewController.swift
```diff
@@ -73,7 +131,14 @@ extension SiteAddressViewController: UITableViewDataSource {
 
 // MARK: - UITableViewDelegate conformance
 extension SiteAddressViewController: UITableViewDelegate {
-
+	/// After the site address textfield cell is done displaying, remove the textfield reference.
+	///
+	func tableView(_ tableView: UITableView, didEndDisplaying cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+		let row = rows[indexPath.row]
+		if row == .siteAddress {
```
ScoutHarris, line 138: Nitpick, but couldn't this just be `if rows[indexPath.row] == .siteAddress {`?

mindgraffiti:
Thanks for the reviews @Gio2018 and @ScoutHarris 🌟

Issue #315, opened 2020-07-06
ULS: Update Navigation and Status Bars to new design


ScoutHarris:
Update Navigation and Status Bars to new design.

Tasks:
- [x] Style navigation bar with large titles:
  - [x] Add the ability to show Large Titles in the navigation bar. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/307)
  - [x] Nav bar colors (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/311, https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/314)
  - [x] Update `Help` button text color. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/327)
  - [x] Use New York font for large titles. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/316)
  - [x] Remove host app icon. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/317)
  - [x] Remove large titles. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/326)
- [x] Style status bar. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/330)

Issues:
- [x] Old nav bar: `Back` button color is blue when logging in straight in with Apple account. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14386#issuecomment-650316300) (https://github.com/wordpress-mobile/WordPress-iOS/pull/14505)
- [x] New nav bar: when `Help` is launched, the nav bar style is incorrect. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14502)
- [x] Nav bars not switching from new > old > new correctly. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14432#pullrequestreview-444292895) (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/326)
- [x] New nav bar: table doesn't scroll. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/318)

---
Design from Zeplin:

[...]

PR #316, opened 2020-07-07
ULS: Use NY font for large titles

ScoutHarris:
Ref: #315 

This updates the navigation bar's large title font to use the New York font.

<kbd>auth</kbd>

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14433

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -12,7 +12,7 @@ def wordpress_authenticator_pods
   pod 'Gridicons', '~> 1.0'
   pod 'WordPressUI', '~> 1.7.0'
   pod 'WordPressKit', '~> 4.0-beta.0' # Don't change this until we hit 5.0 in WPKit
-  pod 'WordPressShared', '~> 1.0-beta.0' # Don't change this until we hit 2.0 in WPShared
+  pod 'WordPressShared', '~> 1.9-beta' # Don't change this until we hit 2.0 in WPShared
 
 
   ## Third party libraries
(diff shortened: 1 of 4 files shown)


Podfile
```diff
@@ -12,7 +12,7 @@ def wordpress_authenticator_pods
   pod 'Gridicons', '~> 1.0'
   pod 'WordPressUI', '~> 1.7.0'
   pod 'WordPressKit', '~> 4.0-beta.0' # Don't change this until we hit 5.0 in WPKit
-  pod 'WordPressShared', '~> 1.0-beta.0' # Don't change this until we hit 2.0 in WPShared
+  pod 'WordPressShared', '~> 1.9-beta' # Don't change this until we hit 2.0 in WPShared
```
mindgraffiti, line 15: ># Don't change this until we hit 2.0 in WPShared

What is that about?

ScoutHarris: I believe it was to relax the pod inter-dependency restrictions. However, the Shared method is in version `1.9.1`, so I bumped Shared to the lowest possible version so I could use the `serifFontForTextStyle` method.

leandroalonso:
@mindgraffiti @ScoutHarris this comment was made by @jkmassel.

If I'm correct, the idea is that we wouldn't need to change the `Podfile` anymore to update the versions (avoiding bumping them all the time as we do).

PR #328, opened 2020-07-19
Authenticator unit tests

charliescheer:
This PR adds unit test coverage for the WPAuth files that appear in the Credentials and Authenticator folders.  
Note: This covers much of what is in those folders, am continuing to add more tests to theses same sections going forward

To review: 
Check out the following files for new tests and review them for usefulness and completeness of the testing coverage added
WordPressAuthenticatorTests.swift
WordPressSourceTagTests.swift
CredentialsTests.swift

There have been several mock providers and spies created to make these tests work, and it would be great to get some feedback on those and have them reviewed for style, completeness, and simplicity:
MockWordpressAuthenticatorProvider.swift
WordPressAuthenticatorDelegateSpy.swift
ModalViewControllerPresentingSpy.swift
MockURLHandler.swift
MockOnePasswordFacade.swift


WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    func testHandleWordPressAuthURLReturnsTrueOnSucceed() {
+        let url = URL(string: "https://wordpress.com/wp-login.php?token=1234567890%26action&magic-login&sr=1&signature=1234567890oienhdtsra")
+        
+        XCTAssertTrue(WordPressAuthenticator.shared.handleWordPressAuthUrl(url!, allowWordPressComAuth: true, rootViewController: UIViewController()))
+    }
+    
+    func testSignInForWPOrgReturnsVC() {
+        let vc = WordPressAuthenticator.signinForWPOrg()
+        
+        XCTAssertTrue((vc as Any) is LoginSiteAddressViewController)
```
mokagio, line 118: Just curious, did this not work?

```swift
XCTAssertTrue(vc is LoginSiteAddressViewController)
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+        
+        WordPressAuthenticator.showLoginForJustWPCom(from: presenterSpy,
+                                                     xmlrpc: "https://example.com/xmlrpc.php",
+                                                     username: "username",
+                                                     connectedEmail: "email-address@example.com")
+        
+        guard let navController = presenterSpy.presentedVC as? LoginNavigationController, let controller = navController.viewControllers.first as? LoginEmailViewController else {
+            XCTFail("Could not fetch correct ViewController")
+            return
+        }
```
mokagio, line 176: XCTest has an API made to simplify writing this kind of checks: [`XCTUnwrap`](https://developer.apple.com/documentation/xctest/3380195-xctunwrap).

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    }
+    
+    func testSignInForWPComReturnsVC() {
+        let vc = WordPressAuthenticator.signinForWPCom()
+        
+        XCTAssertTrue((vc as Any) is LoginEmailViewController)
+    }
+    
+    func testSignInForWPComWithLoginFieldsReturnsVC() {
+        let navController = WordPressAuthenticator.signinForWPCom(dotcomEmailAddress: "example@email.com", dotcomUsername: "username") as! UINavigationController
```
mokagio, line 223: This is another instance in which using `XCTUnwrap` would be handy, as it would prevent the test from crashing in case the return type changes.

It's better to not have tests crash because when they do they stop running, so you don't get information about the rest of tests that didn't get a chance to run. Also, in CI the output for a test run with a crash is not that clear.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+        let vc = navController.topViewController
+        
+        XCTAssertTrue((navController as Any) is UIViewController)
```
mokagio, line 226: I _think_ this assertion is redundant because [`UINavigationController` is a subclass of `UIViewController`](https://developer.apple.com/documentation/uikit/uinavigationcontroller) and the tests already ensures that the instance is a `UINavigationController`.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+        XCTAssertTrue((vc as Any) is LoginWPComViewController)
+    }
+    
+    func testSignInForWPComSetsEmptyLoginFields() {
+        let navController = WordPressAuthenticator.signinForWPCom(dotcomEmailAddress: nil, dotcomUsername: nil) as! UINavigationController
+        let vc = navController.topViewController as! LoginWPComViewController
+        
+        XCTAssertEqual(vc.loginFields.emailAddress, String())
+        XCTAssertEqual(vc.loginFields.username, String())
```
mokagio, line 235:
```suggestion
        XCTAssertEqual(vc.loginFields.emailAddress, "")
        XCTAssertEqual(vc.loginFields.username, "")
```

There's nothing _wrong_ with using `String()`, but I think `""` is clearer. It takes less time to read, too, and the easier we can make our code to read, the easier it will be understand.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+        WordPressAuthenticator.shared.delegate = delegate
+        
+        WordPressAuthenticator.trackOpenedLogin()
+        
+        guard let trackedEvent = delegate.trackedElement else {
+            XCTFail("Event not Tracked")
+            return
+        }
+        
+        XCTAssertEqual(trackedEvent, WPAnalyticsStat.openedLogin)
```
mokagio, line 249: If you were to do the following, the test would still fail in case the event wasn't tracked and you'd have a more compact test.

```suggestion
        XCTAssertEqual(delegate.trackedElement, WPAnalyticsStat.openedLogin)
```

What do you think?

P.S. I didn't try compiling this code. If it doesn't compile, please ignore my comment.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -1,35 +1,42 @@
...
 //
 class WordPressAuthenticatorTests: XCTestCase {
-
+    
+    let timeInterval = TimeInterval(3)
+    
+    override class func setUp() {
+        super.setUp()
+        
+        WordPressAuthenticator.initialize(configuration: MockWordpressAuthenticatorProvider.wordPressAuthenticatorConfiguration(), style: MockWordpressAuthenticatorProvider.wordPressAuthenticatorStyle(.random), unifiedStyle: MockWordpressAuthenticatorProvider.wordPressAuthenticatorUnifiedStyle(.random))
```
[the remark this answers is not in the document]
charliescheer: I appreciate the suggestion.  I think a lot about the human readability of the code that I write.  This sort of suggestion is good for me to consider.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    func testIsGoogleAuthURL() {
+        let url = URL(string: "https://google.com")!
+        
+        XCTAssertTrue(WordPressAuthenticator.shared.isGoogleAuthUrl(url))
+    }
+    
+    func testIsWordPressAuthURL() {
+        let url = URL(string: "https://magic-login")!
+        
+        XCTAssertTrue(WordPressAuthenticator.shared.isWordPressAuthUrl(url))
```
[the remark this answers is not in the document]
charliescheer: Good call.  I added some more asserts to both of the URL tests to make them more detailed

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
charliescheer: Didn't know about that test case.  Super useful.  Updated the guard let functions to use it!

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
charliescheer: I tested this a bit and funny enough it depends on the complexity of the function being tested.  For functions like:
`public class func signinForWPCom() -> LoginEmailViewController`
that function doesn't do much except create the VC and return it (no modifications to the controller at all) if you ask vc is LoginEmailViewController that will always succeed because that is what the function returns.  So that kinda makes that test not very interesting, it is just confirming the function actually works, but to do so first you have to mark it as any or the `is` will always succeed

But the other functions like `signinForWPOrg() -> UIViewController` defines a LoginSiteAddressViewController and then returns it as a UIViewController, and in that case you don't need to first define it as Any to be able to check the that the controller is the correct class

I have removed the calls to Any where it is not needed

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
charliescheer: Good call, removed that assertion

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
charliescheer: Hmmm, I could have sworn that when I first wrote this test I user "" instead of String() and the test failed unless I initialized a string this way...  anyways....  I agree I think it is clearer, not sure why I thought this failed.  Updated

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
charliescheer: Totally, much clearer!

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
         WordPressAuthenticator.deleteLoginInfoForTokenAuth()
         retrievedLoginFields = WordPressAuthenticator.retrieveLoginInfoForTokenAuth()
         retrievedEmail = retrievedLoginFields.username
-
+        
         XCTAssert(email != retrievedEmail, "Saved loginFields should be deleted after calling deleteLoginInfoForTokenAuth.")
     }
+    
+    func testAuthenticatorInitHasCorrectProperties() {
+        //COME BACK TO THIS
```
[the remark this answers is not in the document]
charliescheer: This was a note to myself because I started out trying to create a function that tested the init for the class, but realized pretty quick that the test I was working on was bad.  So I moved on with the intention of returning and missed the note while I was cleaning up the tests before creating the PR.

I removed this test because it's not doing much, and the properties being tested in the init are used throughout the rest of the tests.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -38,21 +45,352 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+        let authenticator = WordPressAuthenticator.shared
```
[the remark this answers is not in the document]
charliescheer: I have gone ahead and changed the access control for the WPAuth class init to be internal so that I could initialize a mock of the authenticator instead of using the shared.  

Thanks for the suggestion!

charliescheer:
I have updated all of the suggested changes and re-requested a review.  These is one thing that I would like to get a second look at and it is the way that the WordPressAuthenticator is being initialized to use in the tests.  

Specifically there are two things that I'm not 100% sure on

First, now that I have made it possible to initialize the WP auth class, for the different tests that are using it I am creating and initializing the a local version of the WP authenticator class.  I did not make a mock or make any changes to the class other than the access control to the init function, and was wondering if that was a good approach, or if it would be better subclassed

Second, I am still initializing the singleton for WordPressAuthenticator in the class setup of the WordPressAuthenticatorClasses.  I left this in because in some functions (specifically the WP authenticator function [openAuthenticationURL(_ url: ](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/428c6784f35cdf944caf191629bda05519af257f/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift#L263)

In this function when it instantiates the NUXLinkAuthViewController, the VC grabs styling properties from the singleton WordPress authenticator.  It could be possible to supply this with the same authenticator that is being used to run the test (to avoid having to initialize the shared authenticator) but that felt like a lot of digging and changes when the styling and the NUX controller are not part of the tests that are being run.  But I am can make that change if we want to avoid initializing the shared authenticator

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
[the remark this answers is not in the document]
charliescheer: Can do.  Thanks for the suggestion

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -1,35 +1,46 @@
...
+        )
+        
+    }
+    
     func testBaseSiteURL() {
         var baseURL = "testsite.wordpress.com"
         var url = WordPressAuthenticator.baseSiteURL(string: "http://\(baseURL)")
         XCTAssert(url == "https://\(baseURL)", "Should force https for a wpcom site having http.")
-
+
```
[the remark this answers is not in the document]
charliescheer: Odd I had that enabled....  Reset it in Xcode, hopefully it will start actually removing the whitespace...

mokagio:
I think we can close this @charliescheer, right?

charliescheer:
Yes.  Sorry hadn't had a chance to close it.  Thanks for all of your help with this one

PR #330, opened 2020-07-21
ULS: Set status bar style per view controller displayed

ScoutHarris:
Ref: #315 

This provides the ability to set the status bar style in each view controller.

Since all views use `LoginViewController`, the default style is set there, which is the current style (`WordPressAuthenticator.shared.style.statusBarStyle`).

Each view controller in the unified flows will need to override `preferredStatusBarStyle` and set it to the new style (`WordPressAuthenticator.shared.unifiedStyle.statusBarStyle`). This has been done in the two new Site Address views.

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14494

### DIFF
--- a/WordPressAuthenticator/Signin/LoginViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginViewController.swift
@@ -32,7 +32,13 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
 
         return delegate
     }
-    
+
+    open override var preferredStatusBarStyle: UIStatusBarStyle {
+        // Set to the old style as the default.
+        // Each VC in the unified flows needs to override this to use the unified style.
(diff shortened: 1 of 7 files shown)


WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -32,7 +32,13 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
 
         return delegate
     }
-    
+
+    open override var preferredStatusBarStyle: UIStatusBarStyle {
+        // Set to the old style as the default.
+        // Each VC in the unified flows needs to override this to use the unified style.
```
mindgraffiti, line 38: Would you add a comment mentioning why we override this in each VC? I know it's because we have independent flows that will be enabled, but I doubt I'll remember that a week from now :)

PR #309, opened 2020-06-23
Restoring method to handle SIWA passwordless 2FA account login.

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/14371

This restores the method that handles logins with an account that:
- Has been used to login via SIWA.
- Does not have a WP password set.
- Has 2FA enabled on the WP account.

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14373

For reference, this was broken with https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/284 and https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/286.

PR #331, opened 2020-07-21
Restoring method to handle SIWA passwordless 2FA account login.

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/14371

This restores the method that handles logins with an account that:
- Has been used to login via SIWA.
- Does not have a WP password set.
- Has 2FA enabled on the WP account.

Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14504

For reference, this is re-implementing the original fix from https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/309.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.20.0"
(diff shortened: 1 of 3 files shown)


ScoutHarris:
Thanks @jkmassel !

PR #332, opened 2020-07-22
UL&S: SiteCredentialsViewController: OnePassword function on username for iOS 12 users

### LINKED ISSUE #322: UL&S: Login via Site Address: Username and Password screen
Build the Username and Password screen UI. Connect it to the existing username and password implementation.

Screen Shot 2020-07-08 at 4 03 01 PM

mindgraffiti:
Closes #322 

This PR adds the OnePassword functionality to the TextFieldTableViewCell for the site URL textfield. While testing on iOS 12 I discovered a bug where the SiteAddressViewController's Continue button would not enable after typing in a URL. I accidentally used a textfield delegate method that was available for iOS 13 only. So instead I created an IBAction in the TextFieldTableCell and attached it directly to the textfield. 

I also fixed a bug where the OnePassword icon was displaying in the wrong color on the old UI.

WPiOS Test PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14511

**Concerns**
@Gio2018 I'd like your opinion on this PR too, if you have a moment. I had to go back to using tuples to pass some of the actions from the textfield in TextFieldTableViewCell because a delegate method was not available in iOS 12. I'm not sure that splitting the textfield actions between the Cell and the ViewController is a good idea. I think it will be confusing to the next person who edits it, but I don't know what to do to make it clearer.

### To test
1. `bundle exec pod install` to ensure pods are up to date. If you get an error, try `bundle exec pod update WordPressKit`
2. Build and run (no errors)
3. Use the test PR https://github.com/wordpress-mobile/WordPress-iOS/pull/14511

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.21.0-beta.3"
+  s.version       = "1.21.0-beta.4"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
--- a/WordPressAuthenticator/Extensions/WPStyleGuide+Login.swift
+++ b/WordPressAuthenticator/Extensions/WPStyleGuide+Login.swift
@@ -36,7 +36,8 @@ extension WPStyleGuide {
     }
 
     /// Adds a 1password button to a WPWalkthroughTextField, if available
-    ///
+    /// - Note: this is for the old UI.
+	///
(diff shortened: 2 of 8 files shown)


Gio2018:
Hey @mindgraffiti !

>  I had to go back to using tuples to pass some of the actions from the textfield in TextFieldTableViewCell because a delegate method was not available in iOS 12

I think it's fine for now. I see you added comments in the code related to this, but maybe we can add a new (non critical) issue to WordPressAuthenticator instead? Otherwise the comments (which can still stay there, as long as they are removed when the change is made) tend to be "lost" in the code

ScoutHarris:
I'll go ahead and merge this and @mindgraffiti  can follow up on @Gio2018 's comment next week.

mindgraffiti:
Thanks for the review Giorgio and Diego! I added #340 to catalog the changes that need made after we drop iOS 12 support. Thank you for merging this Stephenie!

PR #335, opened 2020-07-24
Starts implementing tracking for Google Unified Login.

diegoreymendez:
TBD

PR #357, opened 2020-08-04
Implement AnalyticsTracker.

diegoreymendez:
This PR is the first in a series of PRs to implement the analytics tracking in our unified sign in flows.  It was broken up from a larger PR in order to make the review and testing easier (see: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/335).

## Description:

This PR implements `AnalyticsTracker` - a new class for tracking events during the Sign In flows.

This class provides is a mechanism to track steps, interactions and failures within our flows, while using an internal state machine to update the properties for our tracked events.

This class is not really wired to the rest of the code.  This is intentional in order to help reviewers focus on its functionality and its adherence to our specs.  The provided unit tests are meant to help audit its conformity.

## Testing:

Again, since this isn't wired, we just need to ensure the code looks fine, the documentation complete, and that the logic of the unit tests is what we expect it to be.

Of course, feel free to provide feedback on both the logic and naming / documenting / architectural choices.

## Notes:

Some context on the choices in this PR...

Honestly, the state machine is a bit more complicated than I think it needs to be - I'm really not sure what the advantage of carrying properties across events is.  At first I was aiming for something simpler, but I soon realised the state machine was needed in one form or another.

All in all, I think this PR is following the specs we defined here:

`pbArwn-AP-p2`
`pbArwn-I6-p2`


Podfile
```diff
@@ -11,9 +11,8 @@ def wordpress_authenticator_pods
   ##
   pod 'Gridicons', '~> 1.0'
   pod 'WordPressUI', '~> 1.7.0'
-  pod 'WordPressKit', '~> 4.0-beta.0' # Don't change this until we hit 5.0 in WPKit
-  pod 'WordPressShared', '~> 1.9-beta' # Don't change this until we hit 2.0 in WPShared
-
+  pod 'WordPressKit', '~> 4.14-beta'
+  pod 'WordPressShared', '~> 1.10-beta'
```
mindgraffiti, line 15: This isn't necessary. We update WordPressKit and WordPressShared in the host app's podfile. So for example, Authenticator and WordPress-iOS can have the latest pods by running `bundle exec pod update WordPressKit` (in each repo) and it will update to 4.14-beta.

WordPressAuthenticator.podspec
```diff
@@ -41,8 +41,6 @@ Pod::Spec.new do |s|
   s.dependency 'GoogleSignIn', '~> 5.0.2'
   s.dependency 'WordPressUI', '~> 1.7.0'
 
-  # Use a loose restriction that allows both production and beta versions, up to the next major version.
-  # If you want to update which of these is used, specify it in the host app.
-  s.dependency 'WordPressKit', '~> 4.0-beta.0' # Don't change this until we hit 5.0 in WPKit
-  s.dependency 'WordPressShared', '~> 1.9-beta' # Don't change this until we hit 2.0 in WPShared
+  s.dependency 'WordPressKit', '~> 4.14-beta'
+  s.dependency 'WordPressShared', '~> 1.10-beta'
```
mindgraffiti, line 45: same here. If you're having trouble getting access to WordPressKit 4.14.-beta.N, use the command `bundle exec pod update WordPressKit` while in the repo (no `rake dependencies` support in Authenticator, sadly).

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,341 @@
...
+        case source
+        case step
+    }
+    
+    enum Source: String {
+        case `default`
+        case jetpack
+        case share
+        case deeplink
+        case reauthetication
```
mindgraffiti, line 30:
```suggestion
        case reauthentication
```

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,341 @@
...
+    
+    enum Flow: String {
+        /// The initial flow before we decide whether the user is logging in or signing up
+        case wpCom = "wordpress_com"
+        
+        /// The flow that starts when the user starts the Google login
+        ///
+        case googleLogin = "google_login"
+        
+        case googleSignup = "google_signup"
```
mindgraffiti, line 45: I don't see `google_signup` mentioned in pbArwn-AP-p2. I think this one needs deleted in favor of `signup` ?

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,341 @@
...
+        ///
+        case termsOfService = "terms_of_service_clicked"
+        
+        /// When the user tries to sign up with email from the confirmation screen
+        ///
+        case signUpWithEmail = "signup_with_email"
+        
+        /// When the user tries to sign up with Google from the confirmation screen
+        ///
+        case signUpWithGoogle = "signup_with_google"
```
mindgraffiti, line 145: I think `signUpWithApple` needs added above this?

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,341 @@
...
+    func track(failure: String) {
+        track(event(failure: failure))
+    }
+    
+    // MARK: - Event Construction & Context Updating
+    
+    /// Creates an event for a step.  Updates the state machine.
+    ///
+    /// - Parameters:
+    ///     - step: the step we're tracking..
```
mindgraffiti, line 260:
```suggestion
    ///     - step: the step we're tracking.
```

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,341 @@
...
+    
+    // MARK: - Properties: state machine
+    
+    /// Retrieve the last step, flow and source stored in the state machine.
+    ///
+    private func lastProperties() -> [String: String] {
+        return properties(step: context.lastStep, flow: context.lastFlow, source: context.lastSource)
+    }
+    
+    /// Save the step and flow in the state machine.  The source can only be changed directly using `set(source:)`.
```
mindgraffiti, line 335: >The source can only be changed directly using `set(source:)`.

I do appreciate good guard rails 😀

WordPressAuthenticatorTests/Analytics/SignInTrackerTests.swift
```diff
@@ -0,0 +1,178 @@
...
+        XCTAssertEqual(context.lastStep, .emailOpened)
+    }
+    
+    /// Test that when tracking an event through the SignInTracker, the backing analytics tracker
+    /// receives a matching event.
+    ///
+    /// Ref: pbArwn-AP-p2
+    ///
+    func testEventTracking() {
+        let source = SignInTracker.Source.reauthetication
```
mindgraffiti, line 59:
```suggestion
        let source = SignInTracker.Source.reauthentication
```

Podfile
diegoreymendez: Rolled back!

WordPressAuthenticator.podspec
diegoreymendez: Rolled back!

WordPressAuthenticator/Analytics/SignInTracker.swift
diegoreymendez: This came up in a chat with @renanferrari .  Ref: pbArwn-AP-p2#comment-1160

WordPressAuthenticator/Analytics/SignInTracker.swift
diegoreymendez: I'm still not 100% sure it's the best approach, but that was the intention!  If it needs to change we'll be able to re-evaluate in the actual implementation of the tracking.

WordPressAuthenticator/Analytics/SignInTracker.swift
diegoreymendez: Good catch!  I think this was missing mostly because the list I'm using is shared with Android.

diegoreymendez:
Ready for another look.

WordPressAuthenticator/Analytics/SignInTracker.swift
mindgraffiti: Ugh I never hit send on that comment. Ok, this is going to add a few small steps to this PR.

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,345 @@
...
+        
+        /// The flow that starts when the user starts the Google login
+        ///
+        case googleLogin = "google_login"
+        
+        case googleSignup = "google_signup"
+        
+        /// Flow for Sign in with Apple
+        ///
+        case apple = "siwa_login"
```
mindgraffiti, line 49: We'll also need `siwa_signup`. 

Pinging @ScoutHarris to confirm / deny that ☝️ is true.

WordPressAuthenticator/Analytics/SignInTracker.swift
```diff
@@ -0,0 +1,345 @@
...
+        
+        /// Sign in with the icloud keychain
+        ///
+        case keychain = "icloud_keychain_login"
```
mindgraffiti, line 53: AFAIK, icloud_keychain can't be used during sign up, so this event by itself is correct.

@ScoutHarris to confirm / deny that ☝️  is true

diegoreymendez: Btw, I'm good with changing the name of it if we feel `iCloudLogin` or something similar would be more consistent.

WordPressAuthenticator/Analytics/SignInTracker.swift
diegoreymendez: Just want to double check I got this right.  I've added documentation to this, and the event is coming from this comment: pbArwn-AP-p2#comment-1160

Does that resolve?

WordPressAuthenticator/Analytics/SignInTracker.swift
ScoutHarris: Agreed. We can determine if it's login or signup in `AppleAuthenticator` (as we do currently), so we should add `siwa_signup`. Thanks @mindgraffiti !

WordPressAuthenticator/Analytics/SignInTracker.swift
ScoutHarris: @mindgraffiti is again correct here - keychain is only used for login.

I do like making this one a bit more specific and consistent. To differentiate from the _other_ keychain, maybe we can go with something like `iCloudKeychainLogin`?

WordPressAuthenticator/Analytics/SignInTracker.swift
mindgraffiti: Yep! That is correct.

WordPressAuthenticator/Analytics/SignInTracker.swift
diegoreymendez: Signup for Apple added.

WordPressAuthenticator/Analytics/SignInTracker.swift
diegoreymendez: The flow is now named `iCloudKeychainLogin`.

diegoreymendez:
@ScoutHarris - This is ready for another look.

diegoreymendez:
Thank you!

Issue #352, opened 2020-07-31
ULS: new WP password view


ScoutHarris:
Create new WP password entry view and re-implement functionality.

### V1

Tasks: 
- [x] Initial view. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/353)
- [x] View with content. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/356)
- [x] Implement functionality.
  - [x] Gravatar & email. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/359)
  - [x] Password validation and reset password link. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/363)
  - [x] 1Password support. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/375, https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/377)
  - [x] Password manager support. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/375)
- [x] Show from unified WP path. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/435)
- [x] Show from unified Google path. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/353)
- [x] Show from unified SIWA path. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/376)
- [x] Accessibility. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/378)

Design from Zeplin:

<img width="335" alt="Screen Shot 2020-07-31 at 1 56 18 PM" src="https://user-images.githubusercontent.com/1816888/89072360-a4cf5500-d335-11ea-8774-4c83e9422fe5.png">

PR #363, opened 2020-08-05
ULS Password view: Add password validation and forgot password link action.

ScoutHarris:
Ref: #352 
Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14589

This change: 
- Copies the password validation from the old password VC (`LoginWPComViewController`) to the new one (`PasswordViewController`). 
- Enables the `Reset your password` link.
- Now shows the new 2FA view in the WP flow if it's in the unified Google flow.

| Invalid Password | Reset Password |
|--------|-------|
| invalid_password | reset_password |

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
@@ -16,7 +16,6 @@ class PasswordViewController: LoginViewController {
 
     override var loginFields: LoginFields {
         didSet {
-            // Clear the password (if any) from LoginFields.
             loginFields.password = ""
         }
     }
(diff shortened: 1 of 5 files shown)


WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
```diff
@@ -14,6 +14,13 @@ class PasswordViewController: LoginViewController {
     private var rows = [Row]()
     private var errorMessage: String?
 
+    override var loginFields: LoginFields {
+        didSet {
+            // Clear the password (if any) from LoginFields.
```
diegoreymendez, line 19: The line below is pretty self explanatory and doesn't need documenting.

PR #364, opened 2020-08-07
Fix issues in google auth basic tracking

diegoreymendez:
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14600

This PR addresses two issues with the analytics tracking in the basic Google authentication flow.

1. It tracks the change of the login flow to a signup flow if the user is logging in with Google, and a matching WPCom account does not exist.
2. It tracks the success of the login flow when logging in through Google, when the WPcom account has a 2FA code set.

In order to make this easier, some changes were made to `AnalyticsTracker`:

1. The context is no longer what's meant to be shared.  We now have an `AnalyticsTracker` singleton to make things easier.
2. The flow is no longer set in each tracking call.  It's guarded so that it's possible for us to track a success step without having to know what flow we're in.  We're basically leveraging the advantage of having a state machine for tracking.

Testing is done through the WPiOS PR.

### DIFF
--- a/WordPressAuthenticator/Analytics/AnalyticsTracker.swift
+++ b/WordPressAuthenticator/Analytics/AnalyticsTracker.swift
@@ -215,9 +215,13 @@ public class AnalyticsTracker {
         case createAccount = "create_account"
     }
     
-    /// Provides the sign-in tracking state machine that can be shared across different trackers if needed.
+    /// Shared Instance.
     ///
-    public class Context {
+    public static var shared = AnalyticsTracker()
+    
+    /// State for the analytics tracker.
+    ///
+    private class State {
         var lastFlow: Flow
         var lastSource: Source
         var lastStep: Step
@@ -229,7 +233,7 @@ public class AnalyticsTracker {
         }
     }
     
-    let context: Context
+    private let state = State()
     
     /// The backing analytics tracking method.  Can be overridden for testing purposes.
     ///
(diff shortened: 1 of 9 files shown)


WordPressAuthenticator/Analytics/AnalyticsTracker.swift
```diff
@@ -215,9 +215,13 @@ public class AnalyticsTracker {
...
     
-    /// Provides the sign-in tracking state machine that can be shared across different trackers if needed.
+    /// Shared Instance.
     ///
-    public class Context {
+    public static var shared = AnalyticsTracker()
+    
+    /// State for the analytics tracker.
+    ///
+    private class State {
```
diegoreymendez, line 224: Since this won't be shared or public anymore, it now represents an internal state.

WordPressAuthenticator/Analytics/AnalyticsTracker.swift
```diff
@@ -229,25 +233,24 @@ public class AnalyticsTracker {
...
         self.track = track
     }
     
     // MARK: - Tracking
     
     /// Track a step within a flow.
     ///
-    func track(step: Step, flow: Flow) {
-        track(event(step: step, flow: flow))
+    func track(step: Step) {
```
diegoreymendez, line 252: The flow is now set explicitly when it changes.  No need to specify it for each call.

This ensures that we can log a step even without knowing precisely what flow we're in.  This is particularly useful for loggin in the success in our TwoFactorVC, since it's a success regardless of what flow we're in.

PR #366, opened 2020-08-09
WordPressAuthenticator unit tests

charliescheer:
This PR is a breakout from another [PR](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/328) per the suggestions in this [comment](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/328#pullrequestreview-460608280)  I have taken out much of the unit tests from that PR into this new one so these can get looked over to continue with improving the testing coverage.

In this PR much of the WordPressAuthenticator class and some associated helper classes are covered.  The  tests that were added cover the following sections ow the WordPressAuthenticator pod
WordPressAuthenticator Credentials 
WordPressAuthenticator Sourcetags 
WordPressAuthenticator Notification Dispatching 
WordPressAuthenticator View Presenting 
WordPressAuthenticator URL verification 

Things to check:

- [ ] All tests pass

- [ ] Property injection in function  WordPressAuthenticator.fetchOnePasswordCredentials is done correctly

- [ ] Xcode project and scheme updates are all okay (in the previous PR I have accidentally changed the cocoapod version, just want to make sure these updates don't cause any issues for others)

Any other suggestions and improvements are very welcome.  Thanks!!


WordPressAuthenticator.xcodeproj/xcshareddata/xcschemes/WordPressAuthenticator.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES">
```
mokagio, line 30: Always good to have.

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -71,7 +71,7 @@ import AuthenticationServices
 
     /// Designated Initializer
     ///
-    private init(configuration: WordPressAuthenticatorConfiguration,
+    internal init(configuration: WordPressAuthenticatorConfiguration,
```
mokagio: Nitpick. Since `internal` is the default, I'd omit it.

```suggestion
    init(configuration: WordPressAuthenticatorConfiguration,
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
 
         XCTAssert(email != retrievedEmail, "Saved loginFields should be deleted after calling deleteLoginInfoForTokenAuth.")
     }
+
+//MARK: WordPressAuthenticator Notification Tests
```
mokagio:
```suggestion
  // MARK: WordPressAuthenticator Notification Tests
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    func testDispatchesSupportPushNotificationCleared() {
+        let authenticator = MockWordpressAuthenticatorProvider.getWordpressAuthenticator()
+        let _ = expectation(forNotification: .wordpressSupportNotificationCleared, object: nil, handler: nil)
+
+        authenticator.supportPushNotificationCleared()
+
+        waitForExpectations(timeout: timeInterval, handler: nil)
+    }
+
+//MARK: View Tests
```
mokagio:
```suggestion
  // MARK: View Tests
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -1,10 +1,20 @@
 import XCTest
 @testable import WordPressAuthenticator
 
-
 // MARK: - WordPressAuthenticator Unit Tests
 //
 class WordPressAuthenticatorTests: XCTestCase {
+    let timeInterval = TimeInterval(3)
```
mokagio: What do you think of this variable name to explain to the reader that this value is the default timeout to use in the async expectations?

```suggestion
    let timeout = TimeInterval(3)
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    }
+
+    func testShowLoginFromPresenterReturnsLoginInitialVC() {
+        let presenterSpy = ModalViewControllerPresentingSpy()
+        let expectation = XCTNSPredicateExpectation(predicate: NSPredicate(block: { (_, _) -> Bool in
+            return presenterSpy.presentedVC != nil
+        }), object: .none)
+
+        WordPressAuthenticator.showLoginFromPresenter(presenterSpy, animated: true)
+        wait(for: [expectation], timeout: 3)
```
mokagio: Should this use `timeInterval` (or `timeout` as per my earlier suggestion), too?

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    }
+
+    func testShowLoginForJustWPComPresentsCorrectVC() {
+        let presenterSpy = ModalViewControllerPresentingSpy()
+        let expectation = XCTNSPredicateExpectation(predicate: NSPredicate(block: { (_, _) -> Bool in
+            return presenterSpy.presentedVC != nil
+        }), object: .none)
+
+        WordPressAuthenticator.showLoginForJustWPCom(from: presenterSpy)
+        wait(for: [expectation], timeout: 3)
```
mokagio: Should this use `timeInterval` (or `timeout` as per my earlier suggestion), too?

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+
+        WordPressAuthenticator.showLoginForJustWPCom(from: presenterSpy,
+                                                     xmlrpc: "https://example.com/xmlrpc.php",
+                                                     username: "username",
+                                                     connectedEmail: "email-address@example.com")
+
+        let navController = try XCTUnwrap(presenterSpy.presentedVC as? LoginNavigationController)
+        let controller = try XCTUnwrap(navController.viewControllers.first as? LoginEmailViewController)
+
+        wait(for: [expectation], timeout: 3)
```
mokagio: Should this use `timeInterval` (or `timeout` as per my earlier suggestion), too?

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    }
+
+    func testShowLoginForSelfHostedSitePresentsCorrectVC() {
+        let presenterSpy = ModalViewControllerPresentingSpy()
+        let expectation = XCTNSPredicateExpectation(predicate: NSPredicate(block: { (_, _) -> Bool in
+            return presenterSpy.presentedVC != nil
+        }), object: .none)
+
+        WordPressAuthenticator.showLoginForSelfHostedSite(presenterSpy)
+        wait(for: [expectation], timeout: 3)
```
mokagio: Should this use `timeInterval` (or `timeout` as per my earlier suggestion), too?

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+
+        XCTAssertTrue(presenterSpy.presentedVC is LoginNavigationController)
+    }
+
+    func testSignInForWPComReturnsVC() {
+        let vc = WordPressAuthenticator.signinForWPCom()
+
+        XCTAssertTrue((vc as Any) is LoginEmailViewController)
```
mokagio: Would the same form used in the other tests work here, too?

```suggestion
        XCTAssertTrue(vc is LoginEmailViewController)
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+    func testSignInForWPComWithLoginFieldsReturnsVC() throws {
+        let navController = try XCTUnwrap(WordPressAuthenticator.signinForWPCom(dotcomEmailAddress: "example@email.com", dotcomUsername: "username") as? UINavigationController)
+        let vc = navController.topViewController
+
+        XCTAssertTrue(vc is LoginWPComViewController)
+    }
+
+    func testSignInForWPComSetsEmptyLoginFields() {
+        let navController = WordPressAuthenticator.signinForWPCom(dotcomEmailAddress: nil, dotcomUsername: nil) as! UINavigationController
+        let vc = navController.topViewController as! LoginWPComViewController
```
mokagio: How do you feel about using `try XCTUnwrap` and `as?` here, like you did in a few tests above?

If we can prevent the tests from crashing, we'll get a failure report with more information.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+
+        XCTAssertEqual(vc.loginFields.emailAddress, "")
+        XCTAssertEqual(vc.loginFields.username, "")
+    }
+
+//MARK: WordPressAuthenticator URL verification Tests
```
mokagio:
```suggestion
  // MARK: WordPressAuthenticator URL verification Tests
```

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
```diff
@@ -55,4 +65,190 @@ class WordPressAuthenticatorTests: XCTestCase {
...
+        let loginFields = LoginFields()
+        loginFields.username = "user123"
+        loginFields.password = "knockknock"
+
+        let result = WordPressAuthenticator.openAuthenticationURL(url!, allowWordPressComAuth: false, fromRootViewController: UIViewController())
+
+        XCTAssertFalse(result)
+    }
+
+//MARK: WordPressAuthenticator OnePassword Tests
```
mokagio:
```suggestion
  // MARK: WordPressAuthenticator OnePassword Tests
```

WordPressAuthenticatorTests/Mocks/MockWordpressAuthenticatorProvider.swift
```diff
@@ -0,0 +1,101 @@
+@testable import WordPressAuthenticator
+
+struct MockWordpressAuthenticatorProvider {
```
mokagio: I find the name "mock" here misleading. In most literature, like [here](http://xunitpatterns.com/Mock%20Object.html), mock is a test double.

When I read in the test that "mock provider, get authenticator" was used, the first thing I thought was that the test was checking the behavior of the mock test double instead of the production class.

This type returns actual production code instances, not test doubles.

How do you feel about the name `WordpressAuthenticatorProvider`? Or `WordpressAuthenticatorProviderForTests` to make it clearer that these are preconfigured instances for the tests?

WordPressAuthenticatorTests/Authenticator/WordPressSourceTagTests.swift
```diff
@@ -0,0 +1,175 @@
...
+class WordPressSourceTagTests: XCTestCase {
+    
+    func testGeneralLoginSourceTag() {
+        let tag = WordPressSupportSourceTag.generalLogin
+        
+        let nameExpectation = "generalLogin"
+        let originExpectation = "origin:login-screen"
+        
+        XCTAssertEqual(tag.name, nameExpectation)
+        XCTAssertEqual(tag.origin, originExpectation)
```
mokagio: You could keep the tests more compact by inlining the expected values:

```suggestion
        XCTAssertEqual(tag.name, "generalLogin")
        XCTAssertEqual(tag.origin, "origin:login-screen")
```

WordPressAuthenticatorTests/Authenticator/WordPressSourceTagTests.swift
```diff
@@ -0,0 +1,175 @@
...
+    func testWPComSignupAppleSourceTag() {
+        let tag = WordPressSupportSourceTag.wpComSignupApple
+        
+        let nameExpectation = "wpComSignupApple"
+        let originExpectation = "origin:signup-apple"
+        
+        XCTAssertEqual(tag.name, nameExpectation)
+        XCTAssertEqual(tag.origin, originExpectation)
+    }
+}
```
mokagio, line 124: I'd normally argue against writing tests for constant value, like here were the tests checks that each `static var` on `WordPressSupportSourceTag` calls the init method with the expected values. There is basically no logic in here to test.

But, in this case the test might be useful. Since the tag values are shared with external systems, it's valuable to have an extra check to prevent devs from inadvertently change them.

WordPressAuthenticatorTests/Credentials/CredentialsTests.swift
```diff
@@ -0,0 +1,127 @@
...
+                                          isJetpackLogin: false,
+                                          multifactor: false,
+                                          siteURL: siteURL)
+        
+        let rhs = WordPressComCredentials(authToken: token,
+                                          isJetpackLogin: false,
+                                          multifactor: false,
+                                          siteURL: siteURL)
+        
+        XCTAssertTrue(lhs == rhs)
```
mokagio: It's great that we have a test for the custom `==` implementation.

But since the implementation is custom, what's interesting to test is that it returns `true` in those cases were the instances don't have all the same values, but only those that matter to our implementation.

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/develop/WordPressAuthenticator/Credentials/WordPressComCredentials.swift#L39-L41

That's a point worth generalizing: tests should assert interesting behaviors of the software. Tests are a conversation between the developer and the codebase, the should ask questions the answers to which generate insight in the behavior of the software.

WordPressAuthenticatorTests/Credentials/CredentialsTests.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+    
+    func testWordPressComCredentialsNotEquatable() {
```
mokagio: Like any other naming problem in software development, tests names are hard to assign, but using an appropriate name can go a long way in helping future developers who read the code.

Reading "test `WordPressComCredentials` not `Equatable`" makes me think that the type is not `Equatable`. That per se is something that looking at the type signature would already tell me. Also, the compiler would throw an error if I tried to `==` to instances of a non `Equatable` type.

What this test is doing is proving that two instances with a different `siteURL` value are not equal. How could we rename the test to make it clear?

WordPressAuthenticatorTests/Credentials/CredentialsTests.swift
```diff
@@ -0,0 +1,127 @@
...
+                                                                wporg: wporgCredentials)
+        
+        
+        XCTAssertEqual(authenticatorCredentials.wpcom!.authToken, token)
+        XCTAssertEqual(authenticatorCredentials.wpcom!.isJetpackLogin, false)
+        XCTAssertEqual(authenticatorCredentials.wpcom!.multifactor, false)
+        XCTAssertEqual(authenticatorCredentials.wpcom!.siteURL, siteURL)
+        XCTAssertEqual(authenticatorCredentials.wporg!.username, username)
+        XCTAssertEqual(authenticatorCredentials.wporg!.password, password)
+        XCTAssertEqual(authenticatorCredentials.wporg!.xmlrpc, xmlrpc)
```
mokagio: Does this compile?

```suggestion
        XCTAssertEqual(authenticatorCredentials.wporg?.xmlrpc, xmlrpc)
```

In the same spirit of an earlier comment, preventing crashes in the tests will help in the case in which they'll fail by providing clearer failure information.

WordPressAuthenticatorTests/Mocks/MockOnePasswordFacade.swift
```diff
@@ -0,0 +1,45 @@
...
+    
+    init(username: String, password: String, otp: String?) {
+        self.error = nil
+        self.username = username
+        self.password = password
+        if let otp = otp {
+            self.otp = otp
+        } else {
+            self.otp = nil
+        }
```
mokagio: Because both `otp` values are `String?`, you can just do `self.otp = opt`. If the input `otp` is `nil`, `self.otp` will be `nil`, too.

WordPressAuthenticatorTests/Mocks/MockOnePasswordFacade.swift
```diff
@@ -0,0 +1,45 @@
...
+            return
+        }
+        
+        guard let username = username, let password = password else {
+            failure(.unknown)
+            return
+        }
+        
+        let oneTimePassword = otp
+        success(username, password, oneTimePassword)
```
mokagio: Because all the properties are optional, this function needs to do some checks to makes sure the mock hasn't been configured in an inconsistent way.

Swift's types system offers a few ways to _make inconsistent state unrepresentable_.

My favorite approach for a case like this would be to use `Result`.

```swift
// You can defined nested type in Swift. And if you make them private, they won't be visible 
// outside the parent type, which is handy to create little DTO objects like this one.
private struct Parameters {
  let username: String
  let password: String
  let otp: String?
}

// Using private, the users of this test double don't need to know how it works internally
private let result = Result<Parameters, OnePasswordError>

init(username: String, password: String, otp: String?) {
  result = .success(Parameters(username: username, password: password, otp: otp)
}

init(error: OnePasswordError) {
  result = .failure(error)
}

func findLogin(for url: String, viewController: UIViewController, sender: Any, success: @escaping (String, String, String?) -> Void, failure: @escaping (OnePasswordError) -> Void) {
  switch result {
    case .failure(let error): failure(error)
    case .success(let parameters): success(parameters.username, parameters.password, parameters.otp)
  }
}
```

_I haven't tried this code, it might need a few tweaks to compile._

Notice how with `Result` we remove the change in which some of the parameters are set but other s aren't and have to call `failure(.unknown)`, which would result in a confusing test failure.

WordPressAuthenticatorTests/Mocks/ModalViewControllerPresentingSpy.swift
```diff
@@ -0,0 +1,15 @@
+@testable import WordPressAuthenticator
+
+
+protocol ModalViewControllerPresenting {
+    func present(_ viewControllerToPresent: UIViewController, animated: Bool, completion: (() -> Void)?)
+}
+
+extension UIViewController: ModalViewControllerPresenting {}
```
mokagio: This code should go in the production target, and the methods we test passing the spy should be updated to expect `ModalViewControllerPresenting` instead of `UIViewController`.

Right now, the spy works because it subclasses `UIViewController` not because of the protocol itself. Subclassing `UIViewController` is exactly what we want to avoid by adding the protocol as an extra abstraction layer in between, because when one starts to add `UIViewController`s to the view hierarchy in the tests, things can sometimes go wrong due to window and view management.

mokagio (commented):
Left a few suggestions.

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
charliescheer: That one is a little weird.  Where the other vc is Type tests are testing functions that return a super class of the view controller being checked, with the function signinForWPCom() it returns a LoginEmailViewController or fatal errors.  So to make sure this test does anything I had to convert the VC to any or it would throw up a warning that the is check always succeeds.  

Since there isn't much going on in this test, I think removing it is probably the best option.  As you mentioned with the source tags in a different comment, there isn't really any logic being tested here.  I went ahead and removed it, but what do you think?  Should it be there?

WordPressAuthenticatorTests/Authenticator/WordPressSourceTagTests.swift
charliescheer: Good thing to consider though.  I was generally looking to maximize coverage, and not thinking as much about how this isn't really testing any outcome.  I will keep that in mind going forward

WordPressAuthenticatorTests/Credentials/CredentialsTests.swift
charliescheer: Yes it does :)  Updated to avoid the crashing

WordPressAuthenticatorTests/Mocks/ModalViewControllerPresentingSpy.swift
charliescheer: Makes sense.  Thanks :)

WordPressAuthenticatorTests/Mocks/MockOnePasswordFacade.swift
charliescheer: Thanks for this rework.  That's a much clearer way to approach it!

charliescheer:
Updated the changes.  Let me know if there is anything else that needs updating

WordPressAuthenticatorTests/Authenticator/WordPressAuthenticatorTests.swift
mokagio: Thanks for the extra context. In light of that, then yes **removing the test was the best thing to do** 👍 

If the type system is already ensuring for us that the return value is `LoginEmailViewController`, then there is no added value in writing a test for it, because the compiler will already enforce that.

WordPressAuthenticatorTests/Credentials/CredentialsTests.swift
```diff
@@ -33,32 +33,37 @@ class CredentialsTests: XCTestCase {
...
+        let differentAuthToken = WordPressComCredentials(authToken: "ARSTDBVCXZ(*&^%$",
+                                                         isJetpackLogin: false,
+                                                         multifactor: false,
+                                                         siteURL: siteURL)
+
+        XCTAssertTrue(credential == match)
+        XCTAssertTrue(credential == differentJetpack)
+        XCTAssertTrue(credential == differentMultifactor)
+        XCTAssertFalse(credential == differentSiteURL)
+        XCTAssertFalse(credential == differentAuthToken)
```
mokagio: I'd recommend using `XCTAssertEqual` and `XCTAssertNotEqual`.

The kind of assertions you use makes a difference in how understandable the test failures will be.
Consider these two assertions verifying the same condition:

```swift
let sections: [Section] = getSections()

XCTAssertTrue(sections.count == 1)
XCTAssertEqual(sections.count, 1)
```

If the `sections` array has more than one element, the first will fail with:

```
XCTAssertTrue failed
```


while the second will fail with:

```
XCTAssertEqual failed: ("1") is not equal to ("2")
```

Which one do you think better represent that the array count was not the expected one?

mokagio (approved):
Nice job!

I left a non blocking comment, up to you whether to address it or not. Let's land this in and enjoy more tests around this important code.

WordPressAuthenticatorTests/Credentials/CredentialsTests.swift
charliescheer: I agree I think it makes more sense to use the equal/not equal for these tests.  I updated the tests

PR #367, opened 2020-08-10
Additional tracking for Google Sign In.

diegoreymendez:
Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14592
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14611

Additional tracking for Google Sign In, and some enhancements to make the new analytics tracker easier to use in client Apps.

## Changes:

1. Renames `AnalyticsTracker` to `AuthenticatorAnalyticsTracker`.  The reason this is done is because the `WordPressAuthenticator` module name cannot be used to prefix the module classes.  This is because the module contains a class named `WordPressAuthenticator` (same name as the module) and this causes issues in client Apps when trying to use the module name as a class prefix.
2. Makes some enums in `AuthenticatorAnalyticsTracker.swift` public so that their usable from client Apps.
3. Sets the `.default` analytics tracking source when appropriate.
4. Adds a lot of tracking code.
5. Temporarily guards the unified-tracking so that we only track if `enableUnifiedAuth` is true.

## Testing:

Please see the testing steps in the WPiOS PR.


diegoreymendez:
Hey @mindgraffiti , @ScoutHarris.  I'm temporarily removing this from review to add more successful test cases for the remaining Sign In flows.

WordPressAuthenticator/Services/LoginFacade.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  LoginFacade.swift
+//  WordPressAuthenticator
+//
+//  Created by Diego Rey Mendez on 12/08/2020.
+//  Copyright © 2020 Automattic. All rights reserved.
```
ScoutHarris, line 6: Can you remove this? We don't normally leave them.

WordPressAuthenticator/Signin/LoginEmailViewController.swift
```diff
@@ -213,12 +213,16 @@ open class LoginEmailViewController: LoginViewController, NUXKeyboardResponder {
             }
 
             vc.googleTapped = { [weak self] in
+                guard let self = self else { return }
```
ScoutHarris, line 216: Per our style guide, `return` should be on a separate line. 
```
guard let self = self else { 
return 
}
```

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -138,12 +143,14 @@ class LoginPrologueViewController: LoginViewController {
         }
 
         vc.googleTapped = { [weak self] in
+            guard let self = self else { return }
```
ScoutHarris, line 146: Same here - `return` should be on a separate line.

WordPressAuthenticator/Unified Auth/View Related/2FA/TwoFAViewController.swift
```diff
@@ -387,7 +388,10 @@ private extension TwoFAViewController {
         cell.configureButton(text: WordPressAuthenticator.shared.displayStrings.textCodeButtonTitle)
 
         cell.actionHandler = { [weak self] in
-            self?.requestCode()
+            guard let self = self else { return }
```
ScoutHarris, line 391: Same here - `return` should be on a separate line.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
```diff
@@ -103,6 +103,13 @@ class PasswordViewController: LoginViewController {
     }
 
     override func displayError(message: String, moveVoiceOverFocus: Bool = false) {
+        // The reason why this check is necessary is that we're calling this method
+        // with an empty error message when setting up the VC.  We don't want to track
+        // an empty error when that happens.
+        if message.count > 0 {
```
ScoutHarris, line 109: Just a suggestion - `isEmpty` is more efficient as it doesn't have to count.

WordPressAuthenticator/Services/LoginFacade.swift
diegoreymendez: Yeah, that shouldn't be there.

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
diegoreymendez: Changed.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
diegoreymendez: Changed.

PR #370, opened 2020-08-10
UL&S: Use consistent naming when referring to "Signup"

mindgraffiti:
It should not be "SignUp". No WPiOS testing PR because this change will be incorporated when the `SignupMagicLinkViewController` PR is opened.


mindgraffiti:
Thanks for the review Stephenie!

PR #371, opened 2020-08-10
Constrain nav bar, and bar button appearance changes to auth classes.

aerych:
This PR makes a small change to how the appearance of UINavigationBar and UIBarButtonItem instances is handled. 

Prior to this PR the global appearance proxy for UINavigationBar and UIBarButtonItem was being modified.  This had a consequence of the changes carrying over into other parts of the implementing app that might use different styles. 
This PR constrains the changes to UINavigationBar and UIBarButtonItem instances that appear within a LoginNavigationController instance.

To test:
- Confirm the issue by launching an app that implements the Authenticator.
- Set the authenticator's navbar background style via the WordPressAuthenticatorStyle config to something garish (mauve maybe? chartreuse? assert yourself ;) )
- Run the app and go through the auth flow to successfully log in.
- Note that the app's navbar remains the color you set. The same is true for the bar button items depending on how you've configured styles.
- Re-launch the app and confirm that the app's normal navbar style appears.
- Apply the patch and repeat the test. 
- Confirm that once the auth flow is complete the app's normal navbar style appears.

(Newspack-iOS is a nifty test platform for this one ;) )

@ScoutHarris could I trouble you with this one?

### DIFF
--- a/WordPressAuthenticator/Signin/LoginViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginViewController.swift
@@ -122,11 +122,11 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
             appearance.shadowColor = hideBottomBorder ? .clear : .separator
             appearance.backgroundColor = backgroundColor
             appearance.titleTextAttributes = [.foregroundColor: titleTextColor]
-            UIBarButtonItem.appearance().tintColor = buttonTextColor
+            UIBarButtonItem.appearance(whenContainedInInstancesOf: [LoginNavigationController.self]).tintColor = buttonTextColor
             
-            UINavigationBar.appearance().standardAppearance = appearance
-            UINavigationBar.appearance().compactAppearance = appearance
-            UINavigationBar.appearance().scrollEdgeAppearance = appearance
+            UINavigationBar.appearance(whenContainedInInstancesOf: [LoginNavigationController.self]).standardAppearance = appearance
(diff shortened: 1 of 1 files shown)


ScoutHarris:
Oh, also, just a head's up. I had to change NP's WPKit version to get it to build due to this error.

```
[!] CocoaPods could not find compatible versions for pod "WordPressKit":
  In snapshot (Podfile.lock):
    WordPressKit (= 4.13.0, ~> 4.0-beta.0, ~> 4.13.0)

  In Podfile:
    WordPressAuthenticator (from `../WordPressAuthenticator-iOS`) was resolved to 1.23.0-beta.2, which depends on
      WordPressKit (~> 4.14-beta)

    WordPressKit (~> 4.13.0)
```

aerych:
> Oh, also, just a head's up. I had to update NP's WPKit version to get it to build

Yep yep! I'll sort it eventually.  

I don't think there's a rush to publish this change. I'd be fine waiting til the next update y'all have planned rather than update the podspec now.  What do you think?

ScoutHarris:
> I don't think there's a rush to publish this change. I'd be fine waiting til the next update y'all have planned rather than update the podspec now. What do you think?

If you don't need it soon-ish, that's fine. We don't "plan" updates, but we sure do a lot of them. 😄 We'll pick it up eventually.

aerych:
Perfect. Thanks @ScoutHarris !

PR #372, opened 2020-08-12
Social account logins: Don't update username when using a password manager.

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/14620
Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14626

In the current login flow (not unified), using a password manager on the Password view:
- If logging in via WordPress, selecting any account will still update the username and password. (This is existing behavior.)
- If logging in via Google or SIWA (with an account that has a WP password), only the password will be used. The username is ignored. (This is new behavior.)

### DIFF
--- a/WordPressAuthenticator/Signin/LoginWPComViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginWPComViewController.swift
@@ -187,8 +187,7 @@ class LoginWPComViewController: LoginViewController, NUXKeyboardResponder {
             // Ref: https://git.io/JJSUM
             if loginFields.meta.socialService != nil {
                 emailLabel?.text = loginFields.username
-            }
-            else {
+            } else {
                 loginFields.username = sender.nonNilTrimmedText()
             }
         default:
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator/Signin/LoginWPComViewController.swift
```diff
@@ -181,7 +181,16 @@ class LoginWPComViewController: LoginViewController, NUXKeyboardResponder {
...
         case emailLabel:
-            loginFields.username = sender.nonNilTrimmedText()
+            // The email can only be changed via a password manager.
+            // In this case, don't update username for social accounts.
+            // This prevents inadvertent account linking.
+            // Ref: https://git.io/JJSUM
+            if loginFields.meta.socialService != nil {
+                emailLabel?.text = loginFields.username
+            }
+            else {
```
diegoreymendez, line 191: Is the `else {` on a newline intentional?

ScoutHarris: Woops. Forgot to fix that. Thanks!

Issue #345, opened 2020-07-30
UL&S: Sign up


mindgraffiti:
### UI
- [x] Gravatar + email UITableViewCell #350 
- [x] Navigate to new UnifiedSignUpViewController #346 
- [x] ~Navigation bar, nav bar items, status bar uses correct styling after navigating from old screen~ this is blocked until UnifiedSignUpEmailViewController is built
- [x] Instruction cell, gravatar + email cell displays #348
- [x] Send Link by Email button #348

### Implementation
- [x] Send Link by Email button implementation #355
- [x] iPad support #355
- [x] Accessibility - dynamic text, VoiceOver #358
- [x] Tracks Events

PR #374, opened 2020-08-12
UL&S: SignupMagicLinkViewController: display the UI

mindgraffiti:
Ref. #345 
Testing PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14633

This PR navigates the user to the "magic link sent!" view in the Unified Signup flow. Ensure the UI matches the design. The "Open Mail" button is not yet implemented.

**Design**
Screen Shot 2020-08-12 at 11 30 56 AM

### To test
1. Check out this branch
2. [Set the email](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/develop/WordPressAuthenticator/Unified%20Auth/View%20Related/Sign%20up/UnifiedSignUpViewController.swift#L26) to a valid, test email address (you can use the plus modifier on your email address, but the gravatar image won't appear if you do that).
3. `bundle exec pod install`
4. Build and run
5. Check out the WPiOS testing PR
6. In the WPiOS testing PR, enable the `.unifiedSignup` feature flag.
7. In the WPiOS testing PR, set the podfile to local pod development.
8. `rake dependencies`
9. Build and run
10. Sign up for WordPress.com > Sign up with Email > Send Link by Email. 

**Expected:** the button navigates you to the "magic link sent" view, which looks like this:

<a href="https://user-images.githubusercontent.com/1062444/90188013-60ee3e00-dd80-11ea-83f2-4dc791578282.png"></a>


WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
```diff
@@ -0,0 +1,197 @@
...
+        case persona
+        case instructions
+        case checkSpam
+        case errorMessage
+
+        var reuseIdentifier: String {
+            switch self {
+            case .persona:
+                return GravatarEmailTableViewCell.reuseIdentifier
+            case .instructions, .checkSpam, .errorMessage:
```
mindgraffiti, line 192: Is this formatting used in WPiOS? If not I can change it back to individual cases:
```
case .instructions:
case .checkSpam:
case .errorMessage:
```

ScoutHarris: I think one `case` is fine. And probably preferred (at least, by me 😄 ).

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
```diff
@@ -0,0 +1,197 @@
...
+
+    @IBOutlet private weak var tableView: UITableView!
+    private var rows = [Row]()
+    private var errorMessage: String?
+    private var shouldChangeVoiceOverFocus: Bool = false
+
+    var emailMagicLinkSource: EmailMagicLinkSource?
+    override var sourceTag: WordPressSupportSourceTag {
+        get {
+            if let emailMagicLinkSource = emailMagicLinkSource,
```
Gio2018, line 19: a minor thing, but you could potentially omit the unwrapping since `optionals` are enums already

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
```diff
@@ -0,0 +1,197 @@
...
+                emailMagicLinkSource == .signup {
+                return .wpComSignupMagicLink
+            }
+            return .loginMagicLink
```
ScoutHarris, line 23: There's going to be a separate Login Magic Link VC, right? So I think this sourceTag will always be signup. At least, it should never return login.

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
```diff
@@ -0,0 +1,197 @@
...
+final class SignupMagicLinkViewController: LoginViewController {
+
+    // MARK: Properties
+
+    @IBOutlet private weak var tableView: UITableView!
+    private var rows = [Row]()
+    private var errorMessage: String?
+    private var shouldChangeVoiceOverFocus: Bool = false
+
+    var emailMagicLinkSource: EmailMagicLinkSource?
```
ScoutHarris, line 16: I don't see that `emailMagicLinkSource` is really used for anything. Will it be needed in the future, or is just carry-over fluff from the old VC?

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
```diff
@@ -0,0 +1,197 @@
...
+        localizePrimaryButton()
+        registerTableViewCells()
+        loadRows()
+    }
+
+    /// Some last-minute gatekeeping.
+    ///
+    func validationCheck() {
+        let email = loginFields.username
+        if !email.isValidEmail() {
```
Gio2018, line 54: It's not clear to me why the condition is evaluated twice. If `email.isValidEmail()` returns false, the `assert` should suffice

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
```diff
@@ -0,0 +1,197 @@
...
+            assert(email.isValidEmail(), "The value of loginFields.username was not a valid email address.")
+        }
+
+        emailMagicLinkSource = loginFields.meta.emailMagicLinkSource
+        assert(emailMagicLinkSource != nil, "Must have an email link source.")
```
Gio2018, line 59: would it be possible to write a unit test for this, instead of using `assert` in the production code?

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
mindgraffiti: Accidental carry-over fluff. It was used to differentiate the "Magic Link sent" view as either `.signup` or `.login` for tracking purposes and whether or not to show the 1Password button. Will delete!

WordPressAuthenticator/Unified Auth/View Related/Sign up/SignupMagicLinkViewController.swift
mindgraffiti: I missed it. The emailMagicLinkSource is cruft from the old code, so I will delete these lines.

mindgraffiti:
Thanks for the reviews Stephenie & Gio!

@Gio2018 I'm thinking about what you said in regards to using `assert`. I will replace the `assert` in my next PR, which will implement error handling. That way I can display an email validation error without leaving an `assert` in production code.

PR #376, opened 2020-08-13
ULS: Show new Password and 2FA views in SIWA path

ScoutHarris:
Ref: #352 
Can be tested with WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14647

For SIWA accounts that have a WP password set, this updates the flow to show the new Password and 2FA views.

| Password | 2FA |
|--------|-------|
| password | 2fa |

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.23.0-beta.5"
+  s.version       = "1.23.0-beta.6"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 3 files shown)


ScoutHarris:
Hrm. It works for my account. The gist is the SIWA account needs a WP password and WP 2FA so you see both the new views. And you should only get prompted for a password if the SIWA account is not connected to the WP account (using the same email).  I'll poke around and maybe see if there's a condition I missed, but atm I don't know why it's not working for you.

ScoutHarris:
Hey @mindgraffiti . I tried 2 accounts and it works. However, one of them fails Apple authentication on a simulator, but works just fine on a real device. Would you mind trying on a device and see what happens?

mindgraffiti:
@ScoutHarris I always test SIWA on a device (didn't know we could use sims!). I think I found the issue. My primary Apple account has been fiddled with too much. I'm switching to a "pristine" Apple account now and am testing again.

ScoutHarris:
Thanks for fighting the good fight @mindgraffiti ! 😆

PR #415, opened 2020-08-26
UL&S iCloud Keychain: Simple successful login path

diegoreymendez:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/14734
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14770

This PR implements the basic login flow for the iCloud Keychain authentication flow.

Please refer to the WPiOS PR for details on how to test this.

### DIFF
--- a/WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
+++ b/WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
@@ -8,10 +8,23 @@ import AuthenticationServices
 ///
 @available(iOS 13, *)
 class StoredCredentialsAuthenticator: NSObject {
+
+    // MARK: - Delegates
     
+    private var authenticationDelegate: WordPressAuthenticatorDelegate {
+        guard let delegate = WordPressAuthenticator.shared.delegate else {
+            fatalError()
+        }
+        return delegate
+    }
+    
+    // MARK: - Configuration
+
     private var authConfig: WordPressAuthenticatorConfiguration {
         WordPressAuthenticator.shared.configuration
     }
+
+    // MARK: - Login Support
     
     private lazy var loginFacade: LoginFacade = {
         let facade = LoginFacade(dotcomClientID: authConfig.wpcomClientId,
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
```diff
@@ -8,10 +8,23 @@ import AuthenticationServices
 ///
 @available(iOS 13, *)
 class StoredCredentialsAuthenticator: NSObject {
+
+    // MARK: - Delegates
+    
+    private var authenticationDelegate: WordPressAuthenticatorDelegate {
+        guard let delegate = WordPressAuthenticator.shared.delegate else {
+            fatalError()
```
mindgraffiti, line 16: Is there a more graceful way to fail when the delegate doesn't exist?

WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
```diff
@@ -61,11 +86,9 @@ class StoredCredentialsAuthenticator: NSObject {
             // No-op for now, but we can decide to implement AppleID login through this authenticator
             // by implementing the logic here.
             break
-        case _ as ASPasswordCredential:
-            // TODO: No-op for now.  The code below will be enabled in my next PR.
-            //
-            //let loginFields = LoginFields.makeForWPCom(username: credential.user, password: credential.password)
-            //loginFacade.signIn(with: loginFields)
+        case let credential as ASPasswordCredential:
```
mindgraffiti, line 89: nitpick: In this case it should be plural, `credentials`, since it stores the username and the password.

diegoreymendez: The credential is the username and password combination - we're not dealing with multiple credentials.

Apple names the types `ASAuthorizationAppleIDCredential` and `ASPasswordCredential` in singular form - this is why I followed their lead (and I agree).

diegoreymendez:
> I had 2fa enabled on 1 of my iCloud Keychain accounts, and it failed to navigate to the 2fa screen. From the code it looks like that will be next?

Yep, that's in the test steps pre-requisites:

> Make sure you have valid credentials in your iCloud Keychain for an account that does not have 2FA enabled.

mindgraffiti:
Aaaand I did not carefully read the instructions 😄 . Thanks for answering my question regardless!

WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
diegoreymendez: This is a bit ugly I agree, but I don't feel the solution is making this optional as this delegate isn't supposed to be absent.

In addition to this, I opted (fow now) to use the same snippet we're using in the [Apple Authenticator](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/8318b4183b7cbc016252b94ab291cf7e3bdef5f3/WordPressAuthenticator/Signin/AppleAuthenticator.swift#L29-L34) and the [Google Authenticator](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/8318b4183b7cbc016252b94ab291cf7e3bdef5f3/WordPressAuthenticator/Unified%20Auth/GoogleAuthenticator.swift#L107-L112).

With all that said, I'm planning to make this better in an upcoming PR in order to avoid scope creep in this PR.

mindgraffiti: Oh that's fair. If that's how we usually handle missing delegates, then ignore that request.

Issue #404, opened 2020-08-24
ULS: Get Started view


ScoutHarris:
Create 'Get Started' view to provide login/signup options.

Tasks:
- [x] Initial view. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/412)
- [x] Add TOS table cell and functionality. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/414)
- [x] Add email address field. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/416)
- [x] Add 'Continue' button. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/428)
- [x] Add email address field / 'Continue' button functionality.
  - [x] Validation. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/435)
  - [x] Show unified Login flow if logging in. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/435)
  - [x] Show unified Signup flow if signing up. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/435)
  - [x] Password manager support. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/438)
- [x] Add "OR" option divider. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/431)
- [x] Show social buttons and TOS in bottom button view. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/433)
- [x] Accessibility. (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/438)
- [x] Tracks.

Design from Zeplin:

<img width="692" alt="orig_get_started" src="https://user-images.githubusercontent.com/1816888/91103216-3629a280-e628-11ea-859b-5a4218c4e754.png">

![Device Examples — Get Started](https://user-images.githubusercontent.com/1816888/91773327-cd03db00-eba3-11ea-8a46-b0d6a4f7a6d6.png)

PR #433, opened 2020-09-01
ULS Get Started: add Social login options

ScoutHarris:
Ref: #404 
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14817

This adds Apple, Google, and TOS buttons to the bottom button view. All are functional, so:
- Apple shows the unified Apple flow.
- Google shows the unified Google flow.
- TOS shows the TOS web page.

| iphone_light | iphone_dark |
|--------|-------|

| ipad_light | ipad_dark |
|--------|-------|

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
@@ -1,5 +1,6 @@
 import UIKit
 import SafariServices
+import WordPressKit
 
 class GetStartedViewController: LoginViewController {
 
@@ -14,6 +15,8 @@ class GetStartedViewController: LoginViewController {
     @IBOutlet private weak var trailingDividerLineWidth: NSLayoutConstraint!
 
     private var rows = [Row]()
+    private var buttonViewController: NUXButtonViewController?
+    private let configuration = WordPressAuthenticator.shared.configuration
 
     // Submit button displayed in the table footer.
     private let continueButton: NUXButton = {
@@ -44,6 +47,7 @@ class GetStartedViewController: LoginViewController {
         loadRows()
         setupContinueButton()
         configureDivider()
+        configureSocialButtons()
     }
 
     // MARK: - Overrides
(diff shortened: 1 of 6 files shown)


mindgraffiti (approved):
Sure, sure, make the 3 button view moved to a full-page view look easy 😉

:shipit:

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -204,19 +224,100 @@ private extension GetStartedViewController {
...
+        tracker.set(flow: .loginWithApple)
+        tracker.track(click: .loginWithApple)
+        
+        AppleAuthenticator.sharedInstance.delegate = self
+        AppleAuthenticator.sharedInstance.showFrom(viewController: self)
+    }
+    
+    @objc func googleTapped() {
+        tracker.set(flow: .loginWithGoogle)
+        tracker.track(click: .loginWithGoogle)
```
Gio2018, line 261: should we track these after the `guard let..` or do we want to track this tap regardless if `GoogleAuthViewController.instantiate` fails or not?

Gio2018 (approved):
Hey @ScoutHarris , just had one small comment, other than that :shipit:!

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
ScoutHarris: Hey @Gio2018 . That's a good question. I've seen most of the tracks happen before guards, so I assume tracking is meant to happen no matter what. I need to review tracking on this page anyway (this bit was just copied from another VC) so I'll bring that up with the team and change it if necessary. Thanks!!

ScoutHarris: Hey @Gio2018 . Well that was an easy discussion. We agreed tap events should be tracked regardless, so they should be before guards, etc. And I just noticed I put the `termsOfService` track in the wrong place. Thanks for bringing this up!

PR #435, opened 2020-09-02
ULS Get Started: add email address handling

ScoutHarris:
Ref: #404 
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/278
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14825

This adds email validation and handling to the Get Started view.
- The email address must be valid before the `Continue` button is enabled.
- Existing accounts:
  - If the account is passwordless, a login link email is requested and the `Check your email...` view is displayed.
  - If the account has a password, the Password view is displayed.
- For an email that does not match a WP account, the Sign Up view is displayed.
- If iCloud keychain authentication fails from the prologue:
  - The email address is pre-populated.
  - The error passed in from `StoredCredentialsAuthenticator` is displayed inline.


WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -14,6 +14,9 @@ class GetStartedViewController: LoginViewController {
     @IBOutlet private weak var trailingDividerLine: UIView!
     @IBOutlet private weak var trailingDividerLineWidth: NSLayoutConstraint!
 
+    // This is public so it can be set from StoredCredentialsAuthenticator.
+    var errorMessage: String?
```
diegoreymendez, line 18: Thanks!

WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -211,8 +211,9 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
         let unifiedGoogle = WordPressAuthenticator.shared.configuration.enableUnifiedGoogle && loginFields.meta.socialService == .google
         let unifiedApple = WordPressAuthenticator.shared.configuration.enableUnifiedApple && loginFields.meta.socialService == .apple
         let unifiedSiteAddress = WordPressAuthenticator.shared.configuration.enableUnifiedSiteAddress && !loginFields.siteAddress.isEmpty
+        let unifiedWordPress = WordPressAuthenticator.shared.configuration.enableUnifiedSiteAddress && loginFields.meta.userIsDotCom
```
diegoreymendez: This should probably be using `enableUnifiedWordPress` (instead of `enableUnifiedSiteAddress`).

WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
```diff
@@ -196,8 +196,8 @@ extension StoredCredentialsAuthenticator {
         if let loginFields = loginFields {
             toVC.loginFields = loginFields
         }
-        toVC.errorToPresent = error
 
+        toVC.errorMessage = error.localizedDescription
```
diegoreymendez, line 200: Thanks.

WordPressAuthenticator/Unified Auth/View Related/2FA/TwoFAViewController.swift
```diff
@@ -384,8 +384,8 @@ private extension TwoFAViewController {
     /// Configure the textfield cell.
     ///
     func configureTextField(_ cell: TextFieldTableViewCell) {
-        cell.configureTextFieldStyle(with: .numericCode,
-                                     and: WordPressAuthenticator.shared.displayStrings.twoFactorCodePlaceholder)
+        cell.configure(withStyle: .numericCode,
```
diegoreymendez, line 387: The previous `with` parameter name was actually closer to the standard promoted by Swift.

From https://swift.org/documentation/api-design-guidelines/#naming > Promote Clear Usage > Omit needless words. > More Detail:

> More words may be needed to clarify intent or disambiguate meaning, but those that are redundant with information the reader already possesses should be omitted. In particular, omit words that merely repeat type information.

In this case we already know we're dealing with a style.

diegoreymendez (commented):
Added some comments.

WordPressAuthenticator/Signin/LoginViewController.swift
ScoutHarris: Gah! Thank you!

WordPressAuthenticator/Unified Auth/View Related/2FA/TwoFAViewController.swift
ScoutHarris: > In this case we already know we're dealing with a style.

Well, not necessarily. Previously, it wasn't immediately clear what the params `with` and `and` were. It was even requested during a previous PR review to clarify the param names. Regarding the style, I've simply moved the word from the method name to the param name to make it more clear. And the method signature is now less wordy since I removed the `TextField` bit from the method name (to me, _that_ was obvious).

diegoreymendez: This is minor, no need to address this if you disagree.

I still think the type is implicitly evident, but it's not something worth holding a PR back on.

diegoreymendez (approved):

ScoutHarris:
Thanks @diegoreymendez !

PR #454, opened 2020-09-10
UL&S: Implements tracking for the WPcom flow and Magic Links.

diegoreymendez:
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14878

Implements tracking for the new unified WPcom login flow, and for magic links.

Changes in this PR:

- `AuthenticatorAnalyticsTracker`'s state is now publicly readable in case the client app needs to remember the tracking state.
- `AuthenticatorAnalyticsTracker`'s source, flow and step can be set from outside WPAuth.
- `AuthenticatorAnalyticsTracker` has a new `resetState()` method, so that the default flow and step can be re-established when necessary.
- Removed the prologue flow start tracking, since we don't currently have the new prologue.
- Tracks several events that were still not being tracked in the WPcom and Magic Link flows.
- Fixes an issue that was causing the flow not to be restablished when the support VC was dismissed.

For details please refer to the WPiOS PR.

### DIFF
--- a/WordPressAuthenticator/Signin/AppleAuthenticator.swift
+++ b/WordPressAuthenticator/Signin/AppleAuthenticator.swift
@@ -251,7 +251,7 @@ extension AppleAuthenticator: ASAuthorizationControllerDelegate {
         if let authorizationError = error as? ASAuthorizationError,
             authorizationError.code == .canceled {
             
-            // If the user cancelled the dialogue, we should assume he somehow tapped to dismiss.
+            // If the user cancelled the dialogue, we should assume they somehow tapped to dismiss.
             tracker.track(click: .dismiss)
             return
         }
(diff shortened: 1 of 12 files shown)


WordPressAuthenticator/Signin/AppleAuthenticator.swift
```diff
@@ -246,10 +246,13 @@ extension AppleAuthenticator: ASAuthorizationControllerDelegate {
...
 
     func authorizationController(controller: ASAuthorizationController, didCompleteWithError error: Error) {
-
+        
         // Don't show error if user cancelled authentication.
         if let authorizationError = error as? ASAuthorizationError,
-        authorizationError.code == .canceled {
+            authorizationError.code == .canceled {
+            
+            // If the user cancelled the dialogue, we should assume he somehow tapped to dismiss.
```
mindgraffiti:
```suggestion
            // If the user cancelled the dialogue, we should assume they somehow tapped to dismiss.
```

mindgraffiti (approved):
In the `LoginMagicLinkViewController` `viewDidLoad()` method, it's missing the Tracks event for "magic link sent screen was shown". 
```
{"source":"default","flow":"login_magic_link","step":"magic_link_requested"}
```

There are a few other missing events in this flow that need added, but nothing that will block this PR.

PR #463, opened 2020-09-16
Login: show correct message when login limit reached

### LINKED ISSUE #237: Wrong error displayed for login attempts that exceed the login limit
If I attempt to log in to a WordPress.com account too many times in quick succession, my login attempts are blocked. However, in the UI I don't see the correct error message — instead I see an incorrect password error. I observed the same issue in both the WordPress and WooCommerce apps.

Steps to reproduce:

1. Create a WordPress.com account and set up 2FA. (I noticed this on an account with 2FA; I couldn't reproduce it as readily on an account without 2FA.)
2. Log in to the account with your password.
3. Log out.
4. Repeat steps 2-3 several times.
5. Eventually, you will exceed the login limit and will be stopped from logging in. Notice that the UI shows an error message about an incorrect password.
6. Go to Help > Activity Logs > Current and notice that the actual error is about exceeding the login limit.

Error in the app logs:
> You have exceeded the login limit.  Please wait a few minutes and try again.

Error in the UI:
> It seems like you've entered an incorrect password. Want to give it another try?

Screenshot from WordPress iOS internal v. 14.6.0.20200406:

ScoutHarris:
Fixes: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/237
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/13977
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14914

This adds a check to the login response error for one indicating a password issue. If found, display the custom incorrect password message. Otherwise, show the response error. The effect is, when the max number of failed login attempts is reached, the correct error message is displayed.

| incorrect_password | exceeded_limit |
|--------|-------|

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.25.0-beta.3"
+  s.version       = "1.25.0-beta.4"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


mindgraffiti (approved):
✅  code
✅  ui tested

:shipit:

PR #470, opened 2020-09-17
UL&S: Adds tracking to the prologue, and fixes some other tracking issues.

diegoreymendez:
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/14931

This PR:

- Enabled tracking for the prologue flow.
- Fixes some issues with tracking clicks in the prologue screen.
- Removes some state stack I had created and opted to roll it back to a simpler solution
- Adds `isBeingPresentedInAnyWay` to view controllers for convenience.
- Removes the dismissal of the SIWA sheet, since we're not tracking that for iCloud Keychain either (matching Android's behavior).
- Some other small tracking fixes.

### DIFF
--- a/WordPressAuthenticator/Analytics/AuthenticatorAnalyticsTracker.swift
+++ b/WordPressAuthenticator/Analytics/AuthenticatorAnalyticsTracker.swift
@@ -288,10 +288,6 @@ public class AuthenticatorAnalyticsTracker {
     ///
     public let state = State()
     
-    /// The stored state
-    ///
-    private var pushedState = [State]()
-    
     /// The backing analytics tracking method.  Can be overridden for testing purposes.
     ///
     let track: TrackerMethod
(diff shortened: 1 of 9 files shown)


mindgraffiti (commented):
1. 
🔵 Tracked: unified_login_step <flow: prologue, source: default, step: start>
instead of `step: start` this one should be `step: prologue`

2. 
🔵 Tracked: unified_login_interaction <click: continue_with_wordpress_com, flow: prologue, source: default, step: start>
The `flow:` is correct here, but the `step` has also been altered, which isn't right. It should be `step: prologue` (because the click grabs the previously reported step).

3. ✅ 
4. ✅ 
5. ✅ 
6. (Test 2) 
🔵 Tracked: unified_login_interaction <click: login_with_site_address, flow: prologue, source: default, step: start>
needs `step: prologue` instead of `step: start`

diegoreymendez:
@thuy - Fixed and ready for another look.

ScoutHarris (approved):

diegoreymendez:
Thanks!

PR #477, opened 2020-09-22
ULS: Remove enableUnifiedGoogle flag.

ScoutHarris:
This removes the `enableUnifiedGoogle` and replaces any checks with `enableUnifiedAuth`. 

Can be tested with WPiOS PR https://github.com/wordpress-mobile/WordPress-iOS/pull/14969.

### DIFF
--- a/WordPressAuthenticator/Analytics/AuthenticatorAnalyticsTracker.swift
+++ b/WordPressAuthenticator/Analytics/AuthenticatorAnalyticsTracker.swift
@@ -255,7 +255,7 @@ public class AuthenticatorAnalyticsTracker {
         
         return Configuration(
             appleEnabled: WordPressAuthenticator.shared.configuration.enableUnifiedApple,
-            googleEnabled: WordPressAuthenticator.shared.configuration.enableUnifiedGoogle,
+            googleEnabled: true,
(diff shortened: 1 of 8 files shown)


WordPressAuthenticator/Analytics/AuthenticatorAnalyticsTracker.swift
```diff
@@ -255,7 +255,7 @@ public class AuthenticatorAnalyticsTracker {
         
         return Configuration(
             appleEnabled: WordPressAuthenticator.shared.configuration.enableUnifiedApple,
-            googleEnabled: WordPressAuthenticator.shared.configuration.enableUnifiedGoogle,
+            googleEnabled: true,
```
ScoutHarris, line 258: @diegoreymendez I just realized these should be `WordPressAuthenticator.shared.configuration.enableUnifiedAuth`. I'll correct them in my next PR.

diegoreymendez: Well... it's not so bad as long as unified auth is on, but I think you're right.

diegoreymendez (approved):

PR #488, opened 2020-09-29
Add an optional completion block to login UI presentation completion

jaclync:
## Background

This PR enables the caller of `WordPressAuthenticator.showLogin` to pass a completion block on presentation completion. For a refactoring task in WCiOS https://github.com/woocommerce/woocommerce-ios/issues/2884, we're aiming to reset the tab view controllers after the app is logged out and we can only do so **after** the login UI presentation finishes. Otherwise, the last logged in UI disappears for a split second before the login UI is presented.

## Changes

- In `WordPressAuthenticator.showLogin`, added a completion block to be called when the login UI presentation completes.

## Testing

No user-facing changes are expected from this change, since the default value for `onCompletion` is the same as the value before (`nil`).

For confidence check:

- In `Podfile` of WPiOS/WPiOS, point `WordPressAuthenticator` to branch `wcios-2884/authenticator-presentation-completion` and run `bundle exec pod install`
- Log out from the app
- Log in to a site --> the login flow should be the same as before

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.26.0-beta.12"
+  s.version       = "1.26.0-beta.13"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


mindgraffiti (approved):
✅ Code - nice documentation on the method
✅ UI tested - logged in / logged out on WPiOS from develop and didn't see any issues (didn't expect any, either)

:shipit:

jaclync:
thanks for the review & testing, Thuy! I tested on WCiOS, merging now

PR #493, opened 2020-10-06
Remove localization from accessibility IDs

rachelmcr:
This PR changes the accessibility identifiers on text fields (e.g. site URL, username, password) to use non-localized strings.  (User-facing strings such as accessibility labels are still localized as expected.)

This ensures the accessibility IDs are the same regardless of the device language, so tests can run in any language.

To test:

Ensure UI tests pass (regardless of device language) on this WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15055

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.26.0"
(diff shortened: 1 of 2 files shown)


startuptester (approved):
- [X] All UI tests pass (regardless of device language) on this WPiOS PR: wordpress-mobile/WordPress-iOS#15055

👍

PR #506, opened 2020-10-26
Removes a green background color from a stack view.

aerych:
This PR removes a green background that had been applied to a stack view in the LoginSiteAddressViewController, and bumps the podspec version.  See screenshot:

Screen Shot 2020-10-26 at 12 06 22 PM

To test: 
Confirm the stack view background is no longer green? 
Double check that I formatted the podspec beta version correctly 😬 

@ScoutHarris could I trouble you for a review?

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.27.1-beta"
+  s.version       = "1.28.0-beta.1"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.27.0"
+  s.version       = "1.27.1-beta"
```
ScoutHarris: `1.27.0` is the last release version, and we now leave the minor version for hotfixes, so this should probably be `1.28.0-beta.1`. (We drop the beta version in the host apps, i.e. WordPress will be `1.28.0-beta`.)

aerych: Thank you!

ScoutHarris (approved):
I see no green.

:shipit:

aerych:
Thank you!!!

PR #507, opened 2020-10-26
Issues/configure instruction text

aerych:
This PR makes the instruction text on the LoginUsernamePasswordViewController customizable/configurable.  Up to this point the text was hard coded in the Storyboard (and not localized 😱).  With these changes it can be configured via the corresponding displayString and can be localized. I've introduced this change in order to configure the text for NiOS.

I've opted to make the default text generic rather than reference WooCommerce directly for consistency with the other strings in the auth library.  This will require a follow up patch to WooCommerceiOS to keep the text consistent until it implements the unified auth flow.

To Test:
Test the companion [WCiOS PR](https://github.com/woocommerce/woocommerce-ios/pull/3061) and confirm the instructions can be configured.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.28.0-beta.1"
+  s.version       = "1.28.0-beta.2"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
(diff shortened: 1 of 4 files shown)


ScoutHarris (approved):
🌟 With these changes it can be configured via the corresponding displayString and can be localized.
🌟 I've opted to make the default text generic rather than reference WooCommerce directly

✅ Test the companion WCiOS PR and confirm the instructions can be configured.

:shipit:

aerych:
Woot!  Thank you @ScoutHarris !

PR #516, opened 2020-11-18
Login Prologue button view: make background color a parameter

### LINKED ISSUE #514: Prologue: customize background of buttonViewController
Accept a parameter to change the background of the view that contains the action buttons in LoginPrologueViewController.

This will help with the implementation of woocommerce/woocommerce-ios#3144

Expected changes:

Add an optional parameter to WordPressAuthenticatorUnifiedStyle to signal the background color, defaulting to clear.
Update LoginPrologueViewController to assign the background color passed as parameter

ctarda:
Closes #514 
Ref: woocommerce/woocommerce-ios#3144

This PR:
* adds a property to WordPressAuthenticatorStyle for the background color with default value of `.clear` . This should not break the API (at least it didn't for WooCommerce)
* sets the background color of LoginViewController's buttonViewController to the property passed via WordPressAuthenticatorStyle, defaulting to `.clear` again if unified styles is nil.

Note: I tried to add a test but I didn't find the proper place to do so. Let me know if that's something you would like me to explore, please.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorStyles.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorStyles.swift
@@ -192,6 +192,9 @@ public struct WordPressAuthenticatorUnifiedStyle {
     /// Style: Auth view background colors
     ///
     public let viewControllerBackgroundColor: UIColor
+    
+    /// Style: Auth Prologue buttons background color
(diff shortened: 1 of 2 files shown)


mindgraffiti (approved):
✅  Code
✅  Tested on WPiOS and saw no change.

I'm not sure it would be useful to build unit tests around the Style class, so I didn't expect any to be written.

PR #520, opened 2020-11-23
Partially expose navigation logic to clients of WPAuthenticator

### LINKED ISSUE #515: WordPressAuthenticatorDelegate: Allow insertion of a View Controller in the navigation stack in the completion block for shouldPresentUsernamePasswordController
We want to be able to provide an optional UIViewController as a third parameter to the completion closure of `shouldPresentUsernamePasswordController`

The goal is to provide a way for client apps to customise the UI after a login with site address has been completed, adding a third option to the current navigation options (showSelfHostedUsernamePassword and showWPUsernamePassword). This, for example, would help with the implementation of woocommerce/woocommerce-ios#3161 

This seems like it might require a breaking change to the WordPressAuthenticationDelegate API. If that's the case, it would require updating client apps. 

Expected changes:
* Add the parameter to the completion block of shouldPresentUsernamePasswordController
* Update SiteAddressViewController in a way that if a View Controller is provided (and if other state flags match), navigate to the View controller provided.

ctarda:
Closes #515 

Ref: woocommerce/woocommerce-ios#3161
Ref: p91TBi-3ON-p2

This PR provides a way for clients of WPAuthenticator to insert custom UI into WPAuthenticator's navigation flow in a way that:
* Hands over control to client apps, so those client apps can pass UI back to WPAuthenticator.
* Does not require modifications to the current implementation of WPAuthenticator, but only adds new code to it
* Leaves the responsibility of navigating out of the custom UI to WPAuthenticator
* Does not break much of the existing API.

## Changes
A simple implementation of the [command pattern](https://en.wikipedia.org/wiki/Command_pattern): 
* `NavigationCommand` is the actual command, and it is modeled as a protocol
* Two implementations of NavigationCommand: `NavigateToEnterSite` and `NavigateToEnterAccount`. These two structs lift and duplicate some of the existing code in `SiteAddressViewController`, to navigate, respectively, to the root of the navigation flow that allows logging in with a site address and a wordpress.com account.
* Modifies the signature of the method `shouldPresentUsernamePasswordController` in `WordPressAuthenticatorDelegate`. This modification breaks backwards compatibility. (see more about it in a comment to the PR)

## Testing
Changes break the API of WPAuthenticatorDelegate so this is a bit tricky to test in isolation.

I have submitted a draft PR to WordPress-iOS that integrates with the breaking changes in this PR, and provides a way to test that the changes I am submitting now do not break the existing login flow: wordpress-mobile/WordPress-iOS#15385

* Checkout WordPress-iOS branch `try/ctarda/auth`
* Run `bundle exec pod install`. That will point WPAuthenticator to this branch
* Build and run WordPress-iOS. 
* Log out if necessary.
* Attempt to login with both a WordPress.com account and a self-hosted site. Check that logging in is not broken.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -48,7 +48,7 @@ public protocol WordPressAuthenticatorDelegate: class {
     ///     - site: passes in the site information to the delegate method.
     ///     - onCompletion: Closure to be executed on completion.
     ///
-    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, from: UIViewController?, onCompletion: @escaping (Error?, Bool) -> Void)
+    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (Error?, Bool, UIViewController?) -> Void)
 
     /// Presents the Login Epilogue, in the specified NavigationController.
     ///
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -48,7 +48,7 @@ public protocol WordPressAuthenticatorDelegate: class {
     ///     - site: passes in the site information to the delegate method.
     ///     - onCompletion: Closure to be executed on completion.
     ///
-    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (Error?, Bool, UIViewController?) -> Void)
+    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (WordPressAuthenticatorResult) -> Void)
 
     /// Presents the Login Epilogue, in the specified NavigationController.
     ///
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -48,7 +48,7 @@ public protocol WordPressAuthenticatorDelegate: class {
     ///     - site: passes in the site information to the delegate method.
     ///     - onCompletion: Closure to be executed on completion.
     ///
-    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, navigationController: UINavigationController?, onCompletion: @escaping (Error?, Bool) -> Void)
+    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, from: UIViewController?, onCompletion: @escaping (Error?, Bool) -> Void)
 
     /// Presents the Login Epilogue, in the specified NavigationController.
     ///
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -48,7 +48,7 @@ public protocol WordPressAuthenticatorDelegate: class {
     ///     - site: passes in the site information to the delegate method.
     ///     - onCompletion: Closure to be executed on completion.
     ///
-    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (Error?, Bool) -> Void)
+    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, navigationController: UINavigationController?, onCompletion: @escaping (Error?, Bool) -> Void)
 
     /// Presents the Login Epilogue, in the specified NavigationController.
     ///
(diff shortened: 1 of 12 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
```diff
@@ -48,7 +48,7 @@ public protocol WordPressAuthenticatorDelegate: class {
     ///     - site: passes in the site information to the delegate method.
     ///     - onCompletion: Closure to be executed on completion.
     ///
-    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (Error?, Bool) -> Void)
+    func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (Error?, Bool, UIViewController?) -> Void)
```
ctarda: This would break WPAuhenticator's public API. 

Also, the way I understand it, these three parameters are mutually exclusive. So it would probably be better if the signature of this method was actually something like:

`func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (WPAuthenticatorResult) -> Void)`

Where:
```
enum WPAuthenticatorResult {
    case error(value: Error)
    case presentPasswordController(value: Bool)
    case injectViewController(value: UIViewController)
}
```

or similar.

WordPressAuthenticator/Navigation/NavigateToEnterAccount.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+public struct NavigateToEnterAccount: NavigationCommand {
+    public init() {}
+    public func execute(from: UIViewController?) {
```
ctarda, line 7: WPAuthenticator inserts the actual view controller that triggers this command, so WPAuthenticator knows that this requires manipulating a navigation controller

WordPressAuthenticator/Navigation/NavigateToEnterAccount.swift
```diff
@@ -0,0 +1,22 @@
...
+        continueWithDotCom(navigationController: from?.navigationController)
+    }
+}
+
+
+private extension NavigateToEnterAccount {
+    /// Unified "Continue with WordPress.com" prologue button action.
+    ///
+    private func continueWithDotCom(navigationController: UINavigationController?) {
```
ctarda, line 14: This is lifted from SiteAddressViewController

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -468,7 +468,13 @@ private extension SiteAddressViewController {
             return
         }
         
-        WordPressAuthenticator.shared.delegate?.shouldPresentUsernamePasswordController(for: siteInfo, onCompletion: { (error, isSelfHosted) in
+        WordPressAuthenticator.shared.delegate?.shouldPresentUsernamePasswordController(for: siteInfo, onCompletion: { (error, isSelfHosted, injectedViewController) in
+            // Shortcircuit if clients provide a view controller to be inserted in the stack
+            if let customUI = injectedViewController {
```
ctarda: If client apps inject a view controller, push it to the current navigation controller. If not, fallback to the previous logic

This is why an enumeration, might make more sense here.

ctarda:
Pinging @mindgraffiti and @ScoutHarris

WordPressAuthenticator/Signin/LoginSiteAddressViewController.swift
```diff
@@ -226,7 +226,7 @@ class LoginSiteAddressViewController: LoginViewController, NUXKeyboardResponder
     }
 
     func presentNextControllerIfPossible(siteInfo: WordPressComSiteInfo?) {
-        WordPressAuthenticator.shared.delegate?.shouldPresentUsernamePasswordController(for: siteInfo, onCompletion: { (error, isSelfHosted) in
+        WordPressAuthenticator.shared.delegate?.shouldPresentUsernamePasswordController(for: siteInfo, onCompletion: { (error, isSelfHosted, injectedViewController) in
             guard let originalError = error else {
```
ScoutHarris, line 230: `LoginSiteAddressViewController` is the old auth flow (i.e. it will be nixed in the future). This should be done in`SiteAddressViewController`, which is the new, unified Site Address VC.

ScoutHarris: Sorry. I deleted the first comment. I see now what this is doing.

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
ScoutHarris: In WPiOS, `shouldPresentUsernamePasswordController` just calls `onCompletion`. 

```
func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (Error?, Bool) -> Void) {
    onCompletion(nil, true)
}
```

So fixing it should be straightforward.

> So it would probably be better if the signature of this method was actually something like

That sounds reasonable.

ctarda:
@ScoutHarris @mindgraffiti  this is ready for review. 

To (hopefully) facilitate testing, I have submitted a draft PR to WordPress-iOS that points the dependency with WPAuthenticator to this branch, to:
1. Prove that nothing breaks
2. Take as a starting point to integrate the next release of WPAuthenticator (whenever that happens)

ScoutHarris:
Hey @ctarda .

Two things:
1. The WPAuth podspec version needs to be updated. Assuming your prior PR will be merged first, this one should be "1.30.0-beta.2".
2. The WPiOS does need to be merged since this includes a breaking change and that will be needed so WP doesn't break.

ScoutHarris (approved):
Once the podspec version is updated, :shipit: .

ctarda:
Thanks @ScoutHarris . The podspec is updated, merging now!

PR #522, opened 2020-11-25
WordPressComSiteInfo: add properties and update parser

### LINKED ISSUE #521: WordPressComSiteInfo initializer is missing data
WordPressComSiteInfo is initialized with a raw dictionary.

The initializer assigns the site url like this:
`url         = remote["URL"] as? String          ?? ""`

It seems, however, that the url is contained in the key `urlAfterRedirects`

Also, the dictionary passed to this initializer contains keys for `isJetpackActive`, `hasJetpack`, `isWordPress`, `isJetpackConnected` and `exists`. Most (if not all) of these keys will be useful for the integration of WooCommerce's Unified Login

ctarda:
Closes #521 

## Changes
* Add some properties to WordPressComSiteInfo to provide more information related to the state of the Jetpack installation. This will be very helpful for WooCommerce
* Update parser. It seems like the payload key mapping  to url is `urlAfterRedirects`
* Add tests to cover the parser

## Testing
* 👀  on the code
* If CI is ✅  we should we good to go. This should not introduce any breaking changes to WP

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.29.0"
+  s.version       = "1.30.0-beta.1"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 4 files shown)


ScoutHarris (approved):
The podspec version should be updated to "1.30.0-beta.1". After that, 👍 .

ctarda:
Thanks @ScoutHarris . I updated the podspec, now that CI is ✅ I'll merge.

PR #524, opened 2020-11-27
Add a command to pop to the root of the login flow

### LINKED ISSUE #523: Add a NavigationCommand to reset navigation to the first screen of the flow
In order to accommodate the new Unified Login flow in WooCommerce iOS, we need to be able to navigate back to the Login Prologue from one of the error screens specific to WooCommerce.

We would implement that by adding a new implementation of NavigationCommand that pops to the root of the Navigation controller

ctarda:
Closes #523 
Ref: woocommerce/woocommerce-ios#3186

Continuing with the solution discussed in p91TBi-3ON-p2 , that we started to implement in PR #520 , this PR adds one more implementation of the NavigationCommand. In this case, the new command just pops to the root of the navigation controller (and therefore to the LoginPrologue)

## Changes
* Added a new implementation of the NavigationCommand protocol
* This implementation pops to the root of the navigation controller. That seemed like the simplest way to "reset" the Login flow to its entry point, but please let me know if there is a better way. I looked at potential issues with shared state, but I couldn't find any 🤷‍♂️

## How to test
This does not add any breaking changes to the API, and the change is additive, so there is no chance to break anything. 

The best way to test would be testing the PR in WooCommerce that uses this code: woocommerce/woocommerce-ios#3186

That being said, I added a unit test to cover the new command. So if CI is ✅ we should be good to go

CC @pmusolino (who is reviewing the counter part PR in WooCommerce: woocommerce/woocommerce-ios#3186

### DIFF
--- a/WordPressAuthenticator.xcodeproj/project.pbxproj
+++ b/WordPressAuthenticator.xcodeproj/project.pbxproj
@@ -158,10 +158,12 @@
 		D85C36E6256E0DDE00D56E34 /* NavigationToEnterSiteTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D85C36E5256E0DDE00D56E34 /* NavigationToEnterSiteTests.swift */; };
 		D85C36EC256E10EA00D56E34 /* MockNavigationController.swift in Sources */ = {isa = PBXBuildFile; fileRef = D85C36EB256E10EA00D56E34 /* MockNavigationController.swift */; };
 		D85C36F0256E118D00D56E34 /* NavigationToEnterAccountTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D85C36EF256E118D00D56E34 /* NavigationToEnterAccountTests.swift */; };
+		D85C3882256E3FEC00D56E34 /* WordPressComSiteInfoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D85C3881256E3FEC00D56E34 /* WordPressComSiteInfoTests.swift */; };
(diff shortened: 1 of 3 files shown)


mindgraffiti:
@ScoutHarris I trust your navigation logic better than mine. If a future WPiOS dev uses this nifty feature, would it cause a bug on this edge case?

<a href="https://user-images.githubusercontent.com/1062444/100653830-935f7f00-330e-11eb-95d2-5e04fe1b396f.png"></a>

If you're logged into a self-hosted site and attempt to additionally log into a WP.com or Jetpack-connected site, the prologue VC is presented as a modal view instead of being pushed. I could see a future WPiOS dev not knowing about the edge case and blindly using the `NavigateToRoot` command.

I'm asking in case a comment needs added to the method (to warn away WP devs). Or maybe I'm overthinking it ¯\_(ツ)_/¯

mindgraffiti (approved):

ScoutHarris:
Hey @mindgraffiti .

> If a future WPiOS dev uses this nifty feature, would it cause a bug on this edge case?

Sidenote - it's also presented this way from the app delegate if the default account changes. So... maybe 2 edge cases.

To answer your question, to be blunt, only if they weren't paying any attention at all. 😄 Unless I'm underthinking it. But the prologue is presented modally via a very specific method (`WordPressAuthenticator.showLogin`) that is meant to be used in these scenarios. So I think it's clear what's meant to happen. If a dev changes the method used, or changes how this method works, there are very clearly two places it will affect so I would expect appropriate testing/verification would happen.

But I won't argue against comments if you feel it's helpful. I just don't think it's necessary. My 2 cents.

PR #525, opened 2020-11-27
Add new methods to WPAuthenticatorDelegate to allow host apps to handle errors

### LINKED ISSUE #530: Allow host apps to handle errors
At the moment most of the most common errors (i.e. "the site is not a WP site") are handled by WPAuthenticator.

Host apps might want to take over some of those errors. In fact, WooCommerce wants to be able to know about certain errors and want to provide custom ui for them. (see, for reference, woocommerce/woocommerce-ios#3218)

ctarda:
Closes #530 
Ref: p91TBi-3Rn-p2

## Changes
* Added two new methods to `WPAuthenticatorDelegate` to allow host apps to decide if they want to handle errors.
* Bumped the podspec
* Extracted a private method in SiteAddressViewController to execute the insertion of custom UI

## How to test
* This PR provides the functionality for woocommerce/woocommerce-ios#3219 so the best way to test is probably by following the test steps in the linked PR to WooCommerce.
* Please inspect the diff. Let me know if there is a better way 😊


WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
```diff
@@ -68,6 +68,14 @@ public protocol WordPressAuthenticatorDelegate: class {
     ///
     func shouldPresentLoginEpilogue(isJetpackLogin: Bool) -> Bool
 
+    /// Indicates the Host app wants to handle and display a given error.
```
ctarda, line 71: These two methods would be new. This would break WordPress, but the fix is simple: for host apps that are good with WPAuthenticator's error handling, shoudlDisplayError would just return false, and handleError would do nothing

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -140,6 +140,24 @@ final class SiteAddressViewController: LoginViewController {
         }
     }
 
+    override func displayRemoteError(_ error: Error) {
```
ctarda, line 143: This might not be necessary. In fact, I would argue that it would be good to avoid doing it.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -419,6 +437,21 @@ private extension SiteAddressViewController {
 
                 let err = self.originalErrorOrError(error: error as NSError)
 
+                if self.authenticationDelegate.shouldDisplayError(err) {
```
ctarda: This is what actually inserts the custom UI for the case when the url provided for login with site address is not a WordPress site. This code would use some cleanup, but I hope it helps illustrate how this solution would work

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
```diff
@@ -68,6 +68,14 @@ public protocol WordPressAuthenticatorDelegate: class {
...
+    ///
+    func shouldDisplayError(_ error: Error) -> Bool
+
+    /// Signals the Host app that there is an error that needs to be handled.
+    ///
+    func handleError(_ error: Error, onCompletion: @escaping (WordPressAuthenticatorResult) -> Void)
```
ctarda: The parameter in the completion block might just be of type UIViewController. After all, host apps would only call this method if they actually want to inject a custom error UI

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
ScoutHarris: Thanks for the info! After this is merged, I'll create a WP PR to address this.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
ScoutHarris: I'm sorry, but I'm confused by this. It appears you added this override method to implement your error handling. Why do you not agree with what you just did? 😄

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -140,6 +140,17 @@ final class SiteAddressViewController: LoginViewController {
         }
     }
 
+    override func displayRemoteError(_ error: Error) {
+        guard authenticationDelegate.shouldHandleError(error) == true else {
```
ScoutHarris: Can this just be `guard authenticationDelegate.shouldHandleError(error) else {`?

mindgraffiti (approved):
This approach seems reasonable to me, though I am curious about your answer to Stephenie's question (about why you don't like overriding the displayRemoteError) 😀

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
mindgraffiti: iirc, this is part of Woo's coding style. I could be wrong though.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
ctarda: Sorry, that was terrible wording on my side. Let me rephrase my comment:

"I am not 100% sure that this is absolutely necessary. So far I have not seen this override being used in any of the tests I have run using WooCommerce. That does not mean it won't happen as I continue testing. I would prefer if, eventually, we noticed that overriding this method is actually unnecessary and we could get rid of the override"

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
ctarda: Yes, it's probably muscle memory. I'll update it

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
ctarda: And I just found a test case for WooCommerce where these overrides are necessary: a Pressable store with Jetpack installed but not activated

ctarda:
Thanks for the review @mindgraffiti @ScoutHarris 

I bumped the podspec, updated the guard clause and hopefully clarified my comment about the override. 

I'll merge when CI is ✅

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
ScoutHarris: Oh! OK, that makes more sense. 😄 Thank you sir!

PR #529, opened 2020-11-30
Adjust the WordPressUI Dependency definition

jkmassel:
Right now it’ll only accept `1.7.x` versions, and we want to accept anything between 1.x.x and 2.0, including betas.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -39,7 +39,7 @@ Pod::Spec.new do |s|
 
   s.dependency 'Gridicons', '~> 1.0'
   s.dependency 'GoogleSignIn', '~> 5.0.2'
-  s.dependency 'WordPressUI', '~> 1.7.0'
+  s.dependency 'WordPressUI', '~> 1.7-beta'
 
   # Use a loose restriction that allows both production and beta versions, up to the next major version.
   # If you want to update which of these is used, specify it in the host app.


loremattei (approved):
LGTM

WordPressAuthenticator.podspec
```diff
@@ -39,7 +39,7 @@ Pod::Spec.new do |s|
 
   s.dependency 'Gridicons', '~> 1.0'
   s.dependency 'GoogleSignIn', '~> 5.0.2'
-  s.dependency 'WordPressUI', '~> 1.7.0'
+  s.dependency 'WordPressUI', '~> 1.7-beta'
```
AliSoftware, line 42: I was not entirely sure if the `-beta` suffix would mean it would propagate to higher versions (i.e. if we interpret `~> 1.7-beta` to be "anything `> 1.7.0-beta` but lower than `2.0`", would it consider `1.8.0-beta` to be valid or would it interpret it as "any stable version above that one provided, even if that one provided is a beta")

So I went up and checked just to be sure and you'll be happy to know that thankfully it works as we want here:

```ruby
# If we were to use a requirement without the `-beta`
[1] pry(main)> req = Gem::Requirement.new('~> 1.7')
=> Gem::Requirement.new(["~> 1.7"])
[2] pry(main)> req.satisfied_by?(Gem::Version.new('1.7-beta'))
=> false
[3] pry(main)> req.satisfied_by?(Gem::Version.new('1.8.1-beta.1'))
=> true
```
```ruby
# When using requirement with the `-beta`, it still accept higher betas too
[1] pry(main)> req = Gem::Requirement.new('~> 1.7-beta')
=> Gem::Requirement.new(["~> 1.7.pre.beta"])
[2] pry(main)> req.satisfied_by?(Gem::Version.new('1.7.1-beta.1'))
=> true
[3] pry(main)> req.satisfied_by?(Gem::Version.new('1.8.0'))
=> true
[4] pry(main)> req.satisfied_by?(Gem::Version.new('1.8.1-beta.1'))
=> true
[5] pry(main)> req.satisfied_by?(Gem::Version.new('2.1.0'))
=> false
```

AliSoftware (approved):

PR #532, opened 2020-12-01
Add a command to navigate back one step

### LINKED ISSUE #531: Add navigation command to allow host apps to pop the navigation controller
Add an implementation of the NavigationCommand protocol that navigates back just one step. 

Necessary for woocommerce/woocommerce-ios#3218

ctarda:
Closes #531 

## Changes
* A new implementation of the NavigationCommand protocol that pops a navigation command once. So the functionality would be equivalent to the Back button in a navigation controller.
* A unit test

## How to test
This does not introduce any breaking changes, so a ✅ on CI should be enough. That being said, it can be tested in conjunction with woocommerce/woocommerce-ios#3219

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.30.0"
+  s.version       = "1.31.0-beta.1"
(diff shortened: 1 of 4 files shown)


ScoutHarris:
The pod version needs to be bumped to `1.31.0-beta`.

ScoutHarris (approved):
After the pod version is fixed, :shipit: .

PR #534, opened 2020-12-04
Allow host apps to insert custom UI instead of WP.com account creation flow

### LINKED ISSUE #533: Allow host apps to bail out of new account creation flow
Ref: woocommerce/woocommerce-ios#3146

Open an extension point to allow host apps to take over the navigation flow right before navigating to the WordPress.com account creation flow.

ctarda:
Closes #533 
Ref: woocommerce/woocommerce-ios#3146

## Changes
Building up upon the functionality implemented in PR #525 , this PR touches GetStartedViewController, so that:
* When we detect an error after a user attempts log in with an email address, we would ask the host app if it wants to handle this error and intercept the navigation flow
* If the host app wants to handle the error, we would insert the UI that the host app provides. 
* If the host app does not want to handle the error, we continue forward without changes in behaviour.
* Bumped the version in the podspec

## How to test
This PR does not break the public API, but it has the potential to introduce a bug in WordPress. There are two ways I would suggest to test the functionality:
1. Test the corresponding PR in WooCommerce (woocommerce/woocommerce-ios#3286). When users try to log in with a WordPress account and enter an email address that does not match a WP.com account, WooCommerce should insert custom UI. 
2. Test the same flow with WordPress. No behaviour should change and it should still be possible to sign up for a WP.com account. The easiest way to test this is testing the draft PR wordpress-mobile/WordPress-iOS#15435

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.31.0-beta.2"
+  s.version       = "1.31.0-beta.3"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


mindgraffiti (approved):
✅  Tested with WPiOS draft PR
✅  Code

:shipit: !

PR #537, opened 2020-12-08
Reduces the number of Pasteboard alerts displayed by the 2FA screen

bjtitus:
Addresses most of https://github.com/wordpress-mobile/WordPress-iOS/issues/15438.
WordPress PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15454

* Adds new UIPasteboard detection methods on iOS 14.
* Uses `changeCount` instead of the contents to detect pasteboard changes.

### Testing

* Enter username and password for an account with 2FA enabled.
* On the 2FA screen leave the app - You should _not_ see any pop ups about pasteboard usage when leaving the app.
* Copy a URL from Safari.
* Come back to the app - You should _not_ see any alert for pasteboard usage.
* Leave the app and copy a 2FA code.
* Return to the app - You _should_ see an alert for pasteboard usage and the code should be pasted into the field.


WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
```diff
@@ -0,0 +1,24 @@
+extension UIPasteboard {
+
+    /// Detects patterns and values from the UIPasteboard. This will not trigger the pasteboard alert in iOS 14.
+    /// - Parameters:
+    ///   - patterns: The patterns to detect.
+    ///   - completion: Called with the patterns and values if any were detected, otherwise contains the errors from UIPasteboard.
+    @available(iOS 14.0, *)
+    func detect(patterns: Set<UIPasteboard.DetectionPattern>, completion: @escaping (Result<[UIPasteboard.DetectionPattern: Any], Error>) -> Void) {
+        UIPasteboard.general.detectPatterns(for: [.number]) { [weak self] result in
```
frosty: Should we be using the passed in `patterns` value here, rather than hardcoding it for numbers? Same for the call to `detectValues` below.

WordPressAuthenticator/Signin/Login2FAViewController.swift
```diff
@@ -283,19 +284,39 @@ class Login2FAViewController: LoginViewController, NUXKeyboardResponder, UITextF
...
+            UIPasteboard.general.detect(patterns: [.number]) { [weak self] result in
+                switch result {
+                case .success(let detections):
+                    if let pasteString = detections.first?.value as? String {
+                        DispatchQueue.main.async {
+                            self?.handle(code: pasteString)
+                        }
+                    }
+                case .failure(let error):
+                    ()
```
frosty: Maybe just `break` here?

WordPressAuthenticator/Signin/Login2FAViewController.swift
```diff
@@ -13,7 +13,8 @@ class Login2FAViewController: LoginViewController, NUXKeyboardResponder, UITextF
     @IBOutlet weak var sendCodeButton: UIButton!
     @IBOutlet var bottomContentConstraint: NSLayoutConstraint?
     @IBOutlet var verticalCenterConstraint: NSLayoutConstraint?
-    @objc var pasteboardBeforeBackground: String? = nil
+
+    private var pasteboardChangeCountBeforeBackground: Int? = nil
```
frosty: Is there any reason we don't have this as an `Int = 0`?

WordPressAuthenticator/Signin/Login2FAViewController.swift
```diff
@@ -283,19 +284,39 @@ class Login2FAViewController: LoginViewController, NUXKeyboardResponder, UITextF
...
                 return
         }
 
-        switch isValidCode(code: pasteString) {
+        if #available(iOS 14.0, *) {
+            UIPasteboard.general.detect(patterns: [.number]) { [weak self] result in
+                switch result {
+                case .success(let detections):
+                    if let pasteString = detections.first?.value as? String {
+                        DispatchQueue.main.async {
```
frosty: We could dispatch the call to the completion block in our category, so we don't need to do it here or in the other view controller?

frosty (commented):
Works as expected, just a couple of code comments for you!

WordPressAuthenticator/Signin/Login2FAViewController.swift
bjtitus: I usually prefer to keep the Main queue dispatch at the call site since other users might not want to dispatch to main right away.

Since this is only used in one spot, though, I'll just make the change in the two success blocks.

bjtitus:
@wordpress-mobile/platform-9 I'm wondering if the Xcode 12 changes I made look good and whether there's a good way to fix the pod linter warnings.

mokagio (commented):
@bjtitus your changes look sound to me.

Here's my understanding:

The `post_install` hook addresses the minimum deployment target warning that Xcode 12 throws now that it expects an higher value than before. It also takes care of the arm64 Simulator issue now that Xcode needs to support both ARM and Intel. That's similar, probably better actually, than what I ended up doing [for Sodium in Tracks](https://github.com/Automattic/Automattic-Tracks-iOS/pull/148), that is, forking the library and updating its `podspec`

The `podspec` update is what we need to avoid consumers of this library to incur in the arm64 Simulator issue because of it.

Same goes for the project change, although I wonder why that was required to build 🤔 

You can address the `podspec` validation by updating CocoaPods to the be compatible with 1.10 (`'~> 1.10'`). That's the version that introduced fixes for Xcode 12 and arm64 stuff.

With that update, `pod lib lint` fails, but because of 1 warning only. I think it could be ignored for now. Here's the full output:

```
pod lib lint

 -> WordPressAuthenticator (1.31.0-beta.3)
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | xcodebuild:  note: Building targets in parallel
    - NOTE  | xcodebuild:  note: Using codesigning identity override: -
    - NOTE  | [iOS] xcodebuild:  note: Planning build
    - NOTE  | [iOS] xcodebuild:  note: Constructing build description
    - WARN  | [iOS] xcodebuild:  /Users/gio/Developer/a8c/wordpressauthenticator-ios/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift:18:74: warning: variable 'self' was written to, but never read
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'GoogleSignIn' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: MobileCoreServices has been renamed. Use CoreServices instead. (in target '1PasswordExtension' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'AppAuth' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'GTMAppAuth' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'NSURL+IDN' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'UIDeviceIdentifier' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'SVProgressHUD' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file and one is not being generated automatically. (in target 'App' from project 'App')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'Alamofire' from project 'Pods')
    - NOTE  | [iOS] xcodebuild:  warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.2.99. (in target 'FormatterKit' from project 'Pods')

[!] WordPressAuthenticator did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it).
You can use the `--no-clean` option to inspect any issue.
```

I'm not actually sure which warning is the failing one, but I'd say "MobileCoreServices has been renamed. Use CoreServices instead. (in target '1PasswordExtension' from project 'Pods')", simply because if it was the deployment targets one, it the output wouldn't say "1 warning", right? 🤷

frosty (approved):
Changes look good to me!

bjtitus:
@mokagio Thank you so much for taking a look!

Turns out it was an actual compiler warning causing validation to fail. The one not marked with `[warning]` but `WARN` instead. 🤦 

```
- WARN  | [iOS] xcodebuild:  /Users/gio/Developer/a8c/wordpressauthenticator-ios/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift:18:74: warning: variable 'self' was written to, but never read
```

Sorry for the confusion!
I went ahead and updated to the latest cocoapods and ruby versions since there are those Xcode 12 fixes in 1.10.0.

WordPressAuthenticator.podspec
```diff
@@ -45,4 +45,8 @@ Pod::Spec.new do |s|
   # If you want to update which of these is used, specify it in the host app.
   s.dependency 'WordPressKit', '~> 4.18-beta' # Don't change this until we hit 5.0 in WPKit
   s.dependency 'WordPressShared', '~> 1.12-beta' # Don't change this until we hit 2.0 in WPShared
+
+  # Fixing arm64 issue with Xcode 12: https://github.com/CocoaPods/CocoaPods/issues/10104
+  s.user_target_xcconfig = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
```
jkmassel, line 50: This change makes this pod incompatible with Xcode on Apple silicon 🤔 

Did the CocoaPods update make this unnecessary?

jkmassel (approved):
Because this is a bit of a rush it's ok to go out, but I'm concerned that our solution breaks Xcode compatibility – hopefully we can address this soon?

WordPressAuthenticator.podspec
bjtitus: > Did the CocoaPods update make this unnecessary?

Unfortunately not 😢 

There end up being linker errors during linting:

```
    ld: in /var/folders/jb/jdc8rjfx4sx60n_0xnclhtc00000gn/T/CocoaPods-Lint-20201211-68458-riqq0i-WordPressAuthenticator/Pods/GoogleSignIn/Frameworks/GoogleSignIn.framework/GoogleSignIn(GIDEMMErrorHandler_3a47e13d8ca81b41e9cdb7ef5468004a.o), building for iOS Simulator, but linking in object file built for iOS, file '/var/folders/jb/jdc8rjfx4sx60n_0xnclhtc00000gn/T/CocoaPods-Lint-20201211-68458-riqq0i-WordPressAuthenticator/Pods/GoogleSignIn/Frameworks/GoogleSignIn.framework/GoogleSignIn' for architecture arm64
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

PR #544, opened 2020-12-13
Display a spacer cell instead of TOS when host apps remove the link to WP.com terms of service.

### LINKED ISSUE #543: Add padding between email and continue button when TOS is hidden
Ref:  woocommerce/woocommerce-ios#3309
Ref:  woocommerce/woocommerce-ios#3330

PR #536 implemented support for host apps to remove the TOS button in GetStartedViewControlle.

The way I implemented that, however, was by not inserting the TOS cell into the table view, which makes the email cell sit right above the Continue button, without any separation.

The goal of this issue would be to find a way to remove the TOS from the UI but leave the space the TOS cell was using to serve as separation between the email field and the continue button

ctarda:
Closes #543 
Ref: woocommerce/woocommerce-ios#3309
Ref: woocommerce/woocommerce-ios#3330
PR in WooCommerce: woocommerce/woocommerce-ios#3339

The way I implemented the removal of the TOS was by just removing the cell from the table view. That makes the bottom hairline in the email textfield cell to sit right above the top margin of the Continue button. 

I considered just hiding the TOS button instead of removing the cell completely (you can see that if you look at this very PR history) but in the end I thought a Spacer cell (a completely empty one) might be a better way to handle this.

## Changes
* Add a spacer cell
* Declare a new section in GetStartedViewController to model the spacer
* Check if the host app wants to display TOS or not. If it doesn't, add a spacer.

## How to test
* Look at the code. Does the change make sense?
* Check that the corresponding PRs in WooCommerce and WordPress work as expected.
- WooCommerce: woocommerce/woocommerce-ios#3339
- WordPress: wordpress-mobile/WordPress-iOS#15486

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.31.0-beta.5"
+  s.version       = "1.31.0-beta.6"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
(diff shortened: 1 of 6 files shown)


mindgraffiti (approved):
I think you will want to un-check `User Interaction Enabled` for the spacer cell, otherwise VoiceOver will attempt to identify it. I'm unable to test and confirm this claim because my device isn't registered for Woo use.

Screen Shot 2020-12-14 at 10 03 26 AM

Other than that, :shipit: !

WordPressAuthenticator.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.31.0-beta.5"
+  s.version       = "1.31.0-beta.6"
```
ScoutHarris: Since the new release was cut today, this should probably be `1.32.0-beta.1` after @jkmassel finishes the releases.

ctarda:
Thanks @mindgraffiti @ScoutHarris I'll merge develop into the branch, un-check userInteractionEnabled, update the podspec and merge 😊

PR #549, opened 2020-12-16
Release 1.32.0 in trunk

AliSoftware:


WordPressAuthenticator.podspec
```diff
@@ -39,10 +39,10 @@ Pod::Spec.new do |s|
 
   s.dependency 'Gridicons', '~> 1.0'
   s.dependency 'GoogleSignIn', '~> 5.0.2'
-  s.dependency 'WordPressUI', '~> 1.7-beta'
 
   # Use a loose restriction that allows both production and beta versions, up to the next major version.
   # If you want to update which of these is used, specify it in the host app.
+  s.dependency 'WordPressUI', '~> 1.7-beta'
```
AliSoftware, line 45: Moved this here to make it clear that the `-beta` is expected for this one too for the same reason as WPKit and WPShared, explained by the above comment (did this since I almost bumped it to a non-beta when I did that release, but then remembered #529, so felt like a good idea to make it clear)

JavonDavis (approved):
LGTM! 🚢

PR #553, opened 2020-12-23
Fix magic link issues.

diegoreymendez:
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15557
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/322

Please refer to the WPiOS PR for details and testing steps.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -134,8 +134,8 @@ import WordPressKit
 
     /// Attempts to process the specified URL as a WordPress Authentication Link. Returns *true* on success.
     ///
-    @objc public func handleWordPressAuthUrl(_ url: URL, rootViewController: UIViewController) -> Bool {
-        return WordPressAuthenticator.openAuthenticationURL(url, fromRootViewController: rootViewController)
+    @objc public func handleWordPressAuthUrl(_ url: URL, rootViewController: UIViewController, automatedTesting: Bool = false) -> Bool {
+        return WordPressAuthenticator.openAuthenticationURL(url, fromRootViewController: rootViewController, automatedTesting: automatedTesting)
     }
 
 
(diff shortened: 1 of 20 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -309,68 +305,91 @@ import AuthenticationServices
...
 
-
     /// Present a signin view controller to handle an authentication link.
     ///
     /// - Parameters:
     ///     - url: The authentication URL
-    ///     - rootViewController: The view controller to act as the presenter for the signin view controller.
-    ///                           By convention this is the app's root vc.
+    ///     - rootViewController: The view controller to act as the presenter for the signin view controller.  By convention this is the app's root vc.
+    ///     - automatedTesting: for calling this method for automated testing.  It won't sync the account or load any other VCs.
```
Gio2018, line 313: I wonder if there's another way to call this method in unit tests without using a flag? Perhaps we can inject a `Storyboard` instance, which could be nil in case of testing? Not sure if it's possible in this case, but the flag might be harder to maintain and could lead to bugs that are not immediately visible.

WordPressAuthenticator/Credentials/WordPressComCredentials.swift
```diff
@@ -19,17 +19,19 @@ public struct WordPressComCredentials: Equatable {
     /// The site address used during login
     ///
     public var siteURL: String
-
-    /// Designated initializer
+    
+    private let wpComURL = "https://wordpress.com"
```
Gio2018, line 23: "https://wordpress.com" is used in multiple locations; maybe not in this PR, but I'm wondering if it makes sense to consolidate the constant in just one place?

Gio2018 (commented):
Hey @diegoreymendez , this looks generally good, just had a couple of comments.

WordPressAuthenticator/Credentials/WordPressComCredentials.swift
diegoreymendez: While I agree with the observation, it was my feeling that trying to address this as part of these PRs is a bit beyond scope.  I think consolidating these constants requires some analysis of its own that would risk causing scope creep in these PRs.

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
diegoreymendez: I agree this isn't ideal, but it was the best solution in terms of simplicity and scope-tightness that I could find.

In my modest opinion the reason I'm having to do this is because the code isn't well architected.  I could try to inject the code to "continue" but the problem is that this method exists at the boundary between WPiOS and WPAuth... meaning that the logic for spawning the VCs is defined now within WPAuth, and if I made that code injectable, it would need to be moved to WPiOS.

I don't know if that makes sense, but basically injecting logic would mean moving that logic to all host Apps (WCiOS, WPiOS, Woo, etc).

As in the reply below though - I don't particularly disagree... the reason I chose the boolean instead is to cause the smallest possible impact as part of a bugfix (since these PRs are not really enhancements).

Gio2018: I see. If the price is to have an impact on all the apps, then we should probably leave as it is for now.

WordPressAuthenticator/Credentials/WordPressComCredentials.swift
Gio2018: Totally. I noticed this here but it needs not to be addressed in this PR.

Gio2018 (approved):
:shipit:

diegoreymendez:
Thanks Gio!

PR #558, opened 2021-01-08
Bump podspec to 1.33.0-beta.5

ctarda:
We pushed a [commit](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/f3ec73993dd73757f6b923e1b087bd085bc24eaa) to develop by accident (oops) so it's time to bump the podspec.

The change pushed to develop directly is innocuous so I don't think it's necessary to revert it and go through a formal PR review. We can consider this PR as the 👍  associated to that diff. 😊 

## Changes
* Bumped the podspec version to 1.33.0-beta.4

## How to test
* Check that CI is ✅

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.33.0-beta.3"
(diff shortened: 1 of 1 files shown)


ctarda:
The commit in question was incorporated into PR #553, which also bumped the podspec, so this is not necessary anymore. Closing

PR #560, opened 2021-01-11
Release 1.33.0 -> trunk

AliSoftware:
In preparation for Woo and WP code freezes today

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -85,6 +85,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-darwin-19
   x86_64-darwin-20
 
 DEPENDENCIES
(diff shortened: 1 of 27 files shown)


loremattei (approved):
This looks good to me. 

Small note: I see we have a beta version of WordPressKit in `Podfile.lock`, which is not great. We can live with this since we'll release WordPressKit 4.24.0 without any code change, but in the future it would be great if we manage to release the deps in the correct order.

AliSoftware:
Good point. I missed this one since it wasn't listed by `fastlane` (and I grep'd for `beta` versions in Woo's `Podfile` … but I should have looked at the `Podfile.lock` instead to see transitive dependencies too). I'll add this to my list of improvements to do on the lane.

AliSoftware:
I've opened https://github.com/wordpress-mobile/WordPressKit-iOS/pull/324 ; I'll `pod update` on this PR once WPKit is released and live on CDN before merging this PR.

PR #561, opened 2021-01-11
DRY two factor code handling in pasteboard, fix leading zero loss

### LINKED ISSUE #557: Google Authenticator 2FA codes with leading zeros are copied without the leading zero
This causes the continue button to not be enabled as a result:

step1

step2

I think it is because of this Int cast: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/108a03cf8b2b1ae0cca34538bb5c1abdebc6845a/WordPressAuthenticator/Unified%20Auth/View%20Related/2FA/TwoFAViewController.swift#L324

allendav:
Fixes #557 

To test, e.g. with WooCommerce iOS
- Build
- Ensure all unit tests pass, including the two new tests added for this under PasteboardTests
- Update WooCommerce iOS' Podfile to point at a local clone of this branch
- Build WooCommerce iOS with this pod
- Launch
- Sign in to WordPress.com using an account with 2FA enabled
- At the 2FA code prompt, swipe back to the home screen and open your authenticator app
- Select a 2FA code with a leading zero and tap it to copy it
- Swipe back into the WooCommerce iOS app
- Ensure the 2FA code (all six digits) is automatically pasted for you
- Don't actually complete login. Instead, return to the authenticator app and grab a 2FA code without a leading zero.
- Swipe back into the WooCommerce iOS app
- Ensure the 2FA code (all six digits) is automatically pasted for you
- Lastly, do it again but this time grab the code for your actual login and ensure you can complete login with it.

### DIFF
--- a/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
+++ b/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
@@ -25,4 +25,22 @@ extension UIPasteboard {
             }
         }
     }
+
+    /// Detects whether UIPasteboard.general contains a 2FA code.
+    /// - Parameters:
+    ///   - completion: Called with the 2FA code on success, or failure otherwise
+    @available(iOS 14.0, *)
+    func detectTwoFactorCode(completion: (Result<String, Error>) -> Void) {
+        UIPasteboard.general.detect(patterns: [.number]) { result in
+            switch result {
+                case .success(let detections):
+                    if let twoFactorCode = detections.first?.value as? String {
+                        completion(.success(twoFactorCode))
+                        return
+                    }
+            }
+        }
+
+        completion(.failure());
+    }
 }
--- a/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
+++ b/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
@@ -26,20 +26,48 @@ extension UIPasteboard {
         }
     }
 
-    /// Detects whether UIPasteboard.general contains a 2FA code.
+    /// Attempts to detect and return a authenticator code from the pasteboard.
+    /// Expects to run on main thread.
     /// - Parameters:
-    ///   - completion: Called with the 2FA code on success, or failure otherwise
+    ///   - completion: Called with a six digit authentication code on success
     @available(iOS 14.0, *)
-    func detectTwoFactorCode(completion: @escaping (Result<String, Error>) -> Void) {
+    func detectAuthenticatorCode(completion: @escaping (Result<String, Error>) -> Void) {
         UIPasteboard.general.detect(patterns: [.number]) { result in
             switch result {
                 case .success(let detections):
-                    if let twoFactorCode = detections.first?.value as? String {
-                        completion(.success(twoFactorCode))
+                    guard let firstMatch = detections.first else {
+                        completion(.success(""))
                         return
                     }
+                    guard let matchedNumber = firstMatch.value as? NSNumber else {
+                        completion(.success(""))
+                        return
+                    }
+
+                    var authenticationCode = matchedNumber.stringValue
+
+                    /// Reject numbers with decimal points or signs in them
+                    let codeCharacterSet = CharacterSet(charactersIn: authenticationCode)
+                    if !codeCharacterSet.isSubset(of: CharacterSet.decimalDigits) {
+                        completion(.success(""))
+                        return
+                    }
+
+                    /// We need 6 digits. No more, no less.
+                    if authenticationCode.count > 6 {
+                        completion(.success(""))
+                        return
+                    }
+
+                    while authenticationCode.count < 6 {
+                        authenticationCode = "0" + authenticationCode
+                    }
+
+                    completion(.success(authenticationCode))
+                    return
                 case .failure(let error):
                     completion(.failure(error))
+                    return
             }
         }
     }
--- a/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
+++ b/WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
@@ -30,7 +30,7 @@ extension UIPasteboard {
     /// - Parameters:
     ///   - completion: Called with the 2FA code on success, or failure otherwise
     @available(iOS 14.0, *)
-    func detectTwoFactorCode(completion: (Result<String, Error>) -> Void) {
+    func detectTwoFactorCode(completion: @escaping (Result<String, Error>) -> Void) {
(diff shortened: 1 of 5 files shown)


allendav:
Interesting. It isn't the cast after all. It is how .number pattern matching is working with the pasteboard, i.e.:

```
                pasteboard.detectValues(for: [.number]) { valuesResult in
                    switch valuesResult {
                        case .success(let twofactorcode):
                            let foo = twofactorcode.first
                            foo?.value
```

returns 11234 when the pasteboard contains 011234

allendav:
LOL. I've been digging into this more today. Finally have a robust sequence for pulling out two factor codes. Will update the PR with it tomorrow:

```
import UIKit

let pasteboard = UIPasteboard.general

pasteboard.string = "123123" /// Try different values here like 123456 123 123123123 asvab 3.1415 -99 etc

/// Detects a two factor authorization code in the pasteboard (if present)
/// and calls the completion with it, prepended as needed with zeros to yield six digits
func detectTwoFactorCode(completion: @escaping (Result<String, Error>) -> Void) {
    pasteboard.detectPatterns(for: [.number]) { result in
        var twoFactorCode: String = ""

        switch result {
            case .success(let detections):
                if detections.isEmpty {
                    completion(.success(""))
                    return
                }
                
                pasteboard.detectValues(for: [.number]) { valuesResult in
                    switch valuesResult {
                        case .success(let match):
                            if let matchedNumber = match.first?.value as? NSNumber {
                                twoFactorCode = matchedNumber.stringValue
                                
                                /// Reject numbers with decimal points or signs in them
                                let possibleCodeSet = CharacterSet(charactersIn: twoFactorCode)
                                if !possibleCodeSet.isSubset(of: CharacterSet.decimalDigits) {
                                    completion(.success(""))
                                    return
                                }
                                
                                /// Reject numbers with more than 6 digits
                                if twoFactorCode.count > 6 {
                                    completion(.success(""))
                                    return
                                }
                                
                                /// Prepend zeros to get to 6 digits
                                while twoFactorCode.count < 6 {
                                    twoFactorCode = "0" + twoFactorCode
                                }

                                completion(.success(twoFactorCode))
                                return
                            }
                        case .failure(let error):
                            completion(.failure(error))
                            return
                    }
                }
            case .failure(let error):
                completion(.failure(error))
                return
        }
    }
}

detectTwoFactorCode() { result in
    switch result {
    case .success(let twoFactorCode):
        twoFactorCode
    case .failure(_):
        "Failure"
    }
}
```

WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
```diff
@@ -25,4 +25,50 @@ extension UIPasteboard {
...
+
+                    /// We need 6 digits. No more, no less.
+                    if authenticationCode.count > 6 {
+                        completion(.success(""))
+                        return
+                    }
+
+                    while authenticationCode.count < 6 {
+                        authenticationCode = "0" + authenticationCode
+                    }
```
bjtitus: This may be my crazy fears of avoiding `while` loops and mutable state, but would it make more sense to just prepend the necessary zeros here and avoid the loop?

```suggestion
                    let missingDigits = 6 - authenticationCode.count
                    let paddingZeros = String(repeating: "0", count: missing)
                    let paddedAuthenticationCode = paddingZeros + authenticationCode
```

This would also remove the need to make `authenticationCode` mutable, which I'm always in favor of.

WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
```diff
@@ -25,4 +25,50 @@ extension UIPasteboard {
...
+
+                    /// Reject numbers with decimal points or signs in them
+                    let codeCharacterSet = CharacterSet(charactersIn: authenticationCode)
+                    if !codeCharacterSet.isSubset(of: CharacterSet.decimalDigits) {
+                        completion(.success(""))
+                        return
+                    }
+
+                    /// We need 6 digits. No more, no less.
+                    if authenticationCode.count > 6 {
```
bjtitus: Minor thing, but I wonder if the length should be a parameter of this method. If not, it should probably go as a constant at the top. I think I prefer making it a parameter with a default value, though.

```swift
public func detectAuthenticatorCode(length: Int = 6, completion: @escaping (Result<String, Error>) -> Void)
```

WordPressAuthenticatorTests/Authenticator/PasteboardTests.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  PasteboardTests.swift
+//  WordPressAuthenticatorTests
+//
+//  Created by Allen Snook on 1/13/21.
+//  Copyright © 2021 Automattic. All rights reserved.
+//
```
bjtitus: Can you remove this generated header?

We generally keep them out of the projects, but there's no good linter rule available to do it. 😞

bjtitus (changes requested):
Just a few minor code suggestions. Otherwise, everything looks good!

bjtitus:
It's pretty annoying that Apple is stripping the leading zeros and doesn't mention it.
We should file feedback to have them at least add it to the docs or, better yet, don't strip the leading digits in case they're significant...

WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
allendav: Fixed in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/561/commits/e109be8783aa27c3b3bba97ee45ad559aaee4447

WordPressAuthenticator/Extensions/UIPasteboard+Detect.swift
allendav: Fixed in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/561/commits/e109be8783aa27c3b3bba97ee45ad559aaee4447

WordPressAuthenticatorTests/Authenticator/PasteboardTests.swift
allendav: Fixed in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/561/commits/c96f997731e9859ef5b96e0a8e5561a4683869ae

allendav:
> We should file feedback...

Agreed. Done. Feedback ID FB8974839

allendav:
All feedback addressed. Ready for next round of review. Thank you @bjtitus !

bjtitus (approved):
👍 Looks good to me!

Thanks for fixing this @allendav!

PR #564, opened 2021-01-21
Separate login UI creation to its own function so that the host app can determine how it is displayed

jaclync:
For https://github.com/woocommerce/woocommerce-ios/issues/3107 and https://github.com/woocommerce/woocommerce-ios/issues/3429

## Why

When launching the app in logged out state, we call `WordPressAuthenticator.showLogin` to show the login UI in both WPiOS and WCiOS apps. However, since iOS 14 we started seeing a delay presenting the login UI ([similar WPiOS issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15399)). From my testing, it seems to be due to loading the view controller from **Login.storyboard** in the **authenticator pod** which is not easy to change. Because the app window's root view controller is the tab bar UI, the uninitialized tab bar UI is shown briefly before the login UI is presented.

In order to solve the glitch in WCiOS, we are setting the login view controller to the app window's root view controller instead of calling `WordPressAuthenticator.showLogin` to present the login UI from a source view controller. Therefore, this PR moved the login UI creation to a separate public function so that the host app can decide how it is displayed (set to the app window's root view controller in WCiOS' case).

## Changes

- Added a function `loginUI` with the same logic from how a login view controller is created and configured from the pre-existing `showLogin`.

### WPiOS

Please review/test https://github.com/wordpress-mobile/WordPress-iOS/pull/15682 - I think we can merge it once it's approved so that we can also QA the authenticator change in case of regression

### WCiOS

Please see the testing section in https://github.com/woocommerce/woocommerce-ios/pull/3498

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -174,7 +174,7 @@ import WordPressKit
     public class func loginUI(showCancel: Bool = false, restrictToWPCom: Bool = false, onLoginButtonTapped: (() -> Void)? = nil) -> UIViewController? {
         let storyboard = Storyboard.login.instance
         guard let controller = storyboard.instantiateInitialViewController() else {
-            assertionFailure("Cannot instantiate login root controller from Storyboard")
+            assertionFailure("Cannot instantiate initial login controller from Login.storyboard")
             return nil
         }
         if let childController = controller.children.first as? LoginPrologueViewController {
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -156,20 +156,34 @@ import WordPressKit
     ///   - onLoginButtonTapped: Called when the login button on the prologue screen is tapped.
     ///   - onCompletion: Called when the login UI presentation completes.
     public class func showLogin(from presenter: UIViewController, animated: Bool, showCancel: Bool = false, restrictToWPCom: Bool = false, onLoginButtonTapped: (() -> Void)? = nil, onCompletion: (() -> Void)? = nil) {
-        defer {
-            trackOpenedLogin()
+        guard let loginViewController = loginUI(showCancel: showCancel, restrictToWPCom: restrictToWPCom, onLoginButtonTapped: onLoginButtonTapped) else {
+            return
         }
+        presenter.present(loginViewController, animated: animated, completion: onCompletion)
+        trackOpenedLogin()
+    }
 
+    /// Returns the view controller for the login flow.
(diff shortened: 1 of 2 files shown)


diegoreymendez:
This is something we've been discussing for WPiOS with @frosty as well.  Pinging him to loop him in.

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -156,20 +156,34 @@ import WordPressKit
     ///   - onLoginButtonTapped: Called when the login button on the prologue screen is tapped.
     ///   - onCompletion: Called when the login UI presentation completes.
     public class func showLogin(from presenter: UIViewController, animated: Bool, showCancel: Bool = false, restrictToWPCom: Bool = false, onLoginButtonTapped: (() -> Void)? = nil, onCompletion: (() -> Void)? = nil) {
-        defer {
-            trackOpenedLogin()
+        guard let loginViewController = loginUI(showCancel: showCancel, restrictToWPCom: restrictToWPCom, onLoginButtonTapped: onLoginButtonTapped) else {
```
ctarda, line 159: So if I understand correctly, this is the method that WordPress would call to initialize the login flow, while WooCommerce would call `loginUI`?

If that's the case, would it make sense to make the names of both methods a bit more explicit, or rename in a way that the method name indicates that one method presents the login flow while the other leaves the responsibility of presenting the flow to the client app?

I'm curious about what others think though

diegoreymendez: @ctarda - My personal opinion is that we'll probably want to follow suit in WPiOS soon (in showing the login UI as the root VC).  There are several advantages in doing so.

With this in mind I'd say the naming isn't overly important as we'll eventually deprecate `showLogin(...)` entirely - the documentation of these methods should be enough for starters.

diegoreymendez: Correction: we may never deprecate `showLogin(...)` as it is currently used in WPiOS to log in to additional sites (which means it wouldn't be shown as the root VC).

Still the naming `showLogin(...)` and `loginUI(...)` sounds good to me.  The other only option I can think of is to use `loginViewController(...)` instead of `loginUI(...)` but I'm not sure that'd make a huge difference.

ctarda: Sounds good!

ctarda (approved):
Thanks for your feedback @diegoreymendez If you don't have any concerns about the naming, I'd say :shipit:

jaclync:
Thanks for the feedback César and Diego! I tested in WPiOS with this branch https://github.com/wordpress-mobile/WordPress-iOS/pull/15682 and verified the login behavior is as before. Merging the PR now for code freeze next Monday!

PR #571, opened 2021-01-28
Fix misconfiguration in SpacerTableViewCell.xib

### LINKED ISSUE #570: Fix SpacerTableViewCell.xib
There are several issues with this xib file:
* It is missing the reusable identifier
* It is set to inherit the module from target
* it is missing the userlabel

All these issues together are making that, for host apps using this UI element (i.e. WooCommerce) there will be a warning logged in the Console:

```
[Storyboard] Unknown class _TtC31WordPressAuthenticatorResources19SpacerTableViewCell in Interface Builder file.
Error: Unidentified tableViewCell type found.
```

We think this might be causing a rare crash in WooCommerce: woocommerce/woocommerce-ios#3534

ctarda:
Closes #570 

## Changes

There are several issues with SpacerTableViewCellxib file:

* It is missing the reusable identifier
* It is set to inherit the module from target
* it is missing the userlabel

All these issues together are making that, for host apps using this UI element (i.e. WooCommerce) there will be a warning logged in the Console:

```
[Storyboard] Unknown class _TtC31WordPressAuthenticatorResources19SpacerTableViewCell in Interface Builder file.
Error: Unidentified tableViewCell type found.
```

We think this might be causing a rare crash in WooCommerce: woocommerce/woocommerce-ios#3534

## How to test
* The changes in this PR do not affect WordPress-iOS, as the xib modified is not used.
* The best way to test is testing the following PR in WooCommerce: woocommerce/woocommerce-ios#3543


ScoutHarris (approved):

ctarda:
Thanks for the review @ScoutHarris 🙇

Issue #501, opened 2020-10-12
Accessibility: Rebuild the NUXButton


mindgraffiti:
Pair with: https://github.com/wordpress-mobile/WordPress-iOS/issues/15084 because WPiOS uses NUXButton directly in the main project.
Ref. p5T066-1yK-p2#comment-6042

### Background

`NUXButton` looks like a `FancyButton`, but does not act like one. NUXButtons differ because they have a built-in spinner (`UIActivityIndicatorView`) that activates after a user taps the button. (iirc, it handles form validation in some cases as well.)

NUXButtons do not scale well when Dynamic Text is enabled, as seen in the prologue screen and the "Continue with _ " buttons on the Get Started screen. Build a new button that uses dynamic font sizing. [Suggested reading](https://blog.kiprosh.com/dynamic-font-size-in-ios/).

### Screenshots

| NUXButtons | in Authenticator |
| --- | --- |
| ![img_1584-1](https://user-images.githubusercontent.com/1062444/95785683-d05aae00-0c9b-11eb-8aa0-967f8464b0c2.png) | ![img_1585](https://user-images.githubusercontent.com/1062444/95785685-d0f34480-0c9b-11eb-9b70-2b479b11980d.png) |

mokagio:
As of [2025/02/06](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/cce14d7ea5e5af5e52cdcfa75cbec672c1603991) this repo is a public archive and no longer under development.

If you think this issue is still relevant for one or both of the former clients, please open an issue in [WordPress/Jetpack iOS](https://github.com/wordpress-mobile/WordPress-iOS/issues) and [WooCommerce iOS](https://github.com/woocommerce/woocommerce-ios/issues)

Thanks!

PR #572, opened 2021-02-02
Update login / signup buttons copy and font

yaelirub:
WPiOS issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/15703
This PR updates the copy on the login / signup button from "Welcome to WordPress.com" to "Log in or sign up with WordPress.com"

Since the title is more lengthy, this PR also updates the title label font of both buttons from `title3` to `body` per design

Simulator Screen Shot - iPhone 12 Pro - 2021-02-01 at 19 44 51

To Test:
1. Update the WordPressAuthenticator path in WPiOS to this branch
2. Be signed out 
3. See copy as described 
4. See font is body as described 
5. Continue to next screen and see button title fonts are unchanged (title3)

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
@@ -166,7 +166,7 @@ public extension WordPressAuthenticatorDisplayStrings {
                                                    comment: "The button title text for opening the user's preferred email app."),
             createAccountButtonTitle: NSLocalizedString("Create Account",
                                                         comment: "The button title text for creating a new account."),
-            continueWithWPButtonTitle: NSLocalizedString("Continue with WordPress.com",
+            continueWithWPButtonTitle: NSLocalizedString("Log in or sign up with WordPress.com",
                                                comment: "Button title. Takes the user to the login by email flow."),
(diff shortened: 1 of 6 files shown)


leandroalonso (approved):

mattmiklic (approved):
This looks fine in English at the default font size, so I'm going ahead with my 👍 here -- but it would be good to get a build so we can look at longer translations, and non-default system font sizes, to make sure the buttons adapt gracefully.

mattmiklic:
I tested the build with dynamic font sizes enabled and in some other languages. I think we are generally fine to launch this as-is. I do have one suggestion for improvement but as it's not a regression, it can be a separate issue.

At the current font sizes, some of the buttons are split into two lines. [German is a good example](https://user-images.githubusercontent.com/1396552/106797719-fd581580-662a-11eb-8776-4b8a91e155f9.PNG). The same is true [when dynamic font sizes are enabled](https://user-images.githubusercontent.com/1396552/106797753-0812aa80-662b-11eb-8e80-80f994de86a4.PNG). When the text wraps to two lines, we'd benefit from having some top and bottom padding in the button element to keep the text from getting uncomfortably close to the edge of the button shape.

ScoutHarris:
FWIW, there is an open issue to redo that button because of wrapping issues. https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/501.

PR #574, opened 2021-02-04
Shorter sign up - remove Send Link by email screen

yaelirub:
After talking this over with design (@mattmiklic ) we decided to remove this screen:
Simulator Screen Shot - iPhone 12 Pro - 2021-02-02 at 13 00 39

Instead, we will be sending the email immediately after the user entered their email : 

https://user-images.githubusercontent.com/1335657/106846364-96098800-6661-11eb-812e-c80a3ae1cdc2.MP4

To test:
1. copy this branch to pod file in WPiOS 
2.  pod install
3. Test login and signup work as expected with no issues
4. Check tracking is as expected

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.35.0"
+  s.version       = "1.35.1"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
   s.description   = <<-DESC
--- a/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
@@ -401,7 +401,7 @@ private extension GetStartedViewController {
         let errorCode = userInfo[WordPressComRestApi.ErrorKeyErrorCode] as? String
 
         if errorCode == "unknown_user" {
-            self.showSignupViewIfNecessary(error)
+            self.sendEmail()
         } else if errorCode == "email_login_not_allowed" {
                 // If we get this error, we know we have a WordPress.com user but their
                 // email address is flagged as suspicious.  They need to login via their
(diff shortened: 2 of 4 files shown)


yaelirub:
@mattmiklic , do you think we should update the copy on the first screen to let them know it has to be on the device?

mattmiklic:
I don't think we need to add any messaging to the Get Started screen, since at that point they could intend to log in via password and the email note wouldn't be relevant. The email confirmation is clear enough that it should be checked on the device, I think.

I do think it'd be good to add a message to the email confirmation screen letting them know they can go back if they didn't intend to sign up. Something like "Didn't mean to create a new account? Go back to re-enter your email address."

yaelirub:
> "Didn't mean to create a new account? Go back to re-enter your email address."

Love it! I'll add it. Thank you so much, @mattmiklic !

ScoutHarris:
I assume this isn't done yet since the "Didn't mean to create a new account?" message hasn't been added. So I'll just leave a comment that the podspec version needs to be bumped.

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -411,36 +411,45 @@ private extension GetStartedViewController {
             self.displayError(error as NSError, sourceTag: self.sourceTag)
         }
     }
+    
+    // MARK: - Send email
 
-    /// Offer host apps the opportunity to decide if they want to allow signups
+    /// Makes the call to request a magic signup link be emailed to the user.
     ///
-    func showSignupViewIfNecessary(_ error: Error) {
```
ctarda, line 417: This would completely remove the possibility for WooCommerce to allow signups, which I think is something that we want, but I am not completely sure it.

I will chat with Garance.

ctarda (approved):
I did a quick test with WooCommerce, and even though this breaks the API, it does not remove any functionality that we might be using at the moment.

That being said, I would wait to merge, if possible, until @Garance91540  gives the 👍

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
yaelirub: >This would completely remove the possibility for WooCommerce to allow signups

Why? This function is just showing that view that I removed if Im not mistaken. The signup process remains the same minus a screen.

yaelirub:
Added copy:
IMG_7863C43CB5A6-1
@ScoutHarris , @mattmiklic , ready for another look

mattmiklic:
Looking good to me. 👍

WordPressAuthenticator.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.35.0-beta.1"
+  s.version       = "1.36.0-beta.1"
```
ScoutHarris: Shouldn't this be `1.35.0-beta.2`?

ScoutHarris (approved):
LGTM. After the version is corrected, :shipit: .

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
ctarda: Ouch, sorry! I misread the code. I, for some reason, thought that the delegate method to ask host apps if they want to support account signups had been removed too. Go figure.

emilylaguna (approved):
Tested this on:
- iPhone Simulator - iOS 13
- iPhone Simulator - iOS 14
- iPad Simulator - iOS 14
- Real Device: iPhone 11 Pro 

Had no issues other than the [unrelated login issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15822)

LGTM 🚀 !

jkmassel (approved):
Version change LGTM – I presume all the other files have no changed.

:shipit:

PR #578, opened 2021-02-24
Stop tracking prologue step based on view presentation

frosty:
This PR fixes an issue introduced when WordPress for iOS switched to displaying the authenticator as the window's root view controller instead of modally. That change meant that the previous check for `isBeingPresentedInAnyWay` always returned false, so we didn't track the prologue step.

This PR reverts to the previous method of tracking whether this is the first time the view controller has appeared, using a simple property.

Please test using the associated WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15967

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.35.1"
+  s.version       = "1.35.2"
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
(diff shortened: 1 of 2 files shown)


emilylaguna (approved):
This is working for me 🚀. Started logged in:

1. Logged out
2. Got: `🔵 Tracked: unified_login_step <flow: prologue, source: default, step: prologue>`
3. Restarted the app
4. Got `🔵 Tracked: unified_login_step <flow: prologue, source: default, step: prologue>`
5. Went through the login flow did not see that log again

I also tested the `GoogleSignupConfirmationViewController` which uses `isBeingPresentedInAnyWay` and verified it is still working correctly.

PR #580, opened 2021-03-12
Push the podspec to our internal spec repo in addition to trunk

AliSoftware:
My HACK week project was about mirroring our internal pods to our new internal spec repo (to allow us to avoid waiting for the CDN sync).

This PR updates this pod's CircleCI config so that it we now push the podspec to the internal spec repo – in addition to still pushing it to trunk.

⚠️  This will first require us to merge wordpress-mobile/circleci-orbs#69 and make a new release of our orb before we can merge this. Please don't merge this PR before wordpress-mobile/circleci-orbs#69, but apart from this, this PR is ready for review.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   # Using 1.0 of our Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
-  ios: wordpress-mobile/ios@dev:pod-repo-push
+  ios: wordpress-mobile/ios@1.0
 
 # YAML anchors for some common/repeated values
 x-common-params:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,12 +4,22 @@ orbs:
   # Using 1.0 of our Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
   ios: wordpress-mobile/ios@dev:pod-repo-push
 
+# YAML anchors for some common/repeated values
+x-common-params:
+  - &xcode-version "12.2.0"
+  - &podspec "WordPressAuthenticator.podspec"
+  - &on-tags-only
+      tags:
+        only: /.*/
+      branches:
+        ignore: /.*/
+
 workflows:
   test_and_validate:
     jobs:
       - ios/test:
           name: Test
-          xcode-version: "12.2.0"
+          xcode-version: *xcode-version
           workspace: WordPressAuthenticator.xcworkspace
           scheme: WordPressAuthenticator
           device: iPhone 11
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2,14 +2,24 @@ version: 2.1
 
 orbs:
   # Using 1.0 of our Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
-  ios: wordpress-mobile/ios@1.0 
+  ios: wordpress-mobile/ios@dev:pod-repo-push
```
AliSoftware: ⚠️  We need to revert this to `@1.0` once https://github.com/wordpress-mobile/circleci-orbs/pull/69 is merged and we made a new `1.0.x` tag to publish the orb.

jkmassel: Nice.

.circleci/config.yml
```diff
@@ -18,17 +28,21 @@ workflows:
           pod-install: true
       - ios/validate-podspec:
           name: Validate Podspec
-          xcode-version: "12.2.0"
-          podspec-path: WordPressAuthenticator.podspec
+          xcode-version: *xcode-version
+          podspec-path: *podspec
+          bundle-install: true
```
jkmassel, line 33: This felt weird (vs using like `ios/install-dependencies`) but given that it's part of the published orb already, I don't see a lot of benefit to changing it now 🤔

jkmassel (approved):

.circleci/config.yml
mokagio: What was your experience working with orbs from a branch? 

Whenever I tried that, I always experienced a lag between pushing changes to the orb branch and CircleCI applying it when fetching the orb from that branch.

.circleci/config.yml
```diff
@@ -2,14 +2,24 @@ version: 2.1
 
 orbs:
   # Using 1.0 of our Orbs means it will use the latest 1.0.x version from https://github.com/wordpress-mobile/circleci-orbs
-  ios: wordpress-mobile/ios@1.0 
+  ios: wordpress-mobile/ios@dev:pod-repo-push
+
+# YAML anchors for some common/repeated values
+x-common-params:
```
mokagio, line 8: Just curious, why did you choose the `x-` prefix?

.circleci/config.yml
AliSoftware: Ah, I'm glad I'm not alone in this 😅 . Yep, experienced some lag too:

 - Tbh at first I didn't realize that even for `@dev:branch` references we still needed the orb to be published to the CircleCI registry. So I first expected the orb to be immediately available right after a push, while in practice you need to wait for the orb repo's own CircleCI workflow to finish (and to complete the "publish to orb registry" job) before you can hope use it (so just like for an new official release of the orb).
 - Then even once I realized this, I find that there was still some lag/delay between the orb supposedly being published by CircleCI and it being available for the other projects/workflows. I attributed this to both some propagation delay, and to CircleCI's status being yellow that day (apparently they were on maintenance while I was working on this).
 - I also didn't always monitor the orb's own CircleCI workflow page, so at some point in my development process I didn't realize I had a syntax error in my orb code (I used `<<` in my bash script without escaping it), so I kept pushing changes to the orb, but not seeing any changes when using that orb's `@dev:pod-repo-push` branch from WPAuth's own CircleCI job, and was quite puzzled until I realized that the publication of the orb to the registry failed, which is why my changes seemed ignored and were never released

All that is indeed a bit tedious to work with and can make you lose quite some time when you need to work in parallel on the orb and the repo testing it…

.circleci/config.yml
AliSoftware: Yeah I didn't want to touch other parts of the orb, especially since changing things here would be breaking change that would require a bump of the orb to 2.x, which didn't feel necessary – in practice we try to avoid breaking changes in the orb and try our best to make retro-compatible changes unless absolutely necessary.

.circleci/config.yml
AliSoftware: I'm not sure it's an official convention in CircleCI itself, but starting things with `x-` is a common practice in other standards when you want to use keys that are not official / not standard keys that are part of the API / schema.

For example, HTML uses this convention when you want to declare custom headers that are not part of the HTML standard and official headers, but are instead private/custom keys that only your server might know how to interpret.

I felt like using `x-` prefix here would (1) help indicate that this key is not an official CircleCI key but just an arbitrary custom key and (2) ensure that this would never conflict with a new key that CircleCI might introduce in a later version of their config schema one day.

.circleci/config.yml
AliSoftware:
```suggestion
  ios: wordpress-mobile/ios@1.0
```

.circleci/config.yml
mokagio: Thanks for the extra info. Sounds good to me 👍

PR #582, opened 2021-03-22
[Tooling] Standardize podspec format

AliSoftware:
As part of paaHJt-Vl-p2, this simply standardize the podspec file so that all our pods use a similar format and structure, with similar keys and common values.

## Format / Rules

This are the rules/format we'll try to use for all our podspecs (PRs pending for others)

* Update the values. This is debatable
  * Set the `homepage` ~to `http://apps.wordpress.com`~ to the GitHub repo's url
  * Use `:type` and `:file` keys for license to also provide a link to the full license text. Add `LICENSE` file if it doesn't yet exists.
  * Use `Automattic` as the sole author's name for consistency with all the pods
  * Set the `social_media_url` to the ~WordPressiOS~ Automattic twitter account
  * Remove the `requires_arc = true` key since this is the default since a long while now and is obsolete.
* Group the keys like the following, each section separated by a blank line for readability:
  1. Start with `name` and `version` of the pod at the very top
  2. Next is `summary` and `description`
  3. Next are metadata about the pod: `homepage`, `license`, `author` and `social_media_url`
  4. Then are compatibility version restrictions info: `platform` and `swift_version`
  5. Then are all the keys about the pod's source and how to compile it (`source`, `source_files`, and any key related to instructions for CocoaPods on how to build and assemble the pod)
  6. Finally are the list of `dependency` entries.

## How to Review

I will use this `WordPressAuthenticator` PR to collect feedback on the format standardization, but I will open draft PRs on a couple of our other pods to show how the application of those consistency/standard affect them, so feel free to look at those drafts too, to compare the effect those rules will have on them.

* https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/291
* https://github.com/wordpress-mobile/WordPressKit-iOS/pull/376
* https://github.com/wordpress-mobile/WordPressUI-iOS/pull/84
* https://github.com/wordpress-mobile/wpxmlrpc/pull/56

_Once this PR is approved and we agree on the common values and structure, I'll apply any feedback to the other PRs and un-draft them all._

## How to test

* Mainly check the PR diff for the new copies and content of metadata fields.
* To ensure there are no syntax errors after those changes, you can run `pod lib lint *.podspec` from the repo's root.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
                     Plus: WordPress.com *signup* is supported.
                     DESC
 
-  s.homepage      = "http://apps.wordpress.com"
+  s.homepage      = "https://apps.wordpress.com"
   s.license       = { :type => "GPLv2", :file => "LICENSE" }
   s.author        = { "Automattic" => "mobile@automattic.com" }
   s.social_media_url = "https://twitter.com/automattic"
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
 
   s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
   s.license       = { :type => "GPLv2", :file => "LICENSE" }
-  s.author        = { "The WordPress Mobile Team" => "mobile@automattic.com" }
+  s.author        = { "The WordPress Mobile Team" => "mobile@wordpress.org" }
 
   s.platform      = :ios, "11.0"
   s.swift_version = '4.2'
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -12,8 +12,7 @@ Pod::Spec.new do |s|
 
   s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
   s.license       = { :type => "GPLv2", :file => "LICENSE" }
-  s.author        = { "Automattic" => "mobile@automattic.com" }
-  s.social_media_url = "https://twitter.com/automattic"
+  s.author        = { "The WordPress Mobile Team" => "mobile@automattic.com" }
 
   s.platform      = :ios, "11.0"
   s.swift_version = '4.2'
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,20 +1,23 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
   s.version       = "1.36.0-beta.2"
-  s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
 
+  s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
   s.description   = <<-DESC
                     This framework encapsulates everything required to display the Authentication UI
                     and perform authentication against WordPress.com and WordPress.org sites.
 
-                    Plus: WordPress.com *signup*  is supported.
+                    Plus: WordPress.com *signup* is supported.
                     DESC
 
-  s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
-  s.license       = "GPLv2"
-  s.author        = { "WordPress" => "mobile@automattic.com" }
+  s.homepage      = "http://apps.wordpress.com"
+  s.license       = { :type => "GPLv2", :file => "LICENSE" }
+  s.author        = { "Automattic" => "mobile@automattic.com" }
+  s.social_media_url = "http://twitter.com/WordPressiOS"
+
   s.platform      = :ios, "11.0"
   s.swift_version = '4.2'
+
   s.source        = { :git => "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git", :tag => s.version.to_s }
   s.source_files  = 'WordPressAuthenticator/**/*.{h,m,swift}'
   s.private_header_files = "WordPressAuthenticator/Private/*.h"
@@ -26,7 +29,6 @@ Pod::Spec.new do |s|
       'WordPressAuthenticator/**/*.{storyboard,xib}'
     ]
   }
-  s.requires_arc  = true
   s.static_framework = true # This is needed because GoogleSignIn vendors a static framework
   s.header_dir    = 'WordPressAuthenticator'
 
@@ -36,7 +38,6 @@ Pod::Spec.new do |s|
   s.dependency 'lottie-ios', '~> 3.1.6'
   s.dependency 'NSURL+IDN', '0.4'
   s.dependency 'SVProgressHUD', '~> 2.2.5'
-
   s.dependency 'Gridicons', '~> 1.0'
   s.dependency 'GoogleSignIn', '~> 5.0.2'
 
@@ -49,5 +50,4 @@ Pod::Spec.new do |s|
   # Fixing arm64 issue with Xcode 12: https://github.com/CocoaPods/CocoaPods/issues/10104
   s.user_target_xcconfig = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
   s.pod_target_xcconfig = { 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'arm64' }
-
 end
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
                     Plus: WordPress.com *signup* is supported.
                     DESC
 
-  s.homepage      = "https://apps.wordpress.com"
+  s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
   s.license       = { :type => "GPLv2", :file => "LICENSE" }
   s.author        = { "Automattic" => "mobile@automattic.com" }
   s.social_media_url = "https://twitter.com/automattic"
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.homepage      = "http://apps.wordpress.com"
   s.license       = { :type => "GPLv2", :file => "LICENSE" }
   s.author        = { "Automattic" => "mobile@automattic.com" }
-  s.social_media_url = "http://twitter.com/WordPressiOS"
+  s.social_media_url = "https://twitter.com/automattic"
 
   s.platform      = :ios, "11.0"
   s.swift_version = '4.2'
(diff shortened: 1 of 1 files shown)


WordPressAuthenticator.podspec
```diff
@@ -1,20 +1,23 @@
...
                     and perform authentication against WordPress.com and WordPress.org sites.
 
-                    Plus: WordPress.com *signup*  is supported.
+                    Plus: WordPress.com *signup* is supported.
                     DESC
 
-  s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
-  s.license       = "GPLv2"
-  s.author        = { "WordPress" => "mobile@automattic.com" }
+  s.homepage      = "http://apps.wordpress.com"
```
AliSoftware: This is for consistency between all the pods, and because the url to the GitHub repo is already available as part of other metadata (especially `s.source.git`)

WordPressAuthenticator.podspec
```diff
@@ -1,20 +1,23 @@
...
+  s.license       = { :type => "GPLv2", :file => "LICENSE" }
```
AliSoftware, line 14: * We happen to have a `LICENSE` file in that repo, so better point to it.
* But that's apparently not the case for all our repos, so I wonder what's best for other pods for which I'll open similar PRs: create the `LICENSE` file and point to it, or just not list the file name if the file doesn't exist, without creating it?
* It's easy enough to copy/paste the `LICENSE` file we have here to any pod that also uses GPLv2 but doesn't have a `LICENSE` file yet. But not all our pods use `GPLv2` (some use `MIT` license). Not sure that license discrepency is expected (but that's probably a debate for another day)

WordPressAuthenticator.podspec
```diff
@@ -1,20 +1,23 @@
...
+  s.author        = { "Automattic" => "mobile@automattic.com" }
```
AliSoftware: For consistency with other pods.

* Some were using `Automattic` as the author name, others were using `WordPress`. Since those pods are authored mainly by the Automattic company, and more importantly the email address associated with that author is an a8c address, I figured it made more sense to use `Automattic` rather than `WordPress` here
* Some other pods are listing multiple authors, like `mobile@automattic.com` but also and/or actual contributors (a12s). [[Example in WordPressUI](https://github.com/wordpress-mobile/WordPressUI-iOS/pull/84/files)]. I wonder if we will want to keep them listed too, or if we should just simply use a single `mobile@automattic.com` author for all our pods?

WordPressAuthenticator.podspec
```diff
@@ -1,20 +1,23 @@
...
+  s.social_media_url = "http://twitter.com/WordPressiOS"
```
AliSoftware: Was planning to use this same media URL for all our internal pods. I wonder if we have a more suited Twitter account to link there instead?

WordPressAuthenticator.podspec
mokagio: Is there a reason to use HTTP over HTTPS, maybe a CocoaPods convention? If you open the HTTP URL, it redirects to the HTTPS one.

```suggestion
  s.homepage      = "https://apps.wordpress.com"
```

WordPressAuthenticator.podspec
mokagio: The only possible issue I see with this format is if we chose to change license, by means of updating the `LICENSE` file, and forgot to update the `type` key here.

IMHO, it'd be great to have a `LICENSE` file in every open-source repo. I'm not sure if it's an explicit standard, but I've come to expect it from the sample of open-source repos I've interact with so far.

WordPressAuthenticator.podspec
mokagio: It's nice to have your name listen among the authors of a library, but at the same time it's a bit of a maintenance burden. It also leaves the question open as what to do when someone leaves or stops being involved with the project, should they remain in the list because of their contributions so far, or be removed to avoid being contacted about something they're not working on anymore?

Overall, I think a catch-all setup like the one you're proposing makes things more straightforward.

Some of our projects are narrow-focused on WordPress, but other (and maybe more in the future) are more generic. As such, putting everything under the Automattic umbrella makes sense to me. 

As for the authors, it's a matter of keeping things consistent and lowering the amount of ad hoc decision making when modifying the spec 😄

WordPressAuthenticator.podspec
mokagio: The only other suggestion I have is 

```suggestion
  s.social_media_url = "https://twitter.com/automattic"
```

Less specific than WordPress iOS, but might be more encompassing in the context of a standard for all our libraries, including those that cross or are outside the WordPress boundary.

mokagio (approved):
This format looks good to me 👍 

There's something I'd like to raise here, which touches both on the `homepage` and `social_media_url` points, even though is *out of scope* for this PR.

Both the Twitter handle and https://app.wordpress.com are a good landing page from a "marketing" point of view.

image

For a developer visiting them, though, they don't provide on-topic information. My rationale is that, personally, if I were to follow those links as a developer that just discovered the open-source library, I'd like to know more about the other open-source projects that Automattic made.

I'd love to have something like `opensource.wordpress.com` (even though it's almost a tautology give how much of what we do is open source by default). Another option might be to host that on a `.github.io` static site.

Here's some examples:

- https://stripe.com/au/open-source
- http://shopify.github.io/
- https://twitter.github.io/
- https://netflix.github.io/

As I said, this is clearly out of scope, but I'd like to put it out there and would be happy to formalize this in a P2 proposal if it resonates with other reviewers, too.

In the meantime, I'd say the values you suggest are good. My HTTPS vs HTTP and Automattic vs WordPress suggestions are non-blocking.

WordPressAuthenticator.podspec
AliSoftware: No, only because it was the url that was used in other podspecs of ours I copy/paste that from. Good catch!

WordPressAuthenticator.podspec
AliSoftware: The LICENSE file is indeed a standard in OSS world. I'll add the appropriate LICENSE file in the repos that don't have one yet.

WordPressAuthenticator.podspec
AliSoftware: Yeah I agree. I really like crediting people who contributed; but that can be done in a separate place — and GitHub already has a Insights->Contributors pane that is updated automatically. So imho keeping only a8c as a single generic author feels simpler especially if we want to standardize stuff.

AliSoftware:
@mokagio That's a very good point. Only used apps.wordpress.com because I found it already used in other podspecs but now that you mention it, probably not the best landing page for a dev.

What about making a post on the mobile.blog — listing our pods, OSS policy, etc … — and point to it?

Other option is to point to `s.source.git.to_s` (ie the git repo of each pod, like it was before, maybe not that bad after all for a dev-focused landing page.

mokagio:
That's a great idea! After suggesting the `.github.io` I went "🤦 we run the most powerful publishing platform in the world! Why not using that?!"

A post, or, may I suggest, a _page_ on mobile.blog is a much more WordPress-y solution 😄 . Something like `mobile.blog/open-source`?

I'll pitch the idea on the apps P2!

It might take a while to reach consensus and put the page up, though, I wouldn't let that stop merging this PR. We can always circle back and update the URL.

---

> Other option is to point to `s.source.git.to_s` (ie the git repo of each pod, like it was before, maybe not that bad after all for a dev-focused landing page.

That would certainly be a lower maintenance solution 😅 Worth keeping in mind!

AliSoftware:
@mokagio Ok, so `s.homepage = s.source.git` errors because it considers that `s.source` is nil (probably because it's only defined later in the podspec, I guess).

Probably not worth trying to come up with some other clever ruby solution there, so I've decided to just repeat the URL in the spec (without the `.git` at the end, though) – see feb0a2a.

mokagio:
👍

WordPressAuthenticator.podspec
```diff
@@ -1,20 +1,23 @@
...
-                    Plus: WordPress.com *signup*  is supported.
+                    Plus: WordPress.com *signup* is supported.
                     DESC
 
   s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
-  s.license       = "GPLv2"
-  s.author        = { "WordPress" => "mobile@automattic.com" }
+  s.license       = { :type => "GPLv2", :file => "LICENSE" }
+  s.author        = { "Automattic" => "mobile@automattic.com" }
+  s.social_media_url = "https://twitter.com/automattic"
```
jkmassel: Does CocoaPods get mad at us if this key is missing? Tbh I don't feel like it's necessary for Pods _almost_ all of the time that are published by companies (individuals who have most of their presence on Twitter might be a different case).

If it's not required, I'd personally vote to just not have it. But if we must have it, this is the way to go for sure.

WordPressAuthenticator.podspec
jkmassel: Yeah, the `MIT` dependencies are tricky – for now I'd say we just put the MIT License in the `LICENSE` file and we'll figure out aligning licenses another day

WordPressAuthenticator.podspec
AliSoftware: It's not required, so we can totally remove it.

@jkmassel One advantage/impact of [having this attribute with a twitter URL set](https://guides.cocoapods.org/syntax/podspec.html#social_media_url) though is that this makes any update to our pods be included in the [@CocoaPodsFeed twitter bot account](https://twitter.com/cocoapodsfeed).
Do we care? That's the question. But I say no harm in keeping 😉

AliSoftware: After Slack discussion, the concensus is:
 - Use `"The WordPress Mobile Team"` as the author name/key for pods hosted in the wordpress-mobile org, and `"Automattic"` for pods hosted in the automattic org
 - Only use the `social_media_url` (pointing to a8c's Twitter) for pods hosted in the a8c org (and no `social_media_url` for wordpress-mobile org pods)

Made corresponding adjustments in 773916c

AliSoftware: Update 2: Systems made us a wordpress.org email pMz3w-cOb-p2#comment-86996 so we will use it for the `s.author` for pods that are hosted in the `wordpress-mobile` GH org (we will still use the `automattic.com` one for pods hosted in Automattic org)

See b998ba013df0071555e8521df8531418642f224d

WordPressAuthenticator.podspec
```diff
@@ -1,20 +1,22 @@
...
-                    Plus: WordPress.com *signup*  is supported.
-                    DESC
+                    Plus: WordPress.com *signup* is supported.
+                  DESC
 
   s.homepage      = "https://github.com/wordpress-mobile/WordPressAuthenticator-iOS"
-  s.license       = "GPLv2"
-  s.author        = { "WordPress" => "mobile@automattic.com" }
+  s.license       = { :type => "GPLv2", :file => "LICENSE" }
+  s.author        = { "The WordPress Mobile Team" => "mobile@wordpress.org" }
```
AliSoftware, line 15: pMz3w-cOb-p2#comment-86996

PR #583, opened 2021-03-29
Site address login: correctly remove `wp-admin` from the url string

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/16145
WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16186

When a user enters a site address to log in, `wp-admin` is removed before string is validated. Previously, if anything came after `wp-admin` the regex match would fail, causing the API not to recognize the URL. This change updates the regex pattern to remove `wp-admin` and anything after it.

Bonus: I added the clear button to text entry fields in the auth views, set to appear when editing.

<kbd>Screen Shot 2021-03-29 at 3 28 02 PM
</kbd>

Can be tested with the referenced WPiOS PR.

Note: `swiftlint autocorrect` found _a lot_ of changes. The commit for this fix is https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/c03682599606e28298c472327f80a459510f12fb.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.36.0-beta.2"
+  s.version       = "1.36.0-beta.3"
 
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
   s.description   = <<-DESC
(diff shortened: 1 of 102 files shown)


aerych (approved):
Hola @ScoutHarris 👋 
Changes look good to me.  Tested with the WPiOS branch and I could authenticate with a url that included wp-admin/index.php (for example). 
:shipit:

aerych:
Also...wow ... that autocorrect!

PR #589, opened 2021-04-07
Change to UI for accessibility

kduncanwelke:
This update contains a storyboard fix to address this accessibility bug reported in the main WordPress iOS project: [Issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15655)

Previously if the user had large text turned on for accessibility and was using a device with a small screen, if they tried to log in to WordPress using their email, the email entry field was impossible to access. It was obscured by the other login options (apple and google) at the bottom of the table, which was expanding vertically due to the increased text size, causing the section to exceed its original size and cover up the email field. 

This update places the uiview for the extra login options into a scrollview, so if the content exceeds the view's height (like with enlarged text), instead of stretching the view and hiding the above login field, the content will simply scroll instead. It also adds percentage-based heights to ensure the table and other login option section remain proportional - the tableview is also allowed to shrink slightly to allow the "or" divider between the sections to encompass larger text too.

To test, first try the existing situation, and test on a small device (for example, iPhone SE) with larger text turned onto one of the higher settings in accessibility. Open the app and select "Continue with WordPress.com". The field where the email is meant to be entered is not visible because it is covered up with the other login options view, and the large text in the "or" divider is cut off.

Repeat the experiment with the same device and large text setting using the edited storyboard - the email field should now be visible, and other login options and the text beneath now should scroll, with the large text in the "or" divider being visible. This change has no apparent negative impact on smaller text sizes and does not apply on large screens, like iPads.


diegoreymendez:
Hi @kduncanwelke!  I'm testing this PR, and I'm seeing no differences between what I see in develop, and what I see in your PR.

This is what develop looks like currently (notice there's a scroll bar and the email field is available):

largeTextEmailShown

Would you let me know if I should be testing this differently?  Thanks!

diegoreymendez:
Actually, I have to apologize.  I wasn't really testing your changes due to a mistake while checking out your branch.

Your changes make the screen look a bit better since the login field is visible by default:

Simulator Screen Shot - iPhone 12 Pro - 2021-05-04 at 14 56 02

diegoreymendez (approved):
Tested on iPhone and iPad and the changes look good to me.

PR #590, opened 2021-04-20
Disable sending of the email if the enableSignUp isn't on

emilylaguna:
If the `enableSignUp` flag is off would still send the user an email to sign up, this adds a check to make sure the flag is enabled before doing that.

You can test using: https://github.com/wordpress-mobile/WordPress-iOS/pull/16344


momo-ozawa (approved):
🚀

PR #592, opened 2021-04-29
Add the ability to customize the NUXButton styles on the Login Prologue

emilylaguna:
**WPiOS Related PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16407**

In order to achieve the design for the Jetpack Prologue screen I needed to be able to change the NUXButton style just for that view. But since the NUXButton relies on `WordPressAuthenticator.shared.style` this was not possible. 

I updated the NUXButton to include a new `NUXButtonStyle` style struct that defines the 3 different states: `normal`, `highlighted`, and `disabled`. This uses an internal struct called `ButtonStyle` that groups the style into `backgroundColor`, `borderColor`, and `titleColor` .

I also added a new way to configure the style of the button, but will fallback to the "legacy" way if that is not used. This prevents the button from breaking for the other areas it is used in the app while adding new functionality. 

### Testing Steps
- You can follow the steps on the related PR
- Since this also can affect WCiOS, please also test there
- The NUXButton is used throughout WPiOS as well you can test the following locations:
    - Prepublishing bottom sheet
    - Domain registration view
    - Site Assembly

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.37.0-beta.2"
+  s.version       = "1.37.0-beta.3"
 
   s.summary       = "WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps."
   s.description   = <<-DESC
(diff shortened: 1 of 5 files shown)


leandroalonso (approved):

frosty (approved):
This looks great to me – I tested with the Jetpack PR, as well as a range of places in WPiOS where we use NUX buttons 👍

PR #594, opened 2021-05-31
Remove Lottie

jkmassel:
Removes the `lottie` dependency from the project in order to also remove it from WordPress and WooCommerce.

I've manually verified that these classes are not used by WP or WC, but if you'd like to be extra sure, you could point both apps at this PR and try building.


mokagio:
Closing in favor of #640 .

PR #596, opened 2021-06-03
Fix Site Address UI Bugs When Displaying In-Page Errors

shiki:
Part of fixing https://github.com/woocommerce/woocommerce-ios/issues/4034. 

There is a linked WP PR for this: https://github.com/wordpress-mobile/WordPress-iOS/pull/16655.

Originally intended to fix just WooCommerce but apparently, WP has the same issue. So this fixes both apps. 🤸 

## Findings

Both apps have this problem when logging in with a site address. This is for WordPress specifically:

1. Log in with a valid self-hosted site address **domain**. Don't include the scheme. Add a typo.
2. After the error is displayed, fix the typo. Still exclude the scheme.
3. Continue logging in, providing the valid credentials. The app will fail to log the user in. The `WordPressOrgXMLRPCValidator.guessXMLRPCURLForSite` fails because we did not provide the scheme. 

This problem only surfaces when the user made a mistake on the first try. Which is quite likely. It also happens if there was previously an XML-RPC error. Or any other error that is displayed below the site address text field. 

Here is a video showing this bug:

https://user-images.githubusercontent.com/198826/121266808-65886380-c878-11eb-9a6b-bdcd0c961036.mov

If there was no error on the first try, the app is able to log the user in because we [automatically fix the `siteAddress`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/26e7de706cf36aee32a44134ae771e655e7d11cd/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L410). 

You'll also notice in the video that the text disappears when the error is displayed. And the Continue button also does not show the activity indicator anymore if the user tries a second time. It looks like the app is doing nothing. 

### Logging In

For some reason, if the error is displayed at some point, the `loginFields.siteAddress` gets reset because this callback gets called: 

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/45d02a709a9c010131a320767c227d2fe072c146/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L312-L315

I still have no idea why that text field change callback is called. But I fixed it by resetting `loginFields.siteAddress` before any view controller is displayed. See [here for the fix](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/45d02a709a9c010131a320767c227d2fe072c146/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L498-L514). Let me know if there's a better way for this. 

### Disappearing Text

I fixed the disappearing text by resetting the value when the cell is recreated in `configureTextField`: 

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/aa020aa62cc50f6bc24e51d0fb6f64bcf871d79b/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L311

### Activity Indicator

The reason the activity indicator on the button does not show up when the user retries because this block gets called right after the activity indicator is shown:

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/26e7de706cf36aee32a44134ae771e655e7d11cd/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L309

So the activity indicator gets dismissed immediately. 

I fixed this by keeping the “view is loading” state in a variable: 

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/aa020aa62cc50f6bc24e51d0fb6f64bcf871d79b/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L22-L29

And making sure that the `configureSubmitButton` uses that state so the activity indicator is not unintentionally dismissed:

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/aa020aa62cc50f6bc24e51d0fb6f64bcf871d79b/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L94-L98

Let me know if there's a better way to do this. I wasn't quite sure how to properly handle this. I wanted to use a `didSet` handler but it looks like [`configureViewLoading`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/aa020aa62cc50f6bc24e51d0fb6f64bcf871d79b/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L135-L139) is an override from the parent class. So I couldn't just remove it. 

Here is the final result: 

https://user-images.githubusercontent.com/198826/121270712-71c3ef00-c87f-11eb-8bd1-77e3d8b87175.mov

## Testing

1. You can use the branch in https://github.com/wordpress-mobile/WordPress-iOS/pull/16655. 
2. Log in with a valid self-hosted site address **domain**. Don't include the scheme. Add a typo.
3. After the error is displayed, fix the typo. Still exclude the scheme.
4. Press Continue.
5. Confirm that the activity indicator is visible and the text field is disabled.
6. Provide the credentials.
7. Confirm that you were able to log in.

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -289,6 +289,7 @@ private extension SiteAddressViewController {
         // Save a reference to the first textField so it can becomeFirstResponder.
         siteURLField = cell.textField
         cell.textField.delegate = self
+        cell.textField.text = loginFields.siteAddress
         cell.onChangeSelectionHandler = { [weak self] textfield in
             self?.loginFields.siteAddress = textfield.nonNilTrimmedText()
             self?.configureSubmitButton(animating: false)
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -503,8 +503,9 @@ private extension SiteAddressViewController {
         // this causes the value of `loginFields.siteAddress` to be reset to what the user entered.
         //
         // Using the user-entered `loginFields.siteAddress` causes problems when we try to log
-        // the user in. For example, validating their self-hosted site credentials somehow
-        // fails. Especially if they only entered the domain (no url scheme).
+        // the user in especially if they just use a domain. For example, validating their
+        // self-hosted site credentials fails because the
+        // `WordPressOrgXMLRPCValidator.guessXMLRPCURLForSite` expects a complete site URL.
         //
         // This routine fixes that problem. We'll use what we already validated from
         // `fetchSiteInfo()`.
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -494,6 +494,25 @@ private extension SiteAddressViewController {
     }
 
     func presentNextControllerIfPossible(siteInfo: WordPressComSiteInfo?) {
+
+        // Ensure that we're using the verified URL before passing the `loginFields` to the next
+        // view controller.
+        //
+        // In some scenarios, the text field change callback in `configureTextField()` gets executed
+        // right after we validated and modified `loginFields.siteAddress` in `validateForm()`. And
+        // this causes the value of `loginFields.siteAddress` to be reset to what the user entered.
+        //
+        // Using the user-entered `loginFields.siteAddress` causes problems when we try to log
+        // the user in. For example, validating their self-hosted site credentials somehow
+        // fails. Especially if they only entered the domain (no url scheme).
+        //
+        // This routine fixes that problem. We'll use what we already validated from
+        // `fetchSiteInfo()`.
+        //
+        if let verifiedSiteAddress = siteInfo?.url {
+            loginFields.siteAddress = verifiedSiteAddress
+        }
+
         guard siteInfo?.isWPCom == false else {
             showGetStarted()
             return
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -19,6 +19,8 @@ final class SiteAddressViewController: LoginViewController {
     private var errorMessage: String?
     private var shouldChangeVoiceOverFocus: Bool = false
 
+    private var viewIsLoading: Bool = false
+
     // MARK: - Actions
     @IBAction func handleContinueButtonTapped(_ sender: NUXButton) {
         tracker.track(click: .submit)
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -183,21 +202,6 @@ extension SiteAddressViewController: UITableViewDataSource {
...
-extension SiteAddressViewController: UITableViewDelegate {
-    /// After the site address textfield cell is done displaying, remove the textfield reference.
-    ///
-    func tableView(_ tableView: UITableView, didEndDisplaying cell: UITableViewCell, forRowAt indexPath: IndexPath) {
-        guard let row = rows[safe: indexPath.row] else {
-            return
-        }
-
-        if row == .siteAddress {
-            siteURLField = nil
```
shiki, line 196: I removed this because this was preventing the text field from being disabled when the user tapped Continue for the second time. There are no implications in removing this, right? 😅

ScoutHarris:
> I still have no idea why that text field change callback is called

The idea is to enable/disable the `Continue` button on the fly as the user enters the URL. What's _supposed_ to happen is `Continue` is disabled until something resembling a URL is entered. But that's not happening. It's enabled as soon as the first character is entered. It looks like the [`validateSiteForSignin`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/595748c577395f59bba5e48b2aef70a542822eca/WordPressAuthenticator/Model/LoginFields%2BValidation.swift#L16) method isn't doing what we're expecting it to do, specifically the call to `idnEncodedURL`.  

```
    func validateSiteForSignin() -> Bool {
        guard let url = URL(string: NSURL.idnEncodedURL(siteAddress)) else {
            return false
        }

        return !url.absoluteString.isEmpty
    }
```    

I checked internal, and it's doing the same thing, so this definitely pre-dates your PR. But that's why that callback is there.

ScoutHarris:
I know this isn't your doing (it appears to be mine 🤦 ), but there is a `print("🔴 SAVC > fetchSiteInfo")` in SiteAddressViewController [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/aa020aa62cc50f6bc24e51d0fb6f64bcf871d79b/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L470). Would you be a dear and remove it please?

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
ScoutHarris: There doesn't appear to be a side effect. It looks like this was there to clean up setting the `siteURLField` as first responder, but your change seems to have made no difference. Sooooo 👍 .

ScoutHarris (approved):
LGTM! 

:shipit:

shiki:
> It looks like the validateSiteForSignin method isn't doing what we're expecting it to do, specifically the call to idnEncodedURL.

Ah, thanks for the explanation! 

> Would you be a dear and remove it please?

Since you asked nicely... removed in d6447e7. 🤣 

---

Thank you very much for the review, @ScoutHarris!

PR #603, opened 2021-07-09
Add setup instructions to README

thehenrybyrd:
I ran into some trouble building WordPressAuthenticator, and wanted to add instructions to make it more clear for other newcomers.
Please let me know if anything could be more clear here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Plus: WordPress.com *signup*  is supported.
 ## Setup
 
 ```sh
-# Install required ruby 2.6.5 
-rbenv install 2.6.5
+# Install required ruby version
+rbenv install -s
 
 # Install bundler and Gemfile dependencies
 gem install bundler
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Plus: WordPress.com *signup*  is supported.
 rbenv install 2.6.5
 
 # Install bundler and Gemfile dependencies
-gem install bundle
+gem install bundler
 bundle install
 
 # Install CocoaPods dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,20 @@ This framework encapsulates everything required to display the Authentication UI
...
 Plus: WordPress.com *signup*  is supported.
 
+## Setup
+
+```sh
+# Install required ruby 2.6.5 
+rbenv install 2.6.5
+
+# Install bundler and Gemfile dependencies
+gem install bundle
```
ealeksandrov: typo 🙂
```suggestion
gem install bundler
```

README.md
```diff
@@ -6,6 +6,20 @@ This framework encapsulates everything required to display the Authentication UI
 
 Plus: WordPress.com *signup*  is supported.
 
+## Setup
+
+```sh
+# Install required ruby 2.6.5 
+rbenv install 2.6.5
```
jkmassel: I'd suggest `rbenv install -s`

By not specifying the version, it's read from the `.ruby-version` file, so this instruction won't need to be updated if we update the project's ruby version.

The `-s` flag says "don't do anything if that version is already installed", which can be handy for folks who just copy/paste commands 😁

jkmassel (changes requested):
I left one suggestion for improvement – thanks for doing this for everyone!

README.md
thehenrybyrd: Oooh nice! Great suggestion, thank you!

ealeksandrov (approved):

jkmassel (approved):

PR #604, opened 2021-07-09
Add accessibilityIndicator for helpButton

thehenrybyrd:
As part of my work updating WPiOS login & signup tests, I'm writing a test that uses the "Help" button visible in the navbar on Get Started screen. For context, that PR is here: https://github.com/wordpress-mobile/WordPress-iOS/pull/16837
I needed to add an accessibilityIdentifier for that help button, and `NUXViewControllerBase` was the best place for it.

This is the first time I've worked with WordPressAuthenticator, so please let me know if I'm missing anything or should take a different approach.


WordPressAuthenticator/NUX/NUXViewControllerBase.swift
```diff
@@ -267,6 +267,7 @@ extension NUXViewControllerBase where Self: UIViewController, Self: UIViewContro
     private func addHelpButton(to superView: UIView) {
         helpButton.setTitle(NSLocalizedString("Help", comment: "Help button"), for: .normal)
         setHelpButtonTextColor(forUnified: false)
+        helpButton.accessibilityIdentifier = "help-button"
```
jkmassel: My only thought on this would be whether it's possible to have more than one help button on screen at a time (perhaps with a modal and a regular VC?)

Typically this wouldn't matter, but because it's a base class there could be cases where two views would share an accessibility ID, which is suboptimal.

thehenrybyrd: I can't think of a situation where that would happen, but definitely good to think about. Since the help button doesn't appear on the Support screen, it's hard to imagine a situation in which "help" would appear and be overlaid  by a modal also containing that "help" button. I think it would have to be a login flow within a login flow, and since the login flow isn't a modal I don't think that can happen?
But let me know if my imagination is lacking 😆 

This view is definitely used plenty, as that help button appears in every instance of login flow. Whether you're logging in self-hosted or via WordPress.com, "help" appears on the Get Started screen, password screen, and 2FA screen, always in that same spot. When tapped, it brings up the Support screen, which is a modal. None of the items that can be tapped there bring up another screen with the "help" button.

Once logged-in, "help" doesn't appear as a button anywhere that I know of in the current UI. In "Me" it's "Help & Support", for instance.

However, the naming is pretty generic which could cause confusion if additional help buttons were added in non-login-flow areas.
Since this "help" button can appear anywhere we use this view, perhaps a more descriptive name would be `authenticator-help-button` or `login-help-button`?

Let me know what you think!

thehenrybyrd: I've renamed that accessibilityIdentifier to `authenticator-help-button`.

jkmassel: In some cases, the login modal is presented on top of a regular VC if the user's auth token has expired (or been revoked), so I think this solution should solve the issue 🙂

jkmassel (approved):
Approved pending conflict resolution

mokagio:
> Approved pending conflict resolution

@thehenrybyrd notice that, because I just shipped version `1.40.0` stable, the version here should be `1.41.0-beta.1` (or `1.41.0-beta.n+1` where `n` is the beta number on `develop` at the time of solving the conflict.

PR #610, opened 2021-08-03
Update GoogleSignIn to 6.0.1

dvdchr:
Closes #609

This updates `GoogleSignIn` to version `6.0.1`, enabling support for M1. Some adjustments were required due to breaking changes introduced in version `6.0.0`. Google has conveniently provided a [migration checklist](https://developers.google.com/identity/sign-in/ios/quick-migration-guide):

- [x] Update all `[GIDSignIn sharedInstance]` (or `GIDSignIn.sharedInstance()`) calls to `GIDSignIn.sharedInstance` property accesses.
- [x] Move all client configuration (client IDs, etc.) to a `GIDConfiguration` object.
- [x] Update calls to removed methods to their new equivalents:

  v5.x.x | v6.0.0 | Status
  --|--|--
  `signIn` | `signInWithConfiguration:presentingViewController:callback:` | ✅  Done. Previous `signIn` method in `GIDSignInDelegate` extension is migrated.
  `restorePreviousSignIn` | `restorePreviousSignInWithCallback:` | ✅  Done. Method is not used.
  `disconnect` | `disconnectWithCallBack:` | ✅  Done. Previous disconnect method in `GIDSignInDelegate` is not implemented, so there's no need to define the callback.
  `getTokensWithHandler:` | `doWithFreshTokens:` | ✅  Done. Method is not used.
  `refreshTokensWithHandler:` | `doWithFreshTokens` | ✅  Done. Method is not used.
- [x] Remove all references to the `GIDSignInDelegate` protocol and its methods.
  - [x] Move the logic from `signIn:didSignInForUser:withError:` to the callback block of `signInWithConfiguration:presentingViewController:callback:`.
  - [x] Move the logic from `signIn:didDisconnectWithUser:withError:` to the callback block of `disconnectWithCallback:`. 
- [x] Manually connect `GIDSignInButton` to a method that calls `signInWithConfiguration:presentingViewController:callback:` using an `IBAction` or similar.

## Todos:
- [x] Ensure that this change does not affect the authentication flow in WordPress.
- [x] Ensure that this change does not affect the authentication flow in WooCommerce.
- [x] Bump podspec version to `1.41.0-beta.2`

PR #611, opened 2021-08-04
Remove text from back buttons

ealeksandrov:
## Description

Ref: https://github.com/woocommerce/woocommerce-ios/issues/3944

Currently login flow replaces VC titles on back buttons with "Back" text. This:
1. differs from WC general back buttons treatment (only minimal arrow)
2. breaks (shows "back" for every view) navigation history that can be displayed by long press on a back button.

## Screenshots

/ | before | after
--|--|--
button|before_button|after_button
history|before_history|after_history

## Test

Checkout https://github.com/woocommerce/woocommerce-ios/pull/4726 in WC-iOS to test this change on WC.
Or switch Pod dependency on a branch (like in PR above) and do the same test in WP-iOS.

1. Go through login flow and verify that back buttons have only arrow on them, without text (`<`)
2. Go deep in login flow and long press on back button to see navigation history. Verify that it shows correct titles for previous views (unfortunately now we have similar "Log In" for most of them).

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
-  s.name          = "WordPressAuthenticator"
-  s.version       = "1.41.0-beta.3"
+  s.name          = 'WordPressAuthenticator'
+  s.version       = '1.41.0-beta.3'
   
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressAuthenticator"
-  s.version       = "1.41.0-beta.1"
+  s.version       = "1.41.0-beta.3"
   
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
-  s.name          = "WordPressAuthenticator"
-  s.version       = "1.41.0-beta.1"
+  s.name          = 'WordPressAuthenticator'
+  s.version       = '1.41.0-beta.3'
```
ealeksandrov, line 3: Because https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/610 added `1.41.0-beta.2`.
Will update to proper latest version + add a tag before final merge.

ScoutHarris (approved):
Hey @ealeksandrov .

I tested this with WPiOS `develop`.

✅ Go through login flow and verify that back buttons have only arrow on them, without text (<)
✅ Go deep in login flow and long press on back button to see navigation history. 

This is the navigation history from the 2FA view. If that looks correct, then 👍 .

Screen Shot 2021-08-04 at 4 44 01 PM

ealeksandrov:
@ScoutHarris thank you for the review! 🚀

PR #615, opened 2021-08-20
Add Buildkite

jkmassel:
Adds Buildkite to this repo, including:
- Rubocop (incl fixes)
- Fastlane
- Removing CircleCI Publishing

You can verify that Pod Publishing works with this job: https://buildkite.com/wordpress-mobile/wordpress-authenticator-ios/builds?branch=1.41.0-beta.999

PR #616, opened 2021-08-23
Merge 1.42.0 stable version bump into develop

mokagio:


mokagio:
Note that the Buildkite checks failed because we're in the transition phase in which Buildkite is registered to the web hooks but the PR that adds it, #615, hasn't been merged yet.

PR #617, opened 2021-08-24
Merge version 1.42.0 into trunk

mokagio:
Merges the 1.42.0 changes into `trunk`.

No test required other than verifying the `.podspec` changes and that CI passes.


WordPressAuthenticator.podspec
```diff
@@ -1,6 +1,8 @@
+# frozen_string_literal: true
+
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '1.41.0'
+  s.version       = '1.42.0'
```
AliSoftware, line 5: The ethernal question about "should a `s.swift_version` bump from `4.2` to `5.0` warrant a major or minor version bump in the library?"… (never had a clear answer about this)

Given this lib is mostly used internally I think it's ok, but 🤷

Podfile
```diff
@@ -43,8 +49,6 @@ target 'WordPressAuthenticatorTests' do
   project 'WordPressAuthenticator.xcodeproj'
   wordpress_authenticator_pods
 
-  pod 'OHHTTPStubs', '8.0.0'
-  pod 'OHHTTPStubs/Swift', '8.0.0'
```
AliSoftware, line 47: Still a bit sad about letting those go… 😄  ^^

AliSoftware (approved):
Commit list makes sense and diff matches all the changes made since last version – including normal version bump in podspec. :shipit:

PR #620, opened 2021-10-05
Make the domain searches instantaneous

diegoreymendez:
Improvements to `SearchTableViewCell.swift`:

- Adds a `liveSearch` property that adds support for immediate / live searches.  This property defaults to `false` to try to avoid affecting existing code.
- Adds an `allowSpaces` property that adds support for stripping whitespaces from the user's input.  This property defaults to `false` to try to avoid affecting existing code.

## Testing:

To test this, use the WordPress iOS PR.

1. Start the domain registration flow.
2. In the domain suggestions screen, start typing with the device keyboard.
3. Try typing spaces in different parts of the search text.
4. Try clearing the field.

Make sure you can't input whitespaces.  Make sure live searches work as expected.

## Associated PRs:

WordPress iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17266

### DIFF
--- a/WordPressAuthenticator/UI/SearchTableViewCell.swift
+++ b/WordPressAuthenticator/UI/SearchTableViewCell.swift
@@ -98,12 +98,16 @@ extension SearchTableViewCell: UITextFieldDelegate {
         let hasValidEdits = sanitizedString.count > 0 || range.length > 0
 
         if hasValidEdits {
-            guard let currentText = (textField.text as? NSString) else {
+            guard let currentText = (textField.text as? NSString),
+                  let start = textField.position(from: textField.beginningOfDocument, offset: range.location),
+                  let end = textField.position(from: start, offset: range.length),
+                  let textRange = textField.textRange(from: start, to: end) else {
+
                 // This shouldn't really happen but if it does, let's at least let the edit go through
                 return true
             }
 
-            textField.text = currentText.replacingCharacters(in: range, with: sanitizedString)
+            textField.replace(textRange, withText: sanitizedString)
             startLiveSearchAfterEdit()
         }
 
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator/UI/SearchTableViewCell.swift
```diff
@@ -62,16 +73,65 @@ private extension SearchTableViewCell {
...
+            guard let start = textField.position(from: textField.beginningOfDocument, offset: range.location),
+                  let end = textField.position(from: start, offset: range.length),
+                  let textRange = textField.textRange(from: start, to: end) else {
+
+                // This shouldn't really happen but if it does, let's at least let the edit go through
+                return true
+            }
+
+            textField.replace(textRange, withText: sanitizedString)
+            startLiveSearchAfterEdit()
```
frosty: Should we check for `liveSearch` before calling `startLiveSearchAfterEdit`? I know that method also checks it internally, but I wonder if we shouldn't assume that.

frosty (approved):
Works for me! Just one small query here, up to you if you want to tweak it :) Don't forget to bump the podspec version.

WordPressAuthenticator/UI/SearchTableViewCell.swift
diegoreymendez: It would probably be cleaner to check outside of it (as you're suggesting), and document it as a pre-condition for calling the method.

I'll also remove the check inside because if it's properly documented as a precondition it shouldn't be checked twice.

diegoreymendez: I've also generalized the method a bit removing references to "after edits" because it's not really a requirement in any way.

frosty (approved):

PR #625, opened 2021-12-03
Remove code targeting iOS 12 or lower

hassaansaleh:
## Description
There was some code targeting iOS versions prior to iOS 13. And since the deployment target is iOS 13, these instances can safely be removed. This PR removes these instances. The changes in this PR introduces no visible changes, it's more of a cleanup.
The changes include:
- Removal of unneeded `@available` and `if #available` instances.
- ~~Refactoring `GravatarEmailTableViewCell` and `TextFieldTableViewCell` to remove `onChangeSelectionHandler` and instead depend on `UITextFieldDelegate` method `textFieldDidChangeSelection` which is available staring iOS 13.~~

References: https://github.com/wordpress-mobile/WordPress-iOS/issues/17516

## Testing
To test this use the WordPress iOS App, but make sure to replace this line in the Podfile
`pod 'WordPressAuthenticator', '~> 1.42.1'`
With this
`pod 'WordPressAuthenticator', :git => 'https://github.com/hassaansaleh/WordPressAuthenticator-iOS.git', :branch => 'issue/17516-remove-iOS12-code'`

### What to test
1. ~~Check that interacting with textfields is working correctly in the following screens:~~
    1. ~~GetStartedViewControler~~
    2. ~~PasswordViewController~~
    3. ~~SiteCredentialsViewController~~
    4. ~~SiteAddressViewController~~
2. ~~For PasswordViewController, check when the email is changed from a Password Manager, that it's handled correctly.~~

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -503,7 +503,6 @@ import WordPressKit
     }
 }
 
-@available(iOS 13.0, *)
 public extension WordPressAuthenticator {
 
     func getAppleIDCredentialState(for userID: String, completion:  @escaping (ASAuthorizationAppleIDProvider.CredentialState, Error?) -> Void) {
--- a/WordPressAuthenticator/NUX/NUXButton.swift
+++ b/WordPressAuthenticator/NUX/NUXButton.swift
@@ -36,20 +36,12 @@ public struct NUXButtonStyle {
 
     open override var isEnabled: Bool {
         didSet {
-            if #available(iOS 13, *) {
-                activityIndicator.color = activityIndicatorColor(isEnabled: isEnabled)
-            }
+            activityIndicator.color = activityIndicatorColor(isEnabled: isEnabled)
         }
     }
 
     @objc let activityIndicator: UIActivityIndicatorView = {
-        let indicator: UIActivityIndicatorView
-        if #available(iOS 13, *) {
-            indicator = UIActivityIndicatorView(style: .medium)
-        } else {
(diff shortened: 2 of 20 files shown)


diegoreymendez (changes requested):
I think the cleanup of the code that's for versions of iOS earlier than 13 looks good.  It's also in-line with the main intent behind the PR.

I'm not so sure about the change from `onChangeSelectionHandler` to `textFieldDidChangeSelection`, for a few reasons:

- This goes beyond cleaning up code that's not used anymore, as it's modifying live code.
- Tracking `editing-changed` is not semantically equivalent to tracking `selection-changed`.  While both may coincidentally work for us, the first type of event is what we're aiming to track.  As an example: I can trigger `textFieldDidChangeSelection` by moving the caret around in the text field, but that shouldn't result in a call to `handleEmailFieldDidChange`.
- And last but not least: this code is a bit of a critical part of our Apps, and the pre-existing code works well.  I'd be wary to make changes that could introduce problems.

Let me know what you think.

hassaansaleh:
@diegoreymendez Well the motivation behind these changes were based on a comment I found in the code, it felt like this was the original intent for whoever wrote this part.

> When we no longer support iOS 12, `textFieldDidChangeSelection`, and `onChangeSelectionHandler` can be deleted in favor of adding the delegate method to view controllers.

However I feel that your concerns are totally valid, these changes do diverge from the current task at hand. So will revert them 👍

diegoreymendez:
Thanks @hassaansaleh: that comment perfectly explains the direction you took in the PR.

Also thanks for understanding and rolling the code back.

diegoreymendez (approved):
This PR is good to go, nice work! :)

PR #636, opened 2022-01-23
Use `Follow Readable Width` for views

wargcm:
Related PR: wordpress-mobile/WordPress-iOS#17870

## Description

Currently the margins for iPad are being manually calculated based on the view's width. This change removes that calculation and uses the readable content guide.

Additionally, there were some constraints breaking on the button view when it was embedded into a view which was larger than itself. I added some constraints to this view with lower priorities so that it breaks the correct constraints and does not mess up the buttons.

## Screenshots - iPhone

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

## Screenshots - iPad Portrait

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

## Screenshots - iPad Landscape

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

| Before | After |
|--------|------|
|  |  |

### DIFF
--- a/WordPressAuthenticator/NUX/NUXButtonView.storyboard
+++ b/WordPressAuthenticator/NUX/NUXButtonView.storyboard
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17701" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="19455" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
     <device id="retina4_7" orientation="portrait" appearance="light"/>
     <dependencies>
         <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="17703"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="19454"/>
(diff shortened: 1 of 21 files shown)


leandroalonso:
@wargcm there's any reason why the buttons are not following the same margin as the upper content?

An example (compare the text and the beginning of the button):



The same for the social buttons.

I tested some of the flows and it all worked as expected, except for these buttons' alignment.

leandroalonso:
~@mattmiklic since you worked on the Unified Login & Sign Up project could you please take a look at the screenshots here and see if the changes are ok?~

~Basically, it's a technical change that affects the left and right margins. We were calculating those manually, with this change we rely on some default values provided for iOS.~

leandroalonso:
@wargcm I revisited all screens and checked the code and I think it would be better to leave things as it is:

1. This is not a really necessary change. It's good, but nothing groundbreaking
2. There are other flows that might be affected. `NuxButtonViewController` is used on WPiOS, for example.
3. We would need more people to double-check all these flows on Woo and also on WPiOS (we have many flows, so it's not that simple and it will take time)

So we should probably play safe here and avoid running into bad surprises. Let me know what do you think.

But thanks for tackling that anyway, superb job!

wargcm:
@leandroalonso It's possible those screens you saw the issue on are covered in the other PR: [wordpress-mobile/WordPress-iOS/#17870](https://github.com/wordpress-mobile/WordPress-iOS/pull/17870). That or it might've been a screen I missed for adjusting the margins on the button view. I was concerned about changing `NuxButtonViewController` too much due to downstream dependencies. The margins for the buttons on this view are `16` so I adjusted anywhere that embedded that view to have the margins offset by 16. I did that because I figured changing those individual screens would be easier than finding everything that uses the button view.

I agree and am good to close out this PR and the other. If we wanted to revisit it in the future, then I think more focused PRs would be a good idea. Maybe screen by screen? It would take longer but would allow reviewers to thoroughly test each screen and know the impact.

PR #640, opened 2022-03-07
Remove Lottie

mokagio:
Remove the Lottie third-party library and the screens using it.

These screens were no longer used by the client apps, so the removal is "safe".

Nevertheless, I bumped the major version to represent the breaking change from dropping the Lottie library and consequently removing the `UIViewController` using it.

Hat tip @jkmassel for having done most of the work already. 🎩 

See, both with green unit tests:

- https://github.com/woocommerce/woocommerce-ios/pull/6379
- https://github.com/wordpress-mobile/WordPress-iOS/pull/18109


WordPressAuthenticator.xcodeproj/project.pbxproj
```diff
@@ -14,6 +14,7 @@
 		3F550D4E23DA429B007E5897 /* AppSelectorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F550D4D23DA429B007E5897 /* AppSelectorTests.swift */; };
 		3F550D5123DA4A9C007E5897 /* LinkMailPresenter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F550D5023DA4A9C007E5897 /* LinkMailPresenter.swift */; };
 		3F550D5323DA4AC6007E5897 /* URLHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F550D5223DA4AC6007E5897 /* URLHandler.swift */; };
+		3F9439BE27D6F9B60067183A /* LoginPrologueViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F9439BD27D6F9B60067183A /* LoginPrologueViewController.swift */; };
```
mokagio, line 17: Unfortunately, `LoginPrologueViewController` is required in the codebase despite being used directly by the clients of this library. The reason is that the Login storyboard uses it. Without it, the storyboard fails to decode.  I originally removed it but had to put it back and didn't feel it granted a `git revert -n` dance, which is why it looks moved here.

**I think there's space for way more files removal, but I don't want to go overboard with this PR.**

I also think it would be fine to ship this as 2.0.0 and then remove more files, even though that might consist as another breaking change. After all, we are the only consumers of this library, and the successful tests done for this PR shows the old login flow code is unused. Also, even though I haven't checked, I'd assume most if not all of the code now removable would be `internal`. Removing that wouldn't affect clients.

WordPressAuthenticator.xcodeproj/project.pbxproj
```diff
@@ -1144,7 +1141,6 @@
 				"${BUILT_PRODUCTS_DIR}/WordPressKit/WordPressKit.framework",
 				"${BUILT_PRODUCTS_DIR}/WordPressShared/WordPressShared.framework",
 				"${BUILT_PRODUCTS_DIR}/WordPressUI/WordPressUI.framework",
-				"${BUILT_PRODUCTS_DIR}/lottie-ios/Lottie.framework",
```
AliSoftware, line 1147: Worth updating the `pbxproj`'s project format version to a more recent one so that it supports using `xcfilelist` (making CocoaPods then use them and have those changes be outside of the `pbxproj`)?

Maybe for a separate PR though (similar to the idea you floated around about removing more files later… though the change in pbxproj format would not be a breaking one, so could be worth doing befor a new major version bump…? 🤷 )

WordPressAuthenticator/Signin/LoginProloguePageViewController.swift
```diff
@@ -2,7 +2,11 @@ import UIKit
 import WordPressShared
 
 class LoginProloguePageViewController: UIPageViewController {
-    @objc var pages: [UIViewController] = []
+    // This property is a legacy of the previous UX iteration. It ought to be removed, but that's
+    // out of scope at the time of writing. It's now `private` to prevent using it within the
+    // library in the meantime
+    @objc private var pages: [UIViewController] = []
```
AliSoftware, line 8:
```suggestion
    @available(*, deprecated, message: "This property is a legacy of the previous UX iteration and ought to be removed (but that was out of scope at the time of writing).")
    // Property is `private` to prevent using it within the library in the meantime, see deprecation message above
    @objc private var pages: [UIViewController] = []
```

AliSoftware (approved):
Overall looks good, but left some very small suggestions for pbxproj format update and use of deprecation annotation; take it or leave it.

WordPressAuthenticator.xcodeproj/project.pbxproj
mokagio: Oh good point. Thanks. I'll do that in a dedicated PR, to isolated that project structure change.

WordPressAuthenticator/Signin/LoginProloguePageViewController.swift
mokagio: That's a nice idea, thanks.

I tried it locally, and got a lot (well, 10) warnings:

image

_All the warnings in the mini map are because of the deprecated `page`._

Part of me welcomes them. "Yes! Make some noise in the build so we get rid of dead code ASAP." Another part of me doesn't want them because, while it's true that we ought to get rid of that code, there's no pressing necessity to do it and I wouldn't want other, legit warnings to get lost among them. 

Currently, I'm leaning toward not throwing the warnings, so I'm going to merge as is. Happy to revisit, though.

PR #652, opened 2022-07-28
Pass a custom error for invalid WP.com email errors

jaclync:
For https://github.com/woocommerce/woocommerce-ios/issues/7318

## Description

In order to track the source of invalid email errors for WCiOS local notifications experiment, we need a way to differentiate the two flows - "Enter Your Store Address" or "Continue With WordPress.com". This PR added a new `SignInError` and `SignInSource` and parsed the error for invalid email based on the [WCiOS implementation](https://github.com/woocommerce/woocommerce-ios/blob/eec2b2178aff5694294d5bd6848df0bb88524a1b/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L492).

WPiOS [does not handle errors in a custom way](https://github.com/wordpress-mobile/WordPress-iOS/blob/d6b813b903c34d4d66ac07c5e452b378fd952e63/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift#L487-L490) like WCiOS.

## Testing steps

Please refer to the WCiOS draft PR to test the integration.

### DIFF
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -397,6 +397,7 @@ class LoginPrologueViewController: LoginViewController {
             DDLogError("Failed to navigate from LoginPrologueViewController to GetStartedViewController")
             return
         }
+        vc.source = .wpCom
 
         navigationController?.pushViewController(vc, animated: true)
     }
@@ -430,15 +431,6 @@ class LoginPrologueViewController: LoginViewController {
     private func presentLoginEmailView() {
         guard let toVC = LoginEmailViewController.instantiate(from: .login) else {
             DDLogError("Failed to navigate to LoginEmailVC from LoginPrologueVC")
-            return
-        }
-
-        navigationController?.pushViewController(toVC, animated: true)
-    }
-
-    private func presentGetStartedView() {
-        guard let toVC = GetStartedViewController.instantiate(from: .getStarted) else {
-            DDLogError("Failed to navigate to GetStartedViewController")
             return
         }
 
--- a/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
@@ -2,6 +2,37 @@ import UIKit
 import SafariServices
 import WordPressKit
 
+/// The source for the sign in flow for external tracking.
+public enum SignInSource {
+    /// Initiated from the WP.com login CTA.
+    case wpCom
+    /// Initiated from the WP.com login flow that starts with site address.
+    case wpComSiteAddress
+    /// Other unspecified sources.
+    case other
+}
+
+/// The error during the sign in flow.
+public enum SignInError: Error {
+    case invalidWPComEmail(source: SignInSource)
+
+    init?(error: Error, source: SignInSource) {
+        let error = error as NSError
+
+        switch error.code {
+        case WordPressComRestApiError.unknown.rawValue:
+            let restAPIErrorCode = error.userInfo[WordPressComRestApi.ErrorKeyErrorCode] as? String
+            if restAPIErrorCode == "unknown_user" {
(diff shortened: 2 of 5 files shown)


WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -436,15 +437,6 @@ class LoginPrologueViewController: LoginViewController {
...
 
-    private func presentGetStartedView() {
-        guard let toVC = GetStartedViewController.instantiate(from: .getStarted) else {
-            DDLogError("Failed to navigate to GetStartedViewController")
-            return
-        }
-
-        navigationController?.pushViewController(toVC, animated: true)
-    }
-
```
itsmeichigo, line 447: I wonder why this is not needed anymore?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -2,6 +2,37 @@ import UIKit
...
 import WordPressKit
 
+/// The source for the sign in flow for external tracking.
+public enum SignInSource {
+    /// Initiated from the WP.com login CTA.
+    case wpCom
+    /// Initiated from the WP.com login flow that starts with site address.
+    case wpComSiteAddress
+    /// Other unspecified sources.
+    case other
```
itsmeichigo: Nit: I'm thinking this case is not really necessary, since we want to be as specific as possible in `enum`s. If we want to avoid breaking changes, when integrating this library we could use `@unknown default` case to handle new cases in the future. WDYT?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -36,6 +67,8 @@ class GetStartedViewController: LoginViewController, NUXKeyboardResponder {
     // This is public so it can be set from StoredCredentialsAuthenticator.
     var errorMessage: String?
 
+    var source: SignInSource = .other
```
itsmeichigo: Since I suggested removing `other` case above, maybe we should keep `source` optional here 🤔.

itsmeichigo (approved):
I understand that the new error type is needed to avoid breaking changes (compared to updating the delegate method `handleError` with a new `source` parameter), and this works with the integration in Woo, so I'm approving this. Leaving a few nit-picking comments regarding the `other` case for `source` below though, please consider it if you think it makes sense.

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
jaclync: I forgot to make a note on this, this is dead code 😅 I was checking where `GetStartedViewController` is initialized in WPAuthenticator, and thought it's good to remove any unused code.

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
jaclync: Yea, I think it makes sense to not include this default case and only errors with specific sources are passed to the client app. I made an update in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/652/commits/8591d33e66b2baf7551b671b4c9adb36443a73fd

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
jaclync: Updated in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/652/commits/8591d33e66b2baf7551b671b4c9adb36443a73fd 👍

jaclync:
> I understand that the new error type is needed to avoid breaking changes (compared to updating the delegate method handleError with a new source parameter), and this works with the integration in Woo, so I'm approving this.

Yes 😅 I was debating whether to make a breaking change for a better solution or a more implicit change with the error type as in this PR. Since we're on a tight schedule in Q3, I went with the latter one.

PR #653, opened 2022-07-29
Send `siteURL` to `showLoginForJustWPCom`

itsmeichigo:
For https://github.com/woocommerce/woocommerce-ios/pull/7373/

# Description

On WCiOS we want to show the login flow for WP.com authentication manually - but when the `presentLoginEpilogue` delegate method is called, the `wpcom` credentials return https://wordpress.com for `siteURL`. 

This PR fixes that by sending `siteURL` to the `showLoginForJustWPCom` method, which will be saved in `loginFields` and returned in the result credentials after authentication succeeds.

# Testing steps

Please refer to the relating WCiOS PR to test the integration.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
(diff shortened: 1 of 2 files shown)


jaclync (approved):
Changes LGTM 👍

PR #654, opened 2022-08-04
Add Buildkite Test Analytics

mokagio:
Sets up Test Analytics.

image

image

### DIFF
--- a/.buildkite/build-and-test.sh
+++ b/.buildkite/build-and-test.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -eu
+
+# We need to use a script as opposed to calling `build_and_test_pod` inline in
+# the pipeline via the `command` node because our CI-VM setup doesn't forward
+# the environment in that mode.
+
+# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16
+gem install bundler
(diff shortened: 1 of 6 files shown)


fastlane/Fastfile
```diff
@@ -9,7 +9,8 @@ platform :ios do
       scheme: 'WordPressAuthenticator',
       prelaunch_simulator: true,
       buildlog_path: File.join(__dir__, '.build', 'logs'),
-      derived_data_path: File.join(__dir__, '.build', 'derived-data')
+      derived_data_path: File.join(__dir__, '.build', 'derived-data'),
+      output_style: 'raw'
```
crazytonyli: This change was for debugging, wasn't it? It should be okay to delete it now?

crazytonyli (changes requested):

fastlane/Fastfile
mokagio: Yes. Great catch, thank you! 🙇‍♂️

mokagio: Addressed by dropping the commit via interactive rebase 👌

crazytonyli (approved):

PR #656, opened 2022-08-05
Pass a custom error for invalid WP.com password errors

jaclync:
For https://github.com/woocommerce/woocommerce-ios/issues/7318

## Description

In order to handle the invalid WP.com password errors for WCiOS local notifications experiment Iteration 3, a `source: SignInSource` is added to `PasswordViewController` and the remote error is passed to the delegate. No external navigation is expected at this point, so the completion block `authenticationDelegate.handleError` is a no-op now.

### Impact to WPiOS

WPiOS [does not handle errors in a custom way](https://github.com/wordpress-mobile/WordPress-iOS/blob/d6b813b903c34d4d66ac07c5e452b378fd952e63/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift#L487-L490) like WCiOS.

## Testing steps

Please refer to the WCiOS draft PR to test the integration: https://github.com/woocommerce/woocommerce-ios/pull/7422

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '2.1.0-beta.5'
+  s.version       = '2.1.0-beta.6'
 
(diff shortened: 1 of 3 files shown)


itsmeichigo (approved):
LGTM :shipit:

PR #660, opened 2022-08-09
[Tooling/CI] Fix Publish step

AliSoftware:
With the newer Xcode CI image introduced in #654, we need to install an even newer version of bundler to work around https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16.

This was fixed by @mokagio in various places in the `pipeline.yml` back when he worked on #654, but the `publish-pod.sh` file was missed in this update of the workaround, so we're patching this here


selanthiraiyan (approved):
Thank you for the super fast fix. 🚀

mokagio:
Great catch!

PR #662, opened 2022-08-10
Prefer magic link to password under a configuration

jaclync:
For https://github.com/woocommerce/woocommerce-ios/issues/7437

In WCiOS, we're experimenting with emphasizing the magic link in place of the password because the magic link has a much higher sign-in rate. In order not to break other apps like WPiOS, the changes are under a new configuration `WordPressAuthenticatorConfiguration.isWPComMagicLinkPreferredToPassword` that is default to `false`.

### Design hHmVq03kKpvTDzV3uV4wqw-fi-14%3A2092

Screen Shot 2022-08-10 at 5 16 52 PM

### Notes on the code changes

Sorry about the slightly 500+ diffs 🙇🏻‍♀️ the main changes are on the new view controller `MagicLinkRequestedViewController ` and its xib.

I decided to implement the new screen `MagicLinkRequestedViewController` with a lighter weight xib than a storyboard. I ran into some rabbit holes trying to use the shared button styles with `NUXButton`, which is initialized with a Storyboard. After some trials and errors, I was able to use `NUXButtonViewController` as a child view controller in a container view. Then I ran into a problem with the "Use password to sign in" CTA because it's using link style, and the WPAuthenticator doesn't have a shared style for it. After all, I implemented the primary CTA with `NUXButtonViewController` and the "Use password to sign in" CTA with a `UIButton` and a new extension `applyLinkButtonStyle` similar to WCiOS.

### Notes on the UX

Right now, after requesting a magic link successfully from the password screen, it still navigates to the pre-existing magic link requested screen `LoginMagicLinkViewController`. I'm going to confirm with design later, because otherwise it feels like an infinite loop between the password and the new magic link requested screen.

## Testing steps

Please refer to the testing steps in the WCiOS PR.

## Example screenshots

light | dark
-- | --
 | 

tablet:


WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -516,16 +518,21 @@ private extension GetStartedViewController {
     /// Show the Password entry view.
     ///
     func showPasswordView() {
```
selanthiraiyan: 1. Nit: Now that the method doesn't always show the password view, would it make sense to rename it and update the comments?
2. Question: This method is also being called from [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/74830ef88500460a525ad4b0f103b0b1e25a44d1/WordPressAuthenticator/Unified%20Auth/View%20Related/Get%20Started/GetStartedViewController.swift#L840). I haven't tested that flow. But, do we want the magic link behaviour there as well?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -516,16 +518,21 @@ private extension GetStartedViewController {
...
-        vc.trackAsPasswordChallenge = false
-
-        navigationController?.pushViewController(vc, animated: true)
+        configureViewLoading(true)
+        let coordinator = PasswordCoordinator(navigationController: navigationController,
+                                              source: source,
+                                              loginFields: loginFields,
+                                              tracker: tracker,
+                                              configuration: WordPressAuthenticator.shared.configuration)
+        passwordCoordinator = coordinator
```
selanthiraiyan, line 545: ❓ Why are we storing this in `passwordCoordinator`? 

I tried removing this and noticed that we need this [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/74830ef88500460a525ad4b0f103b0b1e25a44d1/WordPressAuthenticator/Unified%20Auth/View%20Related/Password/PasswordCoordinator.swift#L72)

- Instead of asking `PasswordCoordinator` to present the password view, can we pass the necessary properties to `MagicLinkRequestedViewController` and ask it present the screen?
- We could even create a separate coordinator just for showing `PasswordViewController` and store it in `MagicLinkRequestedViewController`

I am sure that you considered the above and several other options, and I am curious to know your thoughts. 😄

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        self.loginWithPassword = loginWithPassword
+        super.init(nibName: "MagicLinkRequestedViewController", bundle: WordPressAuthenticator.bundle)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override var sourceTag: WordPressSupportSourceTag {
+        .loginMagicLink
```
selanthiraiyan: Nit:
This being a new screen should we consider adding `sourceTag`? 
This might be helpful when we work on help center experiment. 😅 Internal ref - pe5sF9-n8-p2

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+        setupEmailLabel()
+        setupCannotFindEmailLabel()
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        tracker.set(flow: .loginWithMagicLink)
+
+        if isMovingToParent {
```
selanthiraiyan: Nit: Kindly consider using `isBeingPresentedInAnyWay`.

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.xib
```diff
@@ -0,0 +1,148 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
selanthiraiyan, line 1: Screenshot 2022-08-11 at 9 50 00 AM
IMHO, it might be helpful to fix these constraints to make changes to this file in future.

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+final class MagicLinkRequestedViewController: LoginViewController {
+
+    // MARK: Properties
+
+    @IBOutlet private weak var titleLabel: UILabel!
+    @IBOutlet private weak var subtitleLabel: UILabel!
+    @IBOutlet private weak var emailLabel: UILabel!
+    @IBOutlet private weak var cannotFindEmailLabel: UILabel!
+    @IBOutlet weak var buttonContainerView: UIView!
+    @IBOutlet weak var loginWithPasswordButton: UIButton!
```
selanthiraiyan: Nit: We could make `buttonContainerView` and `loginWithPasswordButton` private as well.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordCoordinator.swift
```diff
@@ -0,0 +1,89 @@
...
+         configuration: WordPressAuthenticatorConfiguration) {
+        self.navigationController = navigationController
+        self.source = source
+        self.loginFields = loginFields
+        self.tracker = tracker
+        self.configuration = configuration
+    }
+
+    func start() async {
+        if configuration.isWPComMagicLinkPreferredToPassword {
```
selanthiraiyan, line 24: Super nit: What do you think about using `guard`?

selanthiraiyan (approved):
Great work introducing a coordinator. 🙇 

It works great.

I left a few questions and non-blocking nits.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordCoordinator.swift
jaclync: Since this is an experiment, I don't think the magic link preferred flow is the assumed path. It's also a configuration that WPiOS isn't adopting yet.

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
jaclync: yes exactly, the coordinator needs to be retained during the lifetime of its possible navigation flow. because there's logic about which view controller to show, I thought it's easier to reuse this flow by using a [coordinator](https://www.hackingwithswift.com/articles/71/how-to-use-the-coordinator-pattern-in-ios-apps). we could reuse it in other cases where it makes sense to prefer a magic link to password.

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.xib
jaclync: This is really tricky, these IB errors are confusing because these 3 views are in a `UIStackView` and their Y position or height is supposed to be dynamic. The actual layout was all fine from my testing on various font sizes in different simulators, the scroll view is scrollable on larger font sizes. I ended up adding minimum placeholder constraints to be removed at build time in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/51f90cd6d9a0a2e0494069afce5eff894baa8348, please let me know if you have better suggestions on this Auto Layout challenge 🙇🏻‍♀️

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.swift
jaclync: Updated in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/6ff631a2cde8e36c40108411a027d14615f723a9

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.swift
jaclync: sure, added in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/31ef4e483e4c1f412afc670795d6ae1196de1bdb

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
jaclync: great question and suggestion! the SIWA flow that triggered the password view isn't easily testable, so I reverted the changes on this in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/4b97bc5b287256879e822a9501a29a4b31eb421d

WordPressAuthenticator/Unified Auth/View Related/Login/MagicLinkRequestedViewController.swift
jaclync: good catch, updated in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/f40cc75e62d12803550b2eb2021615383d5f0a78

jaclync (commented):
Thank you for the review and for catching a few things @selanthiraiyan! I responded to your questions in https://github.com/woocommerce/woocommerce-ios/pull/7449#pullrequestreview-1069139188 to consolidate all the responses in the same PR 🙂 See if you'd like to take another look or I'll merge it early Friday, I tested a few more times.

> In the `Login with site address` flow

> After the following step,
> > Enter a valid non-A8C WP.com email and tap "Continue" --> after a little bit (requesting magic link), a new magic link requested screen should be shown with two CTAs at the bottom

> the below event is tracked,
```
Tracked unified_login_interaction, properties: [AnyHashable("click"): "request_magic_link", AnyHashable("step"): "start", AnyHashable("flow"): "login_site_address", AnyHashable("source"): "default"]
```
> Would it be alright to track a `click` property even though the user didn't manually request magic link?

Thanks for catching this logging mistake, this click event shouldn't have been tracked. I removed this in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/2cb71a6810a78c2e5af8be08c1ea3f7b8d59cc20

-------

> ❓ Just curious
> Do you have a sequence of analytics events planned to measure the success of this iteration? (To compare with other iterations and etc) 

According to the metric for this experiment/iteration in pe5sF9-iy-p2, we want to measure the % of users that enter the magic link login flow and end up successfully logging in goes up.

In this PR, I added a new event to track this step in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/662/commits/f40cc75e62d12803550b2eb2021615383d5f0a78:

`🔵 Tracked unified_login_step, properties: [AnyHashable("source"): "default", AnyHashable("flow"): "login_magic_link", AnyHashable("step"): "magic_link_auto_requested"]`

so we can measure the successful sign-in rate in a funnel with this new step.

selanthiraiyan:
Thanks for making the changes Jaclyn. This looks great now. 🚀

PR #663, opened 2022-08-10
[Sign in with site credentials] - Don't clear username and password after an invalid login attempt.

selanthiraiyan:
For https://github.com/woocommerce/woocommerce-ios/pull/7444

# Description

In the sign-in with site credentials flow, in the Site credentials screen, if I enter an incorrect password, the username and password fields are emptied/cleared. This doesn't leave a chance for the user to edit and fix types.

This PR attempts to stop clearing the username and password when an error occurs. 

- The username/password is cleared when the tableView is reloaded after a failed signin attempt. We can configure the cells with the current username/password from `loginFields` property to fix this.
- We have also updated the `prepareForReuse` method to properly reset all the properties that are customised [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/440fc376cc6895deb5a10358bb3799573e0a1827/WordPressAuthenticator/Unified%20Auth/View%20Related/Reusable%20Views/TextFieldTableViewCell.swift#L90). Not preparing the cell for reusue properly has introduced issues like https://github.com/woocommerce/woocommerce-ios/issues/7403 in the past.

# Testing steps

Please refer to the relating WCiOS PR to test the integration.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '2.2.1-beta.1'
+  s.version       = '2.2.0-beta.2'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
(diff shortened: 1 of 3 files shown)


itsmeichigo (approved):
LGTM 🚀

PR #666, opened 2022-08-16
Site address discovery

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/7468

### Description

This PR adds a new flow for troubleshooting a given site address. This flow reuses the current site address UI and returns the `WordPressSiteInfo` to a new delegate method for checking the problem of the site and show appropriate error message on the host app.

While testing the integration in https://github.com/woocommerce/woocommerce-ios/pull/7484, I found that for a non-existent site, the check is taking quite long to return an error - the library was checking for WordPress while we could check for its existence first. This PR adds that with a 10-second timeout and displays the error inline.

This PR also updates analytics by tracking a new flow `site_discovery`.

### Testing
Please check https://github.com/woocommerce/woocommerce-ios/pull/7484 for testing steps.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -90,6 +90,11 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     ///
     func sync(credentials: AuthenticatorCredentials, onCompletion: @escaping () -> Void)
 
+    /// Signals to the Host App that a WordPress site is available and needs validated.
+    /// This method is only triggered when the authenticator method `siteAddressUIForTroubleshooting` is called.
+    ///
+    func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController)
+
     /// Signals the Host App that a given Analytics Event has occurred.
     ///
     func track(event: WPAnalyticsStat)
@@ -101,4 +106,12 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     /// Signals the Host App that a given Analytics Event (with an associated Error) has occurred.
     ///
     func track(event: WPAnalyticsStat, error: Error)
+}
+
+/// Extension with default implementation for optional delegate methods.
+///
+extension WordPressAuthenticatorDelegate {
+    func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController) {
+        // No-op
+    }
 }
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -93,6 +93,10 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     /// Signals to the Host App that a WordPress site is available and needs validated.
     /// This method is only triggered in the site discovery flow.
     ///
+    /// - Parameters:
+    ///     - siteInfo: The fetched site information - can be nil the site doesn't exist or have WordPress
+    ///     - navigationController: the current navigation stack of the site discovery flow.
+    ///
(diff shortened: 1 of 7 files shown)


WordPressAuthenticator/Extensions/UIStoryboard+Helpers.swift
```diff
@@ -22,8 +22,8 @@ enum Storyboard: String {
     /// Returns a view controller from a Storyboard
     /// assuming the identifier is the same as the class name.
     ///
-    func instantiateViewController<T: NSObject>(ofClass classType: T.Type) -> T? {
+    func instantiateViewController<T: NSObject>(ofClass classType: T.Type, creator: ((NSCoder) -> UIViewController?)? = nil) -> T? {
         let identifier = classType.classNameWithoutNamespaces
-        return instance.instantiateViewController(withIdentifier: identifier) as? T
+        return instance.instantiateViewController(identifier: identifier, creator: creator) as? T
```
selanthiraiyan, line 27: TIL [this](https://developer.apple.com/documentation/uikit/uistoryboard/3213989-instantiateviewcontroller) 🙇

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
```diff
@@ -90,6 +90,11 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     ///
     func sync(credentials: AuthenticatorCredentials, onCompletion: @escaping () -> Void)
 
+    /// Signals to the Host App that a WordPress site is available and needs validated.
+    /// This method is only triggered in the site discovery flow.
+    ///
+    func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController?)
```
selanthiraiyan, line 100: Nit:

Kindly consider adding a comment about when `siteInfo` will be `nil`.

OR

Instead of having `siteInfo: WordPressComSiteInfo?` as an optional, what do you think about having another method for handling the case of no site info available scenario? i.e. `troubleshootSite` will act as success flow and the new method will act as failure flow of site discovery process.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -416,8 +434,42 @@ private extension SiteAddressViewController {
...
+
+                    return self.displayError(message: Localization.nonExistentSiteError, moveVoiceOverFocus: true)
+                }
+
+                // Proceeds to check for the site's WordPress
+                self.guessXMLRPCURL(for: self.loginFields.siteAddress)
+            }
+        }
+        task.resume()
+    }
```
selanthiraiyan, line 468: ❓ 
Should we raise a PR in the WP iOS repository and test this change as this code will execute even if `isSiteDiscovery` is false? 🤔 
Internal reference - p91TBi-49U-p2

selanthiraiyan (approved):
The changes look good to me. 
I have also tested the WCiOS PR, and it works as expected. 🎉 

I left a nit and a question.

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
itsmeichigo: Updated the comment in 032faba2260b95334d96b3b9597901075fc6a0fe. I'd prefer keeping the optional site info to make it simple to work with the delegate, and also match what we have in `shouldPresentUsernamePasswordController`.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
itsmeichigo: Makes sense! Added a PR here: https://github.com/wordpress-mobile/WordPress-iOS/pull/19209.

itsmeichigo:
Merging since https://github.com/wordpress-mobile/WordPress-iOS/pull/19209 tests passed

PR #668, opened 2022-08-18
Login: fix a crash after visiting site credentials screen and then auto-filling WP.com credentials

jaclync:
For https://github.com/woocommerce/woocommerce-ios/issues/7512

### Why

After visiting the site credentials screen and then auto-filling WP.com credentials on the email screen, the app crashes. It's because [`SiteCredentialsViewController` sets its `loginFields.meta.userIsDotCom = false` in `viewDidLoad`](https://href.li/?https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/8121ebd0cecc340f329538c4dad1f10eb92b5c3e/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteCredentialsViewController.swift#L46) and `loginFields` is a class (reference type), this `meta.userIsDotCom` is altered for all view controllers including `GetStartedViewController` and thus it goes to the self-hosted path when validating the fields in Objective-C that calls an optional delegate function without checking if it is implemented 😅

### The fix

- Instead of sharing the same `LoginFields` between `GetStartedViewController` and `SiteCredentialsViewController`, the former view controller now sets a copy of its `LoginFields`
- In the crashing Objective-C `LoginFacade.signInToSelfHosted`, it now checks if the delegate implements the optional method `displayLoginMessage` before calling (which triggers a crash without the check)

### Testing steps

Please test the steps in https://github.com/woocommerce/woocommerce-ios/pull/7514

### DIFF
--- a/WordPressAuthenticator/Model/LoginFields.swift
+++ b/WordPressAuthenticator/Model/LoginFields.swift
@@ -57,16 +57,18 @@ public class LoginFields: NSObject {
         return loginFields
     }
 
-    init(username: String = "",
-         password: String = "",
-         siteAddress: String = "",
-         multifactorCode: String = "",
-         nonceInfo: SocialLogin2FANonceInfo? = nil,
-         nonceUserID: Int = 0,
-         restrictToWPCom: Bool = false,
-         emailAddress: String = "",
-         meta: LoginFieldsMeta = LoginFieldsMeta(),
-         storedCredentials: SafariStoredCredentials? = nil) {
+    /// Using a convenience initializer for its Objective-C usage in unit tests.
+    convenience init(username: String,
+                     password: String,
+                     siteAddress: String,
+                     multifactorCode: String,
+                     nonceInfo: SocialLogin2FANonceInfo?,
+                     nonceUserID: Int,
+                     restrictToWPCom: Bool,
+                     emailAddress: String,
+                     meta: LoginFieldsMeta,
+                     storedCredentials: SafariStoredCredentials?) {
+        self.init()
         self.username = username
         self.password = password
         self.siteAddress = siteAddress
--- a/WordPressAuthenticator/Model/LoginFields.swift
+++ b/WordPressAuthenticator/Model/LoginFields.swift
@@ -99,7 +99,7 @@ extension LoginFields {
 
 /// A helper class for storing safari saved password information.
 ///
-struct SafariStoredCredentials {
+class SafariStoredCredentials {
     var storedUserameHash = 0
     var storedPasswordHash = 0
 }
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Model/LoginFields.swift
```diff
@@ -56,11 +56,50 @@ public class LoginFields: NSObject {
...
+              emailAddress: emailAddress,
+              meta: meta.copy(),
+              storedCredentials: storedCredentials)
+    }
 }
 
 /// A helper class for storing safari saved password information.
 ///
-class SafariStoredCredentials {
+struct SafariStoredCredentials {
```
itsmeichigo: Dumb question: why does this need to be a `struct`? And should we update the comment since it's no longer a class?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -752,7 +752,7 @@ private extension GetStartedViewController {
             return
         }
 
-        vc.loginFields = loginFields
+        vc.loginFields = loginFields.copy()
```
itsmeichigo, line 755: I see that there are a few other places in the library where we're sending the reference to the same login fields instead of sending the copy, should we be worried about them? It makes sense just to handle the one with the problem at the moment.

itsmeichigo (approved):
Thanks for the quick fix! It's a good practice to send copies instead of passing around references of the same object 👍

WordPressAuthenticator/Model/LoginFields.swift
jaclync: Great question! Originally, I thought it's better to use `Struct` so that a new struct is created when copying a `LoginFields` instead of copying by reference. However, I decided to revert this change since I don't understand the existing usage of `SafariStoredCredentials` enough. Reverted in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/668/commits/0a3bf419f07ebbd623e92ed9727211d549b611ea

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
jaclync: > I see that there are a few other places in the library where we're sending the reference to the same login fields instead of sending the copy, should we be worried about them?

Yes, we should probably be careful of this shared `LoginFields` behavior when navigating between screens and consider this behavior when looking into login issues. There are many mutations on `LoginFields` in various screens, but this PR is meant to just fix the specific crash/issue so I'll keep the copy just in this use case.

jaclync:
Thanks for the prompt review @itsmeichigo! I responded to your questions and updated the PR, merging now for code freeze.

PR #669, opened 2022-08-19
Site discovery: bypass app transport security when checking site existence

itsmeichigo:
### Description
In #666 I introduced checking site existence for site address login / site discovery. This involves creating a URLRequest to the site address.

There's an issue when the entered site address uses the HTTP scheme, in this case the request fails with `NSURLErrorAppTransportSecurityRequiresSecureConnection` if the host app doesn't allow HTTP in its Info.plist file, causes the existence check to fail.

This PR adds a fix to bypass if the existence check encounters the app transport security error.

### Testing
- Update the `WordPressAuthenticator` library in the host app to point to the latest commit in this branch.
- Navigate to the log in with site address flow in the host app.
- Enter a valid site address to a WP site that has HTTP scheme.
- Notice that the app navigates to the next screen of the flow without error.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '2.2.1-beta.4'
+  s.version       = '2.2.1-beta.5'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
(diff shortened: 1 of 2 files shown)


selanthiraiyan (approved):
Thanks for fixing this, Huong. 
It works as expected. 🎉 

On a side note, we should consider extracting the "site exists" check code into a private method with a completion handler, in a future PR. 🙏

itsmeichigo:
Thank you! It makes sense to extract the logic, updated in 1687e5573d611f98859aefa0439b3f48979dcbd0.

PR #670, opened 2022-08-19
Bump version to 2.3.0

oguzkocer:
In `x.y.z` version scheme, unless we are doing a hotfix, we want to increase the `y` value while creating a new beta/release. It looks like the current beta versions were incorrectly created for `2.2.1-beta.{version}` instead of `2.3.0-beta.{version}` which is minor mistake most of us have made.

While deciding on the final release version between `2.2.1` & `2.3.0`, I thought it was more important to use the correct final release version than to stay consistent with the beta versions and decided to go with `2.3.0`.

cc @jaclync @itsmeichigo (so you can keep this in mind for next beta versions)

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '2.2.1-beta.5'
+  s.version       = '2.3.0'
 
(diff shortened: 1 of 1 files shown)


mokagio (approved):

jaclync:
Sorry @oguzkocer about not using the beta version following the `x.y.z` version scheme! I'll make sure to follow this in future work on any of our libraries.

oguzkocer:
No worries at all @jaclync - my note was just a reminder 🙂

PR #674, opened 2022-08-26
Use a new `Step` value for Enter WPCOM email screen

selanthiraiyan:
Related to - https://github.com/woocommerce/woocommerce-ios/issues/7572 and https://github.com/woocommerce/woocommerce-ios/pull/7573

### Changes
`GetStartedViewController` and `SiteAddressViewController` were using same `Step` values (.start) 

This makes it impossible to differentiate between those two screens in analytics events. 

Meanwhile, WC Android uses ["enter_email_address"](https://github.com/woocommerce/woocommerce-android/blob/trunk/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/login/UnifiedLoginTracker.kt#L154) as `Step` value for Enter email screen (`GetStartedViewController`)

This PR introduces a new `Step` value "enter_email_address" and uses that for `GetStartedViewController`. 

### Testing instructions

Follow the instructions from https://github.com/woocommerce/woocommerce-ios/pull/7573 to test this.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '2.4.0-beta.1'
+  s.version       = '2.4.0-beta.2'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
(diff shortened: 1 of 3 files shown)


jaclync (approved):
LGTM ✅ 

`start` is indeed quite vague. does Android also use a different step value for the site address `SiteAddressViewController` screen? feel free to keep it as it is so that it doesn't affect existing analytics.

selanthiraiyan:
> does Android also use a different step value for the site address SiteAddressViewController screen? feel free to keep it as it is so that it doesn't affect existing analytics.

No, they use `start` in Android for this screen. i.e. The events are identical in iOS and Android for `SiteAddressViewController` screen.

PR #676, opened 2022-09-02
Add properties for prologue background image and prologue buttons blur effect

Gio2018:
This PR Adds two properties to the `WordPressStyles` that allow further customization of the prologue screen, specifically:

- add a custom background image
- add a custom blur effect to the buttons view

can be tested on the related WordPRess-iOS [PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/19277)

### DIFF
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -55,6 +55,9 @@ class LoginPrologueViewController: LoginViewController {
         }
 
         defaultButtonViewMargin = buttonViewLeadingConstraint?.constant ?? 0
+        if let backgroundImage = WordPressAuthenticator.shared.unifiedStyle?.prologueBackgroundImage {
+            view.layer.contents = backgroundImage.cgImage
+        }
     }
 
     override func styleBackground() {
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -55,6 +55,9 @@ class LoginPrologueViewController: LoginViewController {
         }
 
         defaultButtonViewMargin = buttonViewLeadingConstraint?.constant ?? 0
+        if let backgroundImage = WordPressAuthenticator.shared.unifiedStyle?.prologueBackgroundImage {
```
twstokes, line 58: 👋 Hey @Gio2018, would it make sense for this to live in the `styleBackground` function?

twstokes (commented):
Good work @Gio2018! I had one minor question in the code.

Outside of code, I have a visual observation that possibly we should run by @shaunandrews. I'm concerned the contrast ratio isn't very good and fails accessibility:



[These are the values](https://colourcontrast.cc/dbeac9/8fce6b) I picked out for the "Share on Facebook" text and its background, which fail for all font sizes. Do you have thoughts on that?

Gio2018:
> [These are the values](https://colourcontrast.cc/dbeac9/8fce6b) I picked out for the "Share on Facebook" text and its background, which fail for all font sizes. Do you have thoughts on that?

Background and font colors were picked from Figma: if we need to change those, we can do that. Since, as we observed earlier, the scrolling items are more of a decorative element, and, on the other hand, the project deadline is approaching, I'd suggest that we open a separate issue for that and change them when they are ready (maybe @shaunandrews wants to make some tests and have additional thoughts on them, and this wouldn't be blocking if we handle it that way)

twstokes:
> Background and font colors were picked from Figma: if we need to change those, we can do that. Since, as we observed earlier, the scrolling items are more of a decorative element

Agreed that we've determined it's decorative, so maybe it's a non-issue. Having said that, I noticed that VoiceOver is interactive with the scrolling words so we'll want to disable that.

> , and, on the other hand, the project deadline is approaching, I'd suggest that we open a separate issue for that and change them when they are ready 

That sounds like a plan to me.

twstokes (changes requested):
**Edit:** I realize I left these remarks on the wrong PR. They're more applicable to the [WPiOS side](https://github.com/wordpress-mobile/WordPress-iOS/pull/19277) of things. The code LGTM, good to merge once the branch conflict is resolved. I did leave one comment in the PR but it's optional. 👍 

👋 ~There are a couple things I think we should address:~
- ~There's a merge conflict with the PR.~
- ~The scrolling words are interactive with VoiceOver, so we should disable that if they have a decorative role.~

Thanks!

twstokes (approved):

PR #677, opened 2022-09-05
Bump version from 2.4.0 to 3.0.0 – 2.4.0 introduced a breaking change!

mokagio:
The change from e61570ca0b0afcaac496a7d6497b5471f0bb6f43 changed the `WordPressAuthenticatorDelegate` public
interface in a breaking way.

The breaking result can be seen in the [this WordPress iOS build](https://buildkite.com/automattic/wordpress-ios/builds/9808#01830b2b-76f3-4fe1-b385-52e09f417c3c/1745-8411), where we upgraded WordPressAuthenticator following Semantic Versioning and experienced a build failure:

```
WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift:228:1: type 'WordPressAuthenticationManager' does not conform to protocol 'WordPressAuthenticatorDelegate'
```

Given the circumstances, the library version ought to have gone from 2.3.0 to 3.0.0, rather than 2.4.0.

---

I'm leaving this as a draft because we have discussed SemVer in the past and I don't think we got to an agreement. I want to discuss this with @wordpress-mobile/apps-infrastructure internally to check my own assumptions before opening the conversation to the rest of the mobile dev groups.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '2.4.0'
+  s.version       = '3.0.0'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC


AliSoftware (approved):
I'm wondering if we should yank 2.4.0 from CocoaPods or at least [from our private podspec repo](https://github.com/wordpress-mobile/cocoapods-specs/tree/trunk/WordPressAuthenticator/2.4.0) to avoid clients using it if they use `~>2.x` version specifier in their Podfile  🤔  Or is it too late already and removing it would cause more harm than good?

mokagio:
>  Or is it too late already and removing it would cause more harm than good?

I'm afraid this is the case. It's [in use in WooCommerce](https://github.com/woocommerce/woocommerce-ios/blob/31487680fd88e5b392ed13f9b9a987bbfce71554/Podfile.lock#L45) and I worry several devs instance would break if/when running `pod install` before we get a chance to update to 3.0.

Given this is an internal library and that there'll be new versions soon, with no reason to stay on 2.x, I think it's acceptable to let the breaking change for WordPress out in the wild.

As a side note, the 2.4.0 usage never landed on WordPress's `trunk`.

PR #678, opened 2022-09-08
Show verify email screen from the authenticator

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/7597.

## Description
This PR adds a new method to `WordPressAuthenticator` to make it possible to present the verify email flow from the authenticator.

This is needed for when the user has set up Jetpack connection and needs to log in again to the app. The necessary input for this flow is the xmlrpc URL, authorized email address, and site address that the user is trying to log in to.

## Testing steps
Please follow the steps in https://github.com/woocommerce/woocommerce-ios/pull/7608 to test this PR.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -194,6 +194,29 @@ import WordPressKit
         showGetStarted(from: presenter, jetpackLogin: jetpackLogin, connectedEmail: connectedEmail, siteURL: siteURL)
     }
 
+    /// Used to present the Verify Email flow from the app delegate
+    ///
+    @objc public class func showVerifyEmailForWPCom(from presenter: UIViewController, xmlrpc: String, connectedEmail: String, siteURL: String) {
+        guard let xmlrpcURL = URL(string: xmlrpc) else {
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -194,6 +194,29 @@ import WordPressKit
         showGetStarted(from: presenter, jetpackLogin: jetpackLogin, connectedEmail: connectedEmail, siteURL: siteURL)
     }
 
+    /// Used to present the Verify Email flow from the app delegate
+    ///
+    @objc public class func showVerifyEmailForWPCom(from presenter: UIViewController, xmlrpc: String, connectedEmail: String, siteURL: String) {
```
selanthiraiyan, line 205: Super nit: It would be great to explain a bit about these parameters in the comments.

selanthiraiyan (approved):
LGTM :shipit:

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
itsmeichigo: Updated in c99f2a7d5f94f9f6e10ee16ae8879c826c4864f7.

PR #680, opened 2022-09-20
Site credential login entry point and site info fetching

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/7745.

## Description
This PR adds a new helper method to the authenticator so that the site credential login flow can be presented directly at the call site. 

Another helper method is for fetching site info, which triggers an existing API call in the library to get the essential information for a site given its URL.

## Testing
Please follow https://github.com/woocommerce/woocommerce-ios/pull/7748 for instructions to test the changes.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -242,6 +242,9 @@ import WordPressKit
         let loginFields = LoginFields()
         loginFields.siteAddress = siteURL
         controller.loginFields = loginFields
+        controller.dismissBlock = { _ in
+            controller.navigationController?.dismiss(animated: true)
+        }
 
         let navController = LoginNavigationController(rootViewController: controller)
         navController.modalPresentationStyle = .fullScreen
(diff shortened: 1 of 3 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -223,6 +223,45 @@ import WordPressKit
...
+        }
+
+        let navController = LoginNavigationController(rootViewController: controller)
+        navController.modalPresentationStyle = .fullScreen
+        presenter.present(navController, animated: true, completion: nil)
+    }
+
+    /// A helper method to fetch site info for a given URL.
+    ///
+    public class func fetchSiteInfo(for siteURL: String, onCompletion: @escaping (Result<WordPressComSiteInfo, Error>) -> Void) {
```
selanthiraiyan, line 259: Super nit: What do you think about adding comments to explain the parameters? 😁

selanthiraiyan (approved):
LGTM  :shipit:

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
itsmeichigo: Yeah - updated the comment in 6071628c95daba05b3593d9339bb3846aff8a8b4.

mokagio:
Merging this PR to ensure `trunk` is in sync with the latest published version.

The `publish-podspec` task failed in CI, but that had to do with the responses from CocoaPods' trunk rather than a build failure.

PR #683, opened 2022-09-24
Bump version to 3.2.0

spencertransier:


mokagio (approved):
👍 – Ready to merge. I'm not sure whether we have a guideline for which commit to tag with, in this case, 3.2.0.

The approach I usually take is to tag the commit in the PR / release branch, not the merge into `trunk`. The reason I do that is because it, to me, the merge commit is unrelated to the release process itself.

spencertransier:
@mokagio Agreed, I just added a tag to the version bump commit. 👍

PR #684, opened 2022-09-26
Respect showCancel state and show cancel button when appropriate

sla8c:
This PR fixes the issue where a user was unable to abandon the WP.com login flow while logged into a self hosted site by showing a Cancel button when showCancel = true is passed from caller.  Related to https://github.com/wordpress-mobile/WordPress-iOS/issues/19245

To test
Please test using the steps in the associated WPiOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/19349

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '3.2.0'
+  s.version       = '3.2.3-beta1'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '3.2.1-beta1'
+  s.version       = '3.3.0-beta.1'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '3.2.3-beta1'
+  s.version       = '3.2.1-beta1'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
@@ -177,6 +177,10 @@ import WordPressKit
             return nil
         }
 
+        if let loginNavController = controller as? LoginNavigationController, let loginPrologueViewController = loginNavController.viewControllers.first as? LoginPrologueViewController {
+            loginPrologueViewController.showCancel = showCancel
+        }
+
         controller.modalPresentationStyle = .fullScreen
         return controller
     }
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -163,6 +163,7 @@ class LoginPrologueViewController: LoginViewController {
         buttonViewController.shadowLayoutGuide = view.safeAreaLayoutGuide
         buttonViewController.topButtonStyle = WordPressAuthenticator.shared.style.prologuePrimaryButtonStyle
         buttonViewController.bottomButtonStyle = WordPressAuthenticator.shared.style.prologueSecondaryButtonStyle
+        buttonViewController.tertiaryButtonStyle = WordPressAuthenticator.shared.style.prologueSecondaryButtonStyle
     }
 
     /// Displays the old UI prologue buttons.


WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
```diff
@@ -177,6 +177,10 @@ import WordPressKit
             return nil
         }
 
+        if let loginNavController = controller as? LoginNavigationController, let loginPrologueViewController = loginNavController.viewControllers.first as? LoginPrologueViewController {
```
staskus, line 180: The `restrictToWPCom` and `onLoginButtonTapped` remains to be unused, I wonder if something else is broken because of this 🤔

There used to be such code and then removed in this PR https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/640: 

```
if let childController = controller.children.first as? LoginPrologueViewController {
        childController.loginFields.restrictToWPCom = restrictToWPCom
        childController.showCancel = showCancel
        childController.onLoginButtonTapped = onLoginButtonTapped
}
```

Maybe such solution needs to be restored?

staskus (commented):
Thanks for the fix! 👍

I found that similar code was removed previously (maybe by accident?). Also there're more fields passed to the method and then unused. I think we should pass them to the `LoginPrologueViewController` as well OR remove from method signature.

WordPressAuthenticator.podspec
```diff
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '3.2.0'
+  s.version       = '3.2.1-beta1'
```
momo-ozawa: [nit] I think we can bump the minor version instead of the micro version - wdyt?
```suggestion
  s.version       = '3.3.0-beta.1'
```

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
sla8c: Thanks for looking!  I wouldn't know how to update the PR steps to test these changes and we have been changing login flow because of Jetpack recently.  It might be worth exploring but not as part of this PR I think

guarani:
👋 Since @momo-ozawa and @staskus got to this well before me, I'll remove myself as a reviewer. Feel free to add me back if you'd like an extra set of eyes. And I'm available to help with future PRs, so don't hesitate to add me as a reviewer again going forward 👍

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
staskus: > I wouldn't know how to update the PR steps to test these changes and we have been changing login flow because of Jetpack recently.

Okay. Yes, it may require more investigation because as seen in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/640, these steps were removed together with the removal of Lottie which for me has little to do with this functionality.

> It might be worth exploring but not as part of this PR I think

Okay, it's our decision how we make these changes. The source of these bugs look to be the same but we could address them in different PRs.

We pass `restrictToWPCom: true` in `MeViewController`. As I understand, it means that we should only allow WordPress.com login if we log in through `MeViewController`. However, at the moment we also allow inputting the address of the self-hosted site, since we don't set `childController.loginFields.restrictToWPCom = restrictToWPCom`. I don't know if there's a valid scenario of logging into a self-hosted site while we're logged in a self-hosted site.

I can create additional issue if needed.

staskus, line 180: Just a suggestion, it looks it could be shortened based on previous implementation:

```suggestion
        let loginPrologueViewController = controller.children.first as? LoginPrologueViewController {
```

staskus (approved):
I tested and cancel button functionality and appearance works as described both on WordPress and Jetpack apps 👍 .

I made a comment about additional issues that could be addressed separately.

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
sla8c: @mokagio just wanted to tag you in here.  It seems like I'm adding back functionality that was removed as part of this change [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/640/files#diff-8b77840eb74486edc937048220e6c0c491d2789095582e5a3029c1406a248e20L179).  Would you mind adding some context on why this might have been removed as part of removing Lottie?  Thanks!

mokagio: @sla8c 

> Would you mind adding some context on why this might have been removed as part of removing Lottie? 

I'm mostly sure that was an oversight. Or rather, a case of not having enough automation to know when removing code  that looks dead from the library point of view (like in the case of [that check](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/640/files#diff-8b77840eb74486edc937048220e6c0c491d2789095582e5a3029c1406a248e20L179)) might result in unexpected behavior downstream.

For additional context, I looked at the Git history and found out that:

1. The whole of `LoginPrologueViewController` was removed [in the first phase of removing Lottie](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/640/commits/89bb95976566faa0baf291e842f7cc733d840698)
2. That, of course, resulted in that check for the first child view controller being a `LoginPrologueViewController` being deleted as well
3. While the library built that way, the Woo UI tests failed, which prompted me to [reintroduce the type](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/640/commits/cf675a96481a79ce6413e1780e89af9c5dd96bbe)
4. It didn't occur to me to reintroduce the check as well, and because there was no test that failed, the whole thing went under the radar until you started working on it.

I don't remember the rational for removing `LoginPrologueViewController` in the first place, but my guess would be that it was thought to be part of a Lottie-driven set of UI components that should no longer have been used. Just a guess, though, based on the other deleted files. I could be off the mark and haven't dug deeper in the code and history.

@sla8c @staskus from the comments above, it looks like you are already on top of this, but I'd stress that it would be very useful to have an automated test to ensure the behavior restored with this PR doesn't degenerate again in the future. I guess given the issue was in the Jetpack/WordPress codebase, the test could live there, at least until we get to adding a demo app to this project. I'm sure @wordpress-mobile/mobile-ui-testing-squad will be happy to help with that.

mokagio:
I'm taking over merging this approved bug fix PR so that it's [approved consumer in Jetpack/WordPress](https://github.com/wordpress-mobile/WordPress-iOS/pull/19349) can be part of 20.9.

WordPressAuthenticator/Authenticator/WordPressAuthenticator.swift
sla8c: @mokagio thanks for the context!  @wordpress-mobile/mobile-ui-testing-squad please let me know if you need anything (a GH issue created) or anything else re: getting a UI test for this on your radar thanks 🙂

mokagio: @sla8c if you have the time, it would be super helpful if you could have a crack at writing the test. 

The infrastructure @wordpress-mobile/mobile-ui-testing-squad is building is meant to empower developers to write tests across the entire low-level/unit-test to high-level/UI-test spectrum. It's always helpful to get feedback on the state of the tooling and the codebase itself from the "users" we're building things for.

PR #686, opened 2022-10-04
Bash cache & `UIPasteBoard` tests updates

mokagio:
This PR started as one to upgrade the project to build using the `xcode-14` image in CI. However I didn't iterate on it fast enough and other PRs (https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/695, https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/699) did most of the work. So, this PR now only tracks minor improvements.

### DIFF
--- a/WordPressAuthenticatorTests/Authenticator/PasteboardTests.swift
+++ b/WordPressAuthenticatorTests/Authenticator/PasteboardTests.swift
@@ -10,13 +10,17 @@ class PasteboardTests: XCTestCase {
     }
 
     func testNominalAuthCode() throws {
-        if #available(iOS 16.0, *) {
-            throw XCTSkip("UIPasteboard doesn't work in iOS 16.0.") // Check https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/696
-        }
-
         guard #available(iOS 14.0, *) else {
             throw XCTSkip("Unsupported iOS version")
         }
+
+        // FIXME: We'll need to find a way to make the test work the new pasteboard rules
+        //
+        // See:
+        // - https://developer.apple.com/forums/thread/713770
+        // - https://sarunw.com/posts/uipasteboard-privacy-change-ios16/
+        // - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/696
+        XCTExpectFailure("Paste board access has changed in iOS 16 and this test is now failing")
 
         let expect = expectation(description: "Could read nominal auth code from pasteboard")
         let pasteboard = UIPasteboard.general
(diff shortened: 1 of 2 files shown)


WordPressAuthenticatorTests/Authenticator/PasteboardTests.swift
```diff
@@ -10,52 +10,62 @@ class PasteboardTests: XCTestCase {
...
         }
 
+        // FIXME: We'll need to find a way to make the test work the new pasteboard rules
+        //
+        // See:
+        // - https://developer.apple.com/forums/thread/713770
+        // - https://sarunw.com/posts/uipasteboard-privacy-change-ios16/
+        // - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/696
+        XCTExpectFailure("Paste board access has changed in iOS 16 and this test is now failing")
+
```
AliSoftware, line 24: What if we happen to run the Unit Tests using an iPhone Simulator that runs iOS 15.4 or older? Shouldn't we guard this against the iOS runtime version?

Probably with something like [this](https://developer.apple.com/documentation/foundation/processinfo/1414876-isoperatingsystematleast):
```suggestion
        // FIXME: We'll need to find a way to make the test work the new pasteboard rules
        //
        // See:
        // - https://developer.apple.com/forums/thread/713770
        // - https://sarunw.com/posts/uipasteboard-privacy-change-ios16/
        // - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/696
        if ProcessInfo.processInfo.isOperatingSystemAtLeast(.init(majorVersion: 16, minorVersion: 0, patchVersion: 0)) {
            XCTExpectFailure("Pasteboard access has changed in iOS 16 and this test is now failing")
        }
```

AliSoftware: Confirmed this is needed, by testing on an iOS 15.4 simulator 😉 
image

AliSoftware (changes requested):

WordPressAuthenticatorTests/Authenticator/PasteboardTests.swift
mokagio: Neat technique.

But do we _really_ need it? Yes, the tests will fail on iOS < 16, but how much of a concern is that for us? What are the chances we'll need to run the tests in iOS < 16?

Thinking about it, it might be beneficial _if_ the `UIPasteboard` changes are such that we'll need different code between iOS < 16 and >= 16. 🤔 I haven't looked at that yet, this might be a good time to do so.

mokagio:
As of [2025/02/06](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/cce14d7ea5e5af5e52cdcfa75cbec672c1603991) this repo is a public archive and no longer under develpoment.

Closing this PR.

Looks like neither WordPress/Jetpack nor WooCommerce have developed the tests further. It might be good to revisit this and see if it would beneficial to port over to those apps. Tracked in https://github.com/Automattic/apps-infra-plans/issues/169

PR #689, opened 2022-10-16
Remove Alamofire from dependencies

crazytonyli:
It's no used in this library.

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - Alamofire (4.8.0)
+  - Alamofire (4.8.2)
   - AppAuth (1.4.0):
     - AppAuth/Core (= 1.4.0)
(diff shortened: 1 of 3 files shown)


Podfile.lock
```diff
@@ -1,5 +1,5 @@
 PODS:
-  - Alamofire (4.8.0)
+  - Alamofire (4.8.2)
```
mokagio, line 2: I wasn't expecting this to change version after it being removed from the `Podfile` and `podspec` 🤔

mokagio: Ah! Nevermind, it's a dependency at [the WordPressKit level](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/79ebd209c439a3947b9eed788050a9f414ce87c0/WordPressKit.podspec#L25).

mokagio (approved):

PR #691, opened 2022-10-25
New configs and updates for simplified login flow i1

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/7899, https://github.com/woocommerce/woocommerce-ios/issues/7900, and https://github.com/woocommerce/woocommerce-ios/issues/7901.
## Description
This PR adds a few updates to support the simplified login flow in the Woo app. We planned to duplicate the screens, but it seems to take so much more time, so I went for conditional checks instead.
- New config `enableSimplifiedLoginI1` to update the prologue, get started, and password screens to not mention the WP.com account during login.
- New config `enableSiteCreationForSimplifiedLoginI1` to show the secondary button in the prologue screen for simplified login.
- New delegate method `showSiteCreation` to ask the host app to navigate to site creation when the site creation button on the prologue screen is tapped.
- Prologue screen: new buttons for simplified login.
- Get Started screen: Pin the Continue button to the bottom for simplified login
- Password screen: show instruction text for WP.com password in simplified login (I'm keeping the screen for WP.org password as-is).

## Testing
Please follow the PR https://github.com/woocommerce/woocommerce-ios/pull/7929 for instructions to test the changes here.

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -105,6 +105,15 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     ///
     func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController?)
 
+    /// Signals to the Host App to navigate to the account creation flow.
+    /// This method is currently used only in the simplified login flow
+    /// when the configuration `enableSimplifiedLoginI1` is enabled
+    ///
+    /// - Parameters:
+    ///     - navigationController: the current navigation stack of the login flow.
+    ///
+    func showAccountCreation(in navigationController: UINavigationController?)
+
     /// Signals the Host App that a given Analytics Event has occurred.
     ///
     func track(event: WPAnalyticsStat)
@@ -122,6 +131,10 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
 ///
 public extension WordPressAuthenticatorDelegate {
     func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController?) {
+        // No-op
+    }
+
+    func showAccountCreation(in navigationController: UINavigationController?) {
         // No-op
     }
 }
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -105,14 +105,14 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     ///
     func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController?)
 
-    /// Signals to the Host App to navigate to the account creation flow.
+    /// Signals to the Host App to navigate to the site creation flow.
     /// This method is currently used only in the simplified login flow
-    /// when the configuration `enableSimplifiedLoginI1` is enabled
+    /// when the configs `enableSimplifiedLoginI1` and `enableSiteCreationForSimplifiedLoginI1` is enabled
     ///
     /// - Parameters:
     ///     - navigationController: the current navigation stack of the login flow.
     ///
-    func showAccountCreation(in navigationController: UINavigationController?)
+    func showSiteCreation(in navigationController: UINavigationController?)
 
     /// Signals the Host App that a given Analytics Event has occurred.
     ///
@@ -134,7 +134,7 @@ public extension WordPressAuthenticatorDelegate {
         // No-op
     }
 
-    func showAccountCreation(in navigationController: UINavigationController?) {
+    func showSiteCreation(in navigationController: UINavigationController?) {
         // No-op
     }
 }
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
@@ -112,7 +112,7 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
     /// - Parameters:
     ///     - navigationController: the current navigation stack of the login flow.
     ///
-    func showSiteCreation(in navigationController: UINavigationController?)
+    func showSiteCreation(in navigationController: UINavigationController)
 
     /// Signals the Host App that a given Analytics Event has occurred.
     ///
@@ -134,7 +134,7 @@ public extension WordPressAuthenticatorDelegate {
         // No-op
     }
 
-    func showSiteCreation(in navigationController: UINavigationController?) {
+    func showSiteCreation(in navigationController: UINavigationController) {
         // No-op
     }
 }
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
@@ -22,6 +22,7 @@ public struct WordPressAuthenticatorDisplayStrings {
     public let googleSignupInstructions: String
     public let googlePasswordInstructions: String
     public let applePasswordInstructions: String
+    public let wpcomPasswordInstructions: String
 
     /// Strings: primary call-to-action button titles.
     ///
@@ -80,7 +81,8 @@ public struct WordPressAuthenticatorDisplayStrings {
                 googleSignupInstructions: String = defaultStrings.googleSignupInstructions,
                 googlePasswordInstructions: String = defaultStrings.googlePasswordInstructions,
                 applePasswordInstructions: String = defaultStrings.applePasswordInstructions,
-                continueButtonTitle: String = defaultStrings.continueButtonTitle,
+                wpcomPasswordInstructions: String = defaultStrings.wpcomPasswordInstructions,
+                continueButtonTitle: String =  defaultStrings.continueButtonTitle,
                 magicLinkButtonTitle: String = defaultStrings.magicLinkButtonTitle,
                 openMailButtonTitle: String = defaultStrings.openMailButtonTitle,
                 createAccountButtonTitle: String = defaultStrings.createAccountButtonTitle,
@@ -123,6 +125,7 @@ public struct WordPressAuthenticatorDisplayStrings {
         self.googleSignupInstructions = googleSignupInstructions
         self.googlePasswordInstructions = googlePasswordInstructions
         self.applePasswordInstructions = applePasswordInstructions
+        self.wpcomPasswordInstructions = wpcomPasswordInstructions
         self.continueButtonTitle = continueButtonTitle
         self.magicLinkButtonTitle = magicLinkButtonTitle
         self.openMailButtonTitle = openMailButtonTitle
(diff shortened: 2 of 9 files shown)


WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
```diff
@@ -19,6 +19,12 @@ class PasswordViewController: LoginViewController {
 
     private let isMagicLinkShownAsSecondaryAction: Bool = WordPressAuthenticator.shared.configuration.isWPComMagicLinkShownAsSecondaryActionOnPasswordScreen
 
+    private let configurations = WordPressAuthenticator.shared.configuration
```
jaclync: nit: maybe `configuration` is more consistent with the struct name and other usages?
```suggestion
    private let configuration = WordPressAuthenticator.shared.configuration
```

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
```diff
@@ -105,6 +105,15 @@ public protocol WordPressAuthenticatorDelegate: AnyObject {
...
     func troubleshootSite(_ siteInfo: WordPressComSiteInfo?, in navigationController: UINavigationController?)
 
+    /// Signals to the Host App to navigate to the site creation flow.
+    /// This method is currently used only in the simplified login flow
+    /// when the configs `enableSimplifiedLoginI1` and `enableSiteCreationForSimplifiedLoginI1` is enabled
+    ///
+    /// - Parameters:
+    ///     - navigationController: the current navigation stack of the login flow.
+    ///
+    func showSiteCreation(in navigationController: UINavigationController?)
```
jaclync: can we update `navigationController` to be non-nil? otherwise, the host app might not know how to present the site creation flow when the `navigationController` is nil.

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -158,7 +158,11 @@ class LoginPrologueViewController: LoginViewController {
             return
         }
 
-        buildUnifiedPrologueButtons(buttonViewController)
+        if configuration.enableSimplifiedLoginI1 {
```
jaclync: nit: instead of a general `enableSimplifiedLoginI1` that is used in various screens here and there, how about calling the configuration `enableOneAuthButtonInPrologue` to only show one login/signup button on the prologue screen?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -128,8 +128,11 @@ class GetStartedViewController: LoginViewController, NUXKeyboardResponder {
         loadRows()
         setupTableFooterView()
         configureDivider()
+
         if screenMode == .signInUsingSiteCredentials {
             configureButtonViewControllerForSiteCredentialsMode()
+        } else if configuration.enableSimplifiedLoginI1 {
```
jaclync: nit: similar to before - instead of a general `enableSimplifiedLoginI1` that is used in various screens here and there, how about calling the configuration here `enableSocialLogin` for whether to show the social buttons or not?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -157,7 +160,8 @@ class GetStartedViewController: LoginViewController, NUXKeyboardResponder {
         hiddenPasswordField?.text = nil
         hiddenPasswordField?.isAccessibilityElement = false
 
-        if screenMode == .signInUsingSiteCredentials {
+        if screenMode == .signInUsingSiteCredentials ||
+            configuration.enableSimplifiedLoginI1 {
```
jaclync: I saw a few cases of these combinational checks on this screen, and was struggling to follow all the conditions. I'm guessing there are two main conditions for the UI from my current understanding:
- Whether the social buttons are shown
- Where the Continue button is placed - below the text field, at the bottom of the screen, etc.

How about having a function for each check based on the configuration and screen mode?

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
```diff
@@ -19,6 +19,12 @@ class PasswordViewController: LoginViewController {
...
+
+    private var isSiteCredentialPassword: Bool {
+        loginFields.siteAddress.isEmpty == false && configurations.enableSiteCredentialsLoginForSelfHostedSites
```
jaclync: unrelated to this PR: might be nice having a screen mode enum since this screen is used for various use cases and it feels pretty fragile with conditional checks here and there 😅

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
```diff
@@ -338,8 +344,9 @@ private extension PasswordViewController {
     func loadRows() {
         rows = [.gravatarEmail]
 
-        // Instructions only for social accounts
-        if loginFields.meta.socialService != nil {
+        // Instructions only for social accounts and simplified WPCom login flow
+        if loginFields.meta.socialService != nil ||
+            (configurations.enableSimplifiedLoginI1 && isSiteCredentialPassword == false) {
```
jaclync: nit: similar to before - instead of a general `enableSimplifiedLoginI1` that is used in various screens here and there, how about moving the `WordPressAuthenticatorDisplayStrings.wpcomPasswordInstructions` to `WordPressAuthenticatorConfiguration.wpcomPasswordInstructions: String?` with an optional value that's shown only when it's non-nil?

jaclync (approved):
Works as expected! 👍 also tried enabling the `enableSiteCreationForSimplifiedLoginI1` configuration and it looks nice!

Since I just looked at the design changes for the simplified login for the first time, it was taking me some time to understand all the changes on different screens. In the comments, I suggested breaking the general `WordPressAuthenticatorConfiguration.enableSimplifiedLoginI1` into separate configurations for readability since this is very specific to the project, and developers outside of Woo mobile wouldn't be able to see the exact changes when checking the code. Also, if we decide to keep only a subset of changes on different screens, having separate configurations would make it easier (just remove the configurations that we're not going with). But feel free to proceed with just one configuration if we plan to remove it in the near future.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
itsmeichigo: Updated in b014e42e11cd53d1e7ec06c888047db95d8c1f24 👍

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift
itsmeichigo: This makes sense - updated in 13242b7

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
itsmeichigo: Good point! It is better to make the config generic - I updated this one and others in 7a7aebe.

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
itsmeichigo: Also updated in 7a7aebe541eb9beaa99e1505777b55e0ac9511d2.

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
itsmeichigo: Also updated in 7a7aebe541eb9beaa99e1505777b55e0ac9511d2.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
itsmeichigo: This is no longer used, so I've removed it in 7a7aebe541eb9beaa99e1505777b55e0ac9511d2.

WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
itsmeichigo: Also updated in 7a7aebe541eb9beaa99e1505777b55e0ac9511d2.

itsmeichigo:
Thank you for the great reviews @jaclync! 

I updated the configs to be more generic so that other apps can use them if necessary. 

I also added some more changes to include the border around the email and avatar on the password screen following the final decision on the design. Please feel free to test again in https://github.com/woocommerce/woocommerce-ios/pull/7929.

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -466,11 +473,11 @@ private extension GetStartedViewController {
     /// Configures appearance of the submit button.
     ///
     func configureContinueButton(animating: Bool) {
-        switch screenMode {
-        case .signInUsingSiteCredentials:
+        if screenMode == .signInUsingSiteCredentials ||
+            configuration.enableSocialLogin == false {
```
jaclync: nit: can we use `showsContinueButtonAtTheBottom` here?

WordPressAuthenticator/Authenticator/WordPressAuthenticatorDisplayStrings.swift
```diff
@@ -79,7 +80,7 @@ public struct WordPressAuthenticatorDisplayStrings {
                 googleSignupInstructions: String = defaultStrings.googleSignupInstructions,
                 googlePasswordInstructions: String = defaultStrings.googlePasswordInstructions,
                 applePasswordInstructions: String = defaultStrings.applePasswordInstructions,
-                continueButtonTitle: String = defaultStrings.continueButtonTitle,
+                continueButtonTitle: String =  defaultStrings.continueButtonTitle,
```
jaclync: super nit: empty space on an unchanged line
```suggestion
                continueButtonTitle: String = defaultStrings.continueButtonTitle,
```

WordPressAuthenticator/Unified Auth/View Related/Reusable Views/GravatarEmailTableViewCell.swift
```diff
@@ -33,6 +44,17 @@ class GravatarEmailTableViewCell: UITableViewCell {
         }
 
         gravatarImageView?.downloadGravatarWithEmail(email, placeholderImage: placeholderImage ?? gridicon)
+
+        gravatarImageViewSizeConstraints.forEach { constraints in
```
jaclync: super nit:
```suggestion
        gravatarImageViewSizeConstraints.forEach { constraint in
```

WordPressAuthenticator/Unified Auth/View Related/Reusable Views/GravatarEmailTableViewCell.swift
```diff
@@ -8,23 +8,34 @@ class GravatarEmailTableViewCell: UITableViewCell {
...
 
     /// Public properties
     ///
     public static let reuseIdentifier = "GravatarEmailTableViewCell"
     public var onChangeSelectionHandler: ((_ sender: UITextField) -> Void)?
 
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        containerView.layer.cornerRadius = 8
+        containerView.layer.borderColor = UIColor.systemGray3.cgColor
```
jaclync: just checking, do we need these setups when `hasBorders` is `false` as well?

jaclync (approved):
Thank you for the updates, the changes look nice! 💟

WordPressAuthenticator/Unified Auth/View Related/Reusable Views/GravatarEmailTableViewCell.swift
itsmeichigo: I think this can be set up once, and we'd only need to update the border width to show/hide the border. But I added an update in 3c70e8845b2945c7e342f6b53cfda335ed63d52a to configure the container view in one place to make it easier to read (performance shouldn't be a concern here since we have only one row of this type in the table view).

itsmeichigo:
Thank you for the 2nd check, I've added the changes following your suggestions in the subsequent commits. Merging this now.

PR #692, opened 2022-10-28
Allow the host app to pass a custom source identifier to the login flow

jaclync:
⚠️ Please note that this PR includes a breaking change to the `WordPressAuthenticatorDelegate` protocol and `NavigateToEnterAccount` initializer, any host app that uses these needs to update its implementation.

## Why

For store creation in WCiOS, the user could enter the store creation flow but choose to log in instead of signing up for an account in the app. Since the login flow is completely handled in the WPAuthententicator library, starting either from the login prologue or via `NavigateToEnterAccount`, we want to pass a source identifier in order for the app to continue the store creation flow instead of the usual login completion path (store picker).

## Changes

An optional `SignInSource` was added to `WordPressAuthenticatorDelegate.presentLoginEpilogue`. In order for the source to be accessed in various places that can call `WordPressAuthenticatorDelegate.presentLoginEpilogue`, I decided to add a variable to the global singleton `WordPressAuthenticator` (it's not the cleanest, but if we don't do this we have to DI this source in so many classes with potential misses). When `GetStartedViewController.source` is set, it also sets `WordPressAuthenticator.signInSource`. This value in the singleton is then used in call sites of `WordPressAuthenticatorDelegate.presentLoginEpilogue` to pass the source to the host app.

## Testing steps

Please check out the WCiOS PR.

### DIFF
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressAuthenticator'
-  s.version       = '3.3.0-beta.1'
+  s.version       = '3.3.0-beta.2'
 
   s.summary       = 'WordPressAuthenticator implements an easy and elegant way to authenticate your WordPress Apps.'
   s.description   = <<-DESC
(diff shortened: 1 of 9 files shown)


selanthiraiyan (approved):
LGTM 🚢

PR #697, opened 2022-11-02
Setup SwiftLint to run locally

mokagio:
Adds SwiftLint to the project, integrates it within Xcode as a build phase, and addresses existing violations.

If you checkout the project locally and build, you should see no warnings.

_CI integration will come later on. Weighing whether to setup Danger or a step that only runs SwiftLint._

---


WordPressAuthenticator/Unified Auth/View Related/Password/PasswordViewController.swift
```diff
@@ -145,7 +145,9 @@ class PasswordViewController: LoginViewController {
             // is the actual error string. So check for "password" in the error string, and show the custom
             // error message. Otherwise, show the actual response error.
             var displayMessage: String {
+                // swiftlint:disable localization_comment
                 if nsError.localizedDescription.contains(NSLocalizedString("password", comment: "")) {
+                // swiftlint:enable localization_comment
```
mokagio, line 150: I didn't see the point in adding a comment explaining the intricate implementation details of why this localized string exists. It wouldn't be useful to translators.

I think we can all agree that error checking based on grepping a string is suboptimal but addressing it is out of scope for this PR.

WordPressAuthenticator/Signin/LoginWPComViewController.swift
```diff
@@ -142,7 +142,10 @@ class LoginWPComViewController: LoginViewController, NUXKeyboardResponder {
                 return NSLocalizedString("To proceed with this Google account, please first log in with your WordPress.com password. This will only be asked once.", comment: "")
             }
 
-            return NSLocalizedString("Please enter the password for your WordPress.com account to log in with your Apple ID.", comment: "")
+            return NSLocalizedString(
+                "Please enter the password for your WordPress.com account to log in with your Apple ID.",
+                comment: "Instructional text shown when requesting the user's password for a login initiated via Sign In with Apple"
```
mokagio, line 147: Thank you SwiftLint for picking this up 😄 

I did my best to add a comment consistent with the other ones in the file.

Podfile
```diff
@@ -49,17 +49,29 @@ target 'WordPressAuthenticatorTests' do
...
 
-  post_install do |installer|
-    # Let Pods targets inherit deployment target from the app
-    # This solution is suggested here: https://github.com/CocoaPods/CocoaPods/issues/4859
-    # =====================================
-    #
-    installer.pods_project.targets.each do |target|
-      target.build_configurations.each do |configuration|
-        pod_ios_deployment_target = Gem::Version.new(configuration.build_settings['IPHONEOS_DEPLOYMENT_TARGET'])
-        configuration.build_settings.delete 'IPHONEOS_DEPLOYMENT_TARGET' if pod_ios_deployment_target <= ios_deployment_target
```
mokagio, line 61: The indentation change here is because I extracted the hook to the global context (I guess you could call it that? _Outside of the `target` block where it was before_).

Makefile
```diff
@@ -0,0 +1,14 @@
...
+# and with system Ruby where Bundler and Fastlane may or may not be available.
+# `make` on the other hand, is always available, therefore `make lint` can be
+# reliably called from both the user's terminal, Xcode, and CI.
+SWIFTLINT=./Pods/SwiftLint/swiftlint
+
+lint:
+	$(SWIFTLINT) lint --quiet
+
+format:
+	$(SWIFTLINT) lint --autocorrect --quiet
```
mokagio, line 14: The approach presented here might be a bit controversial, but is working well for the Pocket Casts iOS devs.

I don't like the overhead of having `Makefile` and `Fastfile` but has one positive aspect. Our long term vision is to remove the need to deal with Ruby for feature devs, but they might still need to run CLI tasks such as these and `make` is an excellent solution for that. Introducing `make` now could pave the way for leaning more into it in the future.

Podfile
```diff
@@ -49,17 +49,29 @@ target 'WordPressAuthenticatorTests' do
...
-      end
+      # This addresses Xcode 12, 13, and 14 showing an "Update to recommended
+      # settings" warning on the Pods project.
+      #
+      # See:
+      #
+      # - https://github.com/CocoaPods/CocoaPods/issues/10189
+      # - https://github.com/CocoaPods/CocoaPods/issues/11553
+      configuration.build_settings.delete 'ARCHS'
+      configuration.build_settings['DEAD_CODE_STRIPPING'] = 'YES'
```
mokagio, line 75: This is the warning I'm referring to, you would have seen it many times in our various apps:

image

Podfile
```diff
@@ -49,17 +49,29 @@ target 'WordPressAuthenticatorTests' do
   pod 'OCMock', '~> 3.4'
   pod 'Expecta', '1.0.6'
   pod 'Specta', '1.0.7'
+end
+
+abstract_target 'CI' do
+  pod 'SwiftLint', '~> 0.49'
+end
```
crazytonyli, line 57: Hmm, what is this black magic? Would adding this abstract target alone make cocoapods to install SwiftLint during CI checks?

crazytonyli: Just tried `pod install` locally. So, what this abstract target does is installing a SwiftLint pod but no using it in any targets, isn't it?

> [!] The abstract target CI is not inherited by a concrete target, so the following dependencies won't make it into any targets in your project:
    - SwiftLint (~> 0.49)

crazytonyli (approved):

Podfile
mokagio: > So, what this abstract target does is installing a SwiftLint pod but no using it in any targets, isn't it?

Oh! I see now that calling the target "CI" might be a bit confusing, in particular when I wrote in the PR description that this doesn't interact with CI 🤦‍♂️ 

> what this abstract target does is installing a SwiftLint pod but no using it in any targets, isn't it?

Yes. It's a bit of a misuses of [`abstract_target`](https://guides.cocoapods.org/syntax/podfile.html#abstract_target) which is meant for "convenient target dependency inheritance", but it does the job of downloading the binary at the specified version without the need of explicitly adding it to any of the targets because we don't need it at run time, just as a tool during build time.

mokagio: I'll rename "CI" to "Tools", which might make this a bit clearer.

PR #698, opened 2022-11-03
Demo app walking skeleton

mokagio:
This PR introduces a [walking skeleton](https://wiki.c2.com/?WalkingSkeleton) implementation of the demo app. Its purpose is to have the demo app building both locally and in CI while importing the framework. I'll be fleshing it out later on properly integrating the framework.

I decided to use a dedicated project for the demo app to reduce the noise in the diff for the framework's project file. This is also consistent with what we'd do in an SPM setup, where the demo app would necessarily have its own project, given the package won't have one (other than the one that Xcode generates in `.swiftpm` if you open the folder containing the `Package.swift` file with it, but that's not tracked in Git).

Next steps:

- #702 
- Add Google Sign In to demo app with existing logic
- Use demo app to implement SDK-less Google Sign In

---


Demo/AuthenticatorDemo.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,361 @@
+// !$*UTF8*$!
```
mokagio, line 1: I decided to use a dedicated project instead of making the demo app a new target in the project where we define WordPressAuthenticator to avoid noise in that project file.

Demo/AuthenticatorDemo/ViewController.swift
```diff
@@ -0,0 +1,10 @@
+import UIKit
+import WordPressAuthenticator
+
+class ViewController: UIViewController {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        print("Error domain is \(WordPressAuthenticator.errorDomain)")
```
mokagio, line 8: Just a simple `print` of a value coming from the framework to make sure everything is wired up alright.

Podfile
```diff
@@ -51,6 +55,12 @@ target 'WordPressAuthenticatorTests' do
   pod 'Specta', '1.0.7'
 end
 
+target 'AuthenticatorDemo' do
+  project 'Demo/AuthenticatorDemo.xcodeproj'
+
+  pod 'WordPressAuthenticator', path: '.'
+end
+
```
mokagio, line 63: I'm not too sure about the quality of this CocoaPods-based setup. But, I'm also not keen in spending too much time on it in light of our current parallel effort to adopt to SPM.

From my point of view, this setup works and lets me work on the demo app without hassle. I think it'll serve us well till the time we'll migrate to SPM.

_Note:_ I did consider migrating to SPM before starting this, but decided against it because of the various dependencies. One of the goal that adding the demo app aims to help achieve is removing GoogleSingIn and I didn't want to spend extra time to figure out how to make it work with SPM only to then remove it.

crazytonyli (approved):

PR #699, opened 2022-11-04
Bump version to 4.1.1

oguzkocer:
This is my first release after the `CHANGELOG.md` introduction, so hopefully it's done correctly 🤞 

Note that the initial commit failed with `VM xcode-13.4.1 is invalid` so I updated the xcode image to 14 which is the image that is in use in WCiOS.

**Question:** Does the `.xcversion` file need to be updated to `14.0` as well? I am guessing yes, but I am not familiar with the xcode version updates, so I'd appreciate some help.

**Update**
---
I had a few CI failures during publishing of the `4.1.0` version and ended up having to bump the version to `4.1.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ _None._
 
 -->
 
-## 4.1.0-beta.1 [Unreleased]
+## 4.2.0 [Unreleased]
 
 ### Breaking Changes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -30,23 +30,33 @@ _None._
 
 -->
 
-## 4.1.0-beta.1 [Unreleased]
+## 4.2.0 [Unreleased]
```
mokagio, line 33: The following is out of scope from this PR, because it correctly update the code based on its existing state, and also a big nitpick.

IMHO, this header should always stay "Unreleased" because it's up to the release manager (eventually the automation) to parse what went into the changelog since the latest release and decide what number to use based on the kind of changes.

I appreciate the idea of suggesting a new version number but I feel it would be neater if the header for unreleased code stayed fixed.

I might open a PR later on to update the template to make this clearer and we can discuss then.

mokagio (approved):
> This is my first release after the `CHANGELOG.md` introduction, so hopefully it's done correctly 🤞

LGTM! 👏 😄 

> Question: Does the `.xcversion` file need to be updated to 14.0 as well? I am guessing yes, but I am not familiar with the xcode version updates, so I'd appreciate some help.

This question comes at a very good time, as @AliSoftware suggested automation changes to fix an issue and overall improved the whole process [in WooCommerce iOS](https://github.com/woocommerce/woocommerce-ios/pull/8033).

If that's okay with you, I'll take this from here. I've suggested some minor changes in the Woo setup and I'd be happy to replicated the setup we'll land with from there in this repo (I'm also actively working on this repo, so it's no extra effort for me).

CHANGELOG.md
AliSoftware: Agreed. Predicted next version number will vary as new items will get added to the unreleased section, which is why we don't include a version number in it until we actually do the next release and decide on it based on content of the section and semver rules

PR #700, opened 2022-11-07
Config to check XMLRPC only when needed

selanthiraiyan:
Related to https://github.com/woocommerce/woocommerce-ios/pull/8054 and https://github.com/woocommerce/woocommerce-ios/issues/8049

### Changes

To be able to defer the XMLRPC checks this PR adds a new configuration property to `WordPressAuthenticatorConfiguration`.
- New error to handle XMLRPC check related failures
- If the property `checkXMLRPCOnlyIfSigningInUsingSiteCredentials` is enabled 
  - XMLRPC checks will be skipped in `SiteAddressViewController`
  - XMLRPC checks will be run in `GetStartedViewController`
  - A loading indicator will be displayed over sign in using site credentials button while running XMLRPC checks.
  - In case of a failure `XMLRPCError` is sent via delegate. (Handled in Woo)
- If the property `checkXMLRPCOnlyIfSigningInUsingSiteCredentials` is disable (default value)
  -  There should be no change to the existing behaviour. 

### Testing
Please follow the testing instructions from https://github.com/woocommerce/woocommerce-ios/pull/8054

### Screenshot

| | Dark | Light |
|-|-|-|
| Loading indicator in "Sign in with store credentials" button | Simulator Screen Shot - iPhone SE (3rd generation) - 2022-11-08 at 08 58 42 | Simulator Screen Shot - iPhone SE (3rd generation) - 2022-11-08 at 08 59 01 |

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
@@ -915,7 +915,7 @@ private extension GetStartedViewController {
                 /// If it does, insert the custom UI provided by the host app and exit early
                 if self.authenticationDelegate.shouldHandleError(xmlrpcError) {
                     self.authenticationDelegate.handleError(xmlrpcError) { customUI in
-                        self.pushCustomUI(customUI)
+                        self.pushCustomUIViewController(customUI)
                     }
 
                     return
@@ -928,19 +928,6 @@ private extension GetStartedViewController {
                     self.displayError(extractedXMLRPCError, sourceTag: self.sourceTag)
                 }
         })
-    }
-
-    /// Push a custom view controller, provided by a host app, to the navigation stack
-    func pushCustomUI(_ customUI: UIViewController) {
-        /// Assign the help button of the newly injected UI to the same help button we are currently displaying
-        /// We are making a somewhat big assumption here: the chrome of the new UI we insert would look like the UI
-        /// WPAuthenticator is already displaying. Which is risky, but also kind of makes sense, considering
-        /// we are also pushing that injected UI to the current navigation controller.
-        if WordPressAuthenticator.shared.delegate?.supportActionEnabled == true {
-            customUI.navigationItem.rightBarButtonItems = navigationItem.rightBarButtonItems
-        }
-
-        navigationController?.pushViewController(customUI, animated: true)
     }
 }
 
--- a/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
@@ -876,7 +876,9 @@ private extension GetStartedViewController {
 // MARK: - XMLRPC checks
 
 private extension GetStartedViewController {
-    func configureSiteCredsButton(loading: Bool) {
+    /// Configures the buttons and navigation item
+    ///
+    func configureScreenForRunningXMLRPCChecks(loading: Bool) {
         buttonViewController?.setTopButtonState(isLoading: false,
                                                 isEnabled: !loading)
         buttonViewController?.setBottomButtonState(isLoading: loading,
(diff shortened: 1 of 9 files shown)


WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -313,7 +319,13 @@ private extension GetStartedViewController {
     // MARK: - Sign in with site credentials Button Action
     @objc func handleSiteCredentialsButtonTapped() {
         tracker.track(click: .signInWithSiteCredentials)
-        goToSiteCredentialsScreen()
+        guard WordPressAuthenticator.shared.configuration.checkXMLRPCOnlyIfSigningInUsingSiteCredentials else {
```
itsmeichigo: Nit: this can be simplified as
```suggestion
        guard configuration.checkXMLRPCOnlyIfSigningInUsingSiteCredentials else {
```

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
```diff
@@ -34,6 +34,12 @@ public enum SignInError: Error {
     }
 }
 
+/// Error due to site's `xmlrpc.php` file not being public
+///
+public enum XMLRPCError: Error {
+    case xmlrpcError(siteAddress: String)
+}
```
itsmeichigo, line 41: As commented in https://github.com/woocommerce/woocommerce-ios/pull/8054, I think this error should be removed since we can use the existing `WordPressOrgXMLRPCValidationError` instead.

itsmeichigo (commented):
We should reuse `WordPressOrgXMLRPCValidationError` instead of creating a new error `XMLRPCError` to avoid confusion - WDYT? 

Can we also reuse `guessXMLRPCURL` implementation in both view controllers with a protocol to avoid duplicated code?

WordPressAuthenticator/Unified Auth/View Related/Get Started/GetStartedViewController.swift
selanthiraiyan: I created this error to be able to pass `siteAddress` to the delegate along with this error. I will see whether there is another way to pass that. Kindly let me know if you already know of a way to pass the `siteAddress`.

itsmeichigo: Interesting, I didn't know the reason behind the use of the new type. Maybe we can get the site address on the Woo app by checking the [`currentSelfHostedSite`](https://github.com/woocommerce/woocommerce-ios/blob/7b905f41eb451db04a65a979ba58ffd6e3327113/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L34)? Does it sound too hacky though?

selanthiraiyan: Thanks for the input, Huong!

I see that `currentSelfHostedSite` will not be set [if `isSiteDiscovery` is true](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/trunk/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L582-L592). Please correct me if I am missing something.

To avoid dealing with edge cases, I think sticking to `XMLRPCError` is straightforward. What do you think?

itsmeichigo: Another hack: we can also add a new method in the delegate that is similar to the current [`handleError`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/964b007c21b7c6dbb0760ba5611bd537081ae5c8/WordPressAuthenticator/Authenticator/WordPressAuthenticatorDelegateProtocol.swift#L90) method, but with an extra param for a site address.

Then the default implementation for this method will call the original one; that way it will not affect existing host apps. Does that sound overcomplicated? If it does, we can keep your current solution and iterate later if needed.

selanthiraiyan:
Closing due to https://github.com/woocommerce/woocommerce-ios/issues/8049#issuecomment-1308155460

PR #701, opened 2022-11-09
New tracking event for XMLRPC error

selanthiraiyan:
Related to https://github.com/woocommerce/woocommerce-ios/issues/8063

This PR adds a new event for tracking XMLRPC URL guessing failures.

### Changes
- New `Failure` enum to track failures.
- Track `loginFailedToGuessXMLRPC` event when guessing XMLRPC fails.

### Testing instructions
Please follow the testing instructions from - https://github.com/woocommerce/woocommerce-ios/pull/8064

---

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -493,9 +493,9 @@ private extension SiteAddressViewController {
                 // It's not guaranteed to be included in the error object depending on the error.
                 DDLogInfo("Error attempting to connect to site address: \(self.loginFields.siteAddress)")
                 DDLogError(error.localizedDescription)
-                // TODO: - Tracks.
-                // WordPressAuthenticator.track(.loginFailedToGuessXMLRPC, error: error)
-                // WordPressAuthenticator.track(.loginFailed, error: error)
+
+                self.tracker.track(failure: .loginFailedToGuessXMLRPC)
+
                 self.configureViewLoading(false)
 
                 guard self.isSiteDiscovery == false else {
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -505,27 +505,41 @@ private extension SiteAddressViewController {
 
                 let err = self.originalErrorOrError(error: error as NSError)
 
+                let errorMessage: String? = {
+                    if let xmlrpcValidatorError = err as? WordPressOrgXMLRPCValidatorError {
+                        return xmlrpcValidatorError.localizedDescription
+                    } else if (err.domain == NSURLErrorDomain && err.code == NSURLErrorCannotFindHost) ||
+                                (err.domain == NSURLErrorDomain && err.code == NSURLErrorNetworkConnectionLost) {
+                        // NSURLErrorNetworkConnectionLost can be returned when an invalid URL is entered.
+                        let msg = NSLocalizedString(
+                            "The site at this address is not a WordPress site. For us to connect to it, the site must use WordPress.",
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -493,9 +493,9 @@ private extension SiteAddressViewController {
                 // It's not guaranteed to be included in the error object depending on the error.
                 DDLogInfo("Error attempting to connect to site address: \(self.loginFields.siteAddress)")
                 DDLogError(error.localizedDescription)
-                // TODO: - Tracks.
-                // WordPressAuthenticator.track(.loginFailedToGuessXMLRPC, error: error)
-                // WordPressAuthenticator.track(.loginFailed, error: error)
+
+                self.tracker.track(failure: .loginFailedToGuessXMLRPC)
```
itsmeichigo, line 497: As I look at [the few lines below](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/c169358f383aa0bcbfb383dfa03ae983d543b266/WordPressAuthenticator/Unified%20Auth/View%20Related/Site%20Address/SiteAddressViewController.swift#L518-L531), I see that the error can be something other than XML-RPC issue. Should we reuse the `message` from those lines and log them when `authenticationDelegate.shouldHandleError` returns `true`?

selanthiraiyan: Sorry, I am trying to understand this. 

I believe the purpose of this event is to track the number of failures happen at this "guessing XMLRPC" step of the login process. 

> I see that the error can be something other than XML-RPC issue

If we eventually remove XMLRPC, those other issues wouldn't really matter to the WooCommerce app as long as we can establish communication with the REST API. (Jetpack tunnel) Please correct me if I am wrong. (i.e. WordPress mobile app can partially work with just XMLRPC so these errors would matter to them.)

> Should we reuse the message from those lines and log them when authenticationDelegate.shouldHandleError returns true?

Do you mean to keep this new tracking event and also add another tracking event for this case?

itsmeichigo: I was under the impression that we want to track the failure disregarding whether the delegate should handle the error. My suggestion would be like so:

```Swift
let message: String? = {
    if let xmlrpcValidatorError = err as? WordPressOrgXMLRPCValidatorError {
        return xmlrpcValidatorError.localizedDescription
        
    } else if (err.domain == NSURLErrorDomain && err.code == NSURLErrorCannotFindHost) ||
                (err.domain == NSURLErrorDomain && err.code == NSURLErrorNetworkConnectionLost) {
        // NSURLErrorNetworkConnectionLost can be returned when an invalid URL is entered.
        let msg = NSLocalizedString(
            "The site at this address is not a WordPress site. For us to connect to it, the site must use WordPress.",
            comment: "Error message shown a URL does not point to an existing site.")
        return msg
        
    } else {
        return nil
    }
}

/// Check if the host app wants to provide custom UI to handle the error.
/// If it does, insert the custom UI provided by the host app and exit early
if self.authenticationDelegate.shouldHandleError(err) {
    self.authenticationDelegate.handleError(err) { customUI in
        self.pushCustomUI(customUI)
    }

    return self.tracker.track(failure: message)
}

if let message {
    self.displayError(message: message, moveVoiceOverFocus: true)
} else {
    self.displayError(error as NSError, sourceTag: self.sourceTag)
}
```

With that, we can still keep the query for the funnel as we did before to calculate the failures for XMLRPC. WDYT?

selanthiraiyan: Thanks for the detailed explanation and code. 😄 

Done in f2ad89e43f769ac144dcb19df89bb5c78efa7c2a

itsmeichigo (approved):
Thanks for the update! :shipit:

PR #704, opened 2022-11-15
WIP – Rough, happy-path-only SDK-less Google sign in

mokagio:
**Super early stage**, here just as a proof of concept.

To test, run the demo app from this branch locally and select the "ASWebAuthenticationSession" option. It will prompt to login with Google. One authorized from the Safari modal, the flow should succeed and print tokens in the console.

## Feedback Please

@crazytonyli @jkmassel I'm keen on your input, in particular in regards to this question: Should we roll our own implementation for OAuth via `ASWebAuthenticationSession`, or use [OpenID's AppAuth](https://github.com/openid/AppAuth-iOS)?

I stumbled into AppAuth while reverse engineering the value for some parameter via break points. GoogleSignIn uses it under the hood. The advantage of using AppAuth would be a ready-made, battle tested, comprehensive OAuth implementation. The disadvantages are the Apache 2.0 license (which I think we'd like to avoid?) and the obvious fact that we'd be depending on the library going forward. Granted, we already depend on in indirectly via GoogleSignIn, but with the point of this process being to remove the upstream dependency, it might be inconsistent to retain AppAuth 🤔 

### Next steps for "Roll our own"

I made a lot of tradeoffs to get this PoC working. If we were to continue down this path, I'd:

- Address the various `FIXME` and `TODO`...
- Use `async throws` as discussed in https://github.com/woocommerce/woocommerce-ios/issues/8104.
- Define and throw granular errors.
- Move all the implementation in the actual framework. It's currently in the demo app purely for convenience.
- Audit our clients to understand what values, if any, they read from the `GIDUser` object we currently pass them, so that we can provide the same info (if deemed useful) via our bespoke one.
- Look for further optimizations in how we interact with Google APIs, e.g. with refreshing the token etc

---


mokagio:
This has become stale and redundant.

PR #705, opened 2022-11-21
Prepare demo app for showing WordPressAuthenticator features

mokagio:
This is an intermediate-step PR to keep future diffs cleaner.  

- Restructures the demo app away from Storyboards
- Sets up a programmatic table view configuration
- Adds a few syntax sugars and utils

Basically, a lot of boilerplate.

See also followup #706 which adds the monster-long WordPressAuthenticator configuration call and presents a working login via email flow.

Screen Shot 2022-11-21 at 12 55 42 pm

---


Demo/AuthenticatorDemo/WKWebView+UserAgent.swift
```diff
@@ -0,0 +1,11 @@
...
+
+extension WKWebView {
+
+    static var userAgent: String {
+        guard let userAgent = Self().value(forKey: "_userAgent") as? String, userAgent.isEmpty == false else {
+            return ""
+        }
+        return userAgent
+    }
+}
```
mokagio, line 11: Would it be useful to move this to the framework to make it available to consumers?

If so, would it be reasonable to keep the `""` fallback under the assumption that we'll always have a user agent from `WKWebView`, or should we return `String?` or make it `throw`?

I'm leaning toward bubbling up the error and not make any assumption, because we're WordPressAuthenticator is a library and falling back to `""` for the user agent doesn't seem appropriate.

Demo/AuthenticatorDemo/ViewController.swift
```diff
@@ -1,10 +1,90 @@
 import UIKit
 import WordPressAuthenticator
 
+/// Starting point for the demo app. It shows a table view with actionable rows each loading a different authentication flow.
+/// (Currently, we only have one authentication flow).
```
mokagio, line 5: This statement is incorrect, but it will be right in #706.

Demo/AuthenticatorDemo/WKWebView+UserAgent.swift
crazytonyli: I'm leaning towards not exposing this as API, at least not this implementation. Creating a `WKWebView` to get a user agent string seems a little bit too expensive to me. Not to mention it uses a private API here. Unless the backend HTTP API requires a "browser like" user agent.

crazytonyli (approved):

PR #709, opened 2022-11-28
Implement existing Google-SignIn flow in the demo app

mokagio:
<table>
<tr>
	<td>Screenshot 2022-11-28 at 4 26 04 pm</td>
	<td>
Screenshot 2022-11-28 at 4 26 16 pm</td>
	<td>Screenshot 2022-11-28 at 4 27 29 pm
</td>
</tr>
</table>

When testing, you might see the following error in the console:

```
[Authorization] ASAuthorizationController credential request failed with error:
Error Domain=com.apple.AuthenticationServices.AuthorizationError Code=1001 "(null)"
```

Notice, though, that it's logged _before_ the start of any login flow. It will be good to chase it down, but it's out of scope for this time.

---

### DIFF
--- a/Demo/AuthenticatorDemo/Info.plist
+++ b/Demo/AuthenticatorDemo/Info.plist
@@ -2,6 +2,17 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>com.googleusercontent.apps.108380595987-ujhrhknecrqli756i72gkcs4aaia6nhb</string>
+			</array>
+		</dict>
+	</array>
 	<key>UIApplicationSceneManifest</key>
 	<dict>
 		<key>UIApplicationSupportsMultipleScenes</key>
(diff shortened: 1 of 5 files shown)


Demo/AuthenticatorDemo/Info.plist
```diff
@@ -2,6 +2,17 @@
...
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>com.googleusercontent.apps.108380595987-ujhrhknecrqli756i72gkcs4aaia6nhb</string>
```
mokagio, line 12: I recall a discussion that might have been similar to what I'm about to write, but it might have been in an internal/private Slack channel so I'm taking the occasion for having it in a searchable medium.

Upon first looking at this, I though it was a security risk because it clearly exposes information relevant to the WordPress app only.

However:

- This is already committed [in the WordPress iOS repo](https://github.com/wordpress-mobile/WordPress-iOS/blob/1b266dbd810b2982b9ca37996d85660fa94490ab/WordPress/Info.plist#L490)
- Any savvy bad actor would be able to extract it from the app at runtime

I'm open to the idea of injecting this at build time in a more secure way like we do for `APICredentials` but, because this value has already "leaked", I would do it in a set of dedicated PR across all our apps.

crazytonyli: I don't think it's a risk to expose the client id.

> The client identifier is not a secret;

According to [this OAuth 2 speficication](https://www.rfc-editor.org/rfc/rfc6749#section-2.2).

crazytonyli (approved):

Demo/AuthenticatorDemo/Info.plist
mokagio: Even better 😄

PR #710, opened 2022-11-29
SDK-less Google SignIn – Part 1 of n

mokagio:
Took the work from https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/704 and replayed the first parts adding tests.

Nothing to test in the demo app yet. Just adding some components to build things with later. I opened this as a ready to review PR because I don't want the diff to become overwhelming.

---

I'll add the changelog entry in the last PR of the series.

### DIFF
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+extension URL {
+
+    // TODO: This is incomplete
+    static func googleSignInAuthURL() throws -> URL {
+        let baseURL = "https://accounts.google.com/o/oauth2/v2/auth"
+
+        let queryItems = [
+            ("response_type", "code")
+        ].map { URLQueryItem(name: $0.0, value: $0.1) }
+
+        if #available(iOS 16.0, *) {
+            return URL(string: baseURL)!.appending(queryItems: queryItems)
+        } else {
+            var components = URLComponents(string: baseURL)!
+            components.queryItems = queryItems
+            return try components.asURL()
+        }
+    }
+}
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -3,10 +3,11 @@ import Foundation
 extension URL {
 
     // TODO: This is incomplete
-    static func googleSignInAuthURL() throws -> URL {
+    static func googleSignInAuthURL(clientId: String) throws -> URL {
         let baseURL = "https://accounts.google.com/o/oauth2/v2/auth"
 
         let queryItems = [
+            ("client_id", clientId),
             ("response_type", "code")
         ].map { URLQueryItem(name: $0.0, value: $0.1) }
 
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -8,6 +8,7 @@ extension URL {
 
         let queryItems = [
             ("client_id", clientId),
+            ("redirect_uri", redirectURI(from: clientId)),
             ("response_type", "code")
         ].map { URLQueryItem(name: $0.0, value: $0.1) }
 
@@ -18,5 +19,14 @@ extension URL {
             components.queryItems = queryItems
             return try components.asURL()
         }
+    }
+
+    private static func redirectURI(from clientId: String) -> String {
+        // Google's client id is in the form: 123-abc245def.apps.googleusercontent.com
+        // The redirect URI uses the reverse-DNS notation.
+        let reverseDNSClientId = clientId.split(separator: ".").reversed().joined(separator: ".")
+        // After that, we add "oautha2callback", as per GIDSignIn.m line 421 at
+        // commit 1b0c4ec33a6fe282f4fa35d8ac64263230ddaf36
+        return "\(reverseDNSClientId):/oauth2callback"
     }
 }
--- a/WordPressAuthenticator/OAuth/ASWebAuthenticationSession+Utils.swift .swift
+++ b/WordPressAuthenticator/OAuth/ASWebAuthenticationSession+Utils.swift .swift
@@ -12,7 +12,7 @@ extension ASWebAuthenticationSession {
                     error: error,
                     // Unfortunately we cannot exted `ASWebAuthenticationSessionError.Code` to add
                     // a custom error for this scenario, so we're left to use a "generic" one.
-                    inconsistentStateError: OAuthErrors.inconsistentASWebAuthenticationSessionCompletion
+                    inconsistentStateError: OAuthError.inconsistentWebAuthenticationSessionCompletion
                 )
             )
         }
--- a/WordPressAuthenticator/OAuth/ASWebAuthenticationSession+Utils.swift .swift
+++ b/WordPressAuthenticator/OAuth/ASWebAuthenticationSession+Utils.swift .swift
@@ -1,0 +1,20 @@
+import AuthenticationServices
+
+extension ASWebAuthenticationSession {
+
+    /// Wrapper around the default `init(url:, callbackULRScheme:, completionHandler:)` where the
+    /// `completionHandler` argument is a `Result<URL, Error>` instead of a `URL` and `Error` pair.
+    convenience init(url: URL, callbackURLScheme: String, completionHandler: @escaping (Result<URL, Error>) -> Void) {
+        self.init(url: url, callbackURLScheme: callbackURLScheme) { callbackURL, error in
+            completionHandler(
+                Result(
+                    value: callbackURL,
+                    error: error,
+                    // Unfortunately we cannot exted `ASWebAuthenticationSessionError.Code` to add
+                    // a custom error for this scenario, so we're left to use a "generic" one.
+                    inconsistentStateError: OAuthErrors.inconsistentASWebAuthenticationSessionCompletion
+                )
+            )
+        }
+    }
+}
--- a/WordPressAuthenticator/OAuth/OAuthError.swift
+++ b/WordPressAuthenticator/OAuth/OAuthError.swift
@@ -1,0 +1,10 @@
+enum OAuthErrors {
+
+    static let inconsistentASWebAuthenticationSessionCompletion = NSError(
+        domain: "org.wordpress.authenticator.oauth",
+        code: 1,
+        userInfo: [
+            NSLocalizedDescriptionKey: "ASWebAuthenticationSession authentication finished with neither a callback URL nor error"
+        ]
+    )
+}
--- a/WordPressAuthenticator/OAuth/OAuthError.swift
+++ b/WordPressAuthenticator/OAuth/OAuthError.swift
@@ -1,10 +1,11 @@
-enum OAuthErrors {
+enum OAuthError: LocalizedError {
 
-    static let inconsistentASWebAuthenticationSessionCompletion = NSError(
-        domain: "org.wordpress.authenticator.oauth",
-        code: 1,
-        userInfo: [
-            NSLocalizedDescriptionKey: "ASWebAuthenticationSession authentication finished with neither a callback URL nor error"
-        ]
-    )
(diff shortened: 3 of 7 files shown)


WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+extension URL {
+
+    // TODO: This is incomplete
+    static func googleSignInAuthURL(clientId: String) throws -> URL {
```
mokagio, line 6: I used to add all sorts of domain-specific methods to types such as `URL` in the past, then realized it's seldom appropriate because it "leaks" those methods to all the possible consumers of the type. 

For example, in this case, any class that can `import Foundation` and access `URL` will now have the `googleSignInAuthURL(...)` method in its autocompletion list.

In an app, that would definitely be confusing. In a relatively narrow focused library, I think that's acceptable.

However, I'm happy to move this logic in an dedicated object if you think it would be better. Something like `GoogleSignInURLBuilder.makeAuthURL(clientId: ...)`.

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -0,0 +1,32 @@
...
+            return try components.asURL()
+        }
+    }
+
+    private static func redirectURI(from clientId: String) -> String {
+        // Google's client id is in the form: 123-abc245def.apps.googleusercontent.com
+        // The redirect URI uses the reverse-DNS notation.
+        let reverseDNSClientId = clientId.split(separator: ".").reversed().joined(separator: ".")
+        // After that, we add "oautha2callback", as per GIDSignIn.m line 421 at
+        // commit 1b0c4ec33a6fe282f4fa35d8ac64263230ddaf36
```
mokagio, line 29: I guess I could add a link to the GitHub page for this...

WordPressAuthenticator/OAuth/ASWebAuthenticationSession+Utils.swift .swift
```diff
@@ -0,0 +1,20 @@
...
+    convenience init(url: URL, callbackURLScheme: String, completionHandler: @escaping (Result<URL, Error>) -> Void) {
+        self.init(url: url, callbackURLScheme: callbackURLScheme) { callbackURL, error in
+            completionHandler(
+                Result(
+                    value: callbackURL,
+                    error: error,
+                    // Unfortunately we cannot exted `ASWebAuthenticationSessionError.Code` to add
+                    // a custom error for this scenario, so we're left to use a "generic" one.
+                    inconsistentStateError: OAuthErrors.inconsistentASWebAuthenticationSessionCompletion
+                )
```
mokagio: By delegating most of the logic in this method on `ASWebAuthenticatorSession` to a general purpose `Result` method, I was able to test most of its behavior.

WordPressAuthenticator/OAuth/OAuthError.swift
```diff
@@ -0,0 +1,10 @@
+enum OAuthErrors {
```
mokagio: Using an `enum` means we can't accidentally call do `let something = OAuthErrors()`.

Although, there might be better typed ways to have a collection of errors. But, since we don't need to specify an error type in `throws` or `Result`, I think a container of `NSError` such as this could also do.

WordPressAuthenticatorTests/GoogleSignIn/URL+GoogleSignInTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    guard let baseURL = components.url else {
+        return XCTFail(
+            "Could not extract `URL` from `URLComponents` created from \(actual).",
+            file: file,
+            line: line
+        )
+    }
+
+    XCTAssertEqual(baseURL.absoluteString, baseURLString, file: file, line: line)
+}
```
mokagio, line 50: This and the other assertion method could go in a dedicated file, I guess.

I might do that if I'll end up writing more tests that need them?

WordPressAuthenticatorTests/GoogleSignIn/URL+GoogleSignInTests.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    guard let queryItems = components.queryItems else {
+        XCTFail("URL \(url) has no query items", file: file, line: line)
+        return
+    }
+
+    XCTAssertTrue(
+        queryItems.contains(where: { $0.name == name && $0.value == value }),
+        "Could not find query item with name '\(name)' and value '\(value ?? "nil")'. Query items found: \(queryItems.map { "'name: \($0.name), value: \($0.value ?? "nil")'" }.joined(separator: ", "))",
```
mokagio, line 74: I considered using the multiline string format for this, but kept the single line for simplicity. What do you think?

WordPressAuthenticator/OAuth/OAuthError.swift
mokagio: The more I think about it, actually, the less I'm happy with `NSError`. Keen to hear what you think.

If / when we'll improve the error system, it might be good to do it in a dedicated PR.

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -0,0 +1,32 @@
...
+            components.queryItems = queryItems
+            return try components.asURL()
+        }
+    }
+
+    private static func redirectURI(from clientId: String) -> String {
+        // Google's client id is in the form: 123-abc245def.apps.googleusercontent.com
+        // The redirect URI uses the reverse-DNS notation.
+        let reverseDNSClientId = clientId.split(separator: ".").reversed().joined(separator: ".")
+        // After that, we add "oautha2callback", as per GIDSignIn.m line 421 at
```
crazytonyli, line 28: Is it intended to use the same URL scheme as GoogleSignIn? Do you think we can use a customized URL scheme, like `wpauthenticator-<google_app_id>://oauth2callback`, so that it's clear this URL should be handled by WordPressAuthenticator, not any other library?

WordPressAuthenticator/OAuth/OAuthError.swift
crazytonyli: I'd prefer using `Swift.Error`, which is more user friendly when it comes to writing error handling code.

mokagio: I've been thinking about this and I wonder if this would be good.

```swift
enum OAuthError {

  struct InconsistentASWebAuthenticationSessionCompletion: LocalizedError {
    let errorDescription = "ASWebAuthenticationSession authentication finished with neither a callback URL nor error"
  }

  struct OtherThingThatWentWrong: LocalizedError {}

  struct YetAnotherThingThatWentWrong: LocalizedError {
    let aProperty: Int
  }

  // ...
}
```

I've seen some codebases use `enum OAuthErro: Error` but I don't like that because you end up with long `switch case` for `errorDescription` and the other `LocalizedError` properties. This approach still has the nice namespacing, but also keeps each error self contained. Removing one would amount to remove a continuous group of lines, rather that a `case name` definition here and various `case .name: ...` usages there.

What do you think?

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
mokagio: Good question. I wasn't able to get it to work with anything else when prototyping #740 but I learned more since... 🤔 

I might give this another go.

To answer "Is it intended to use the same URL scheme as GoogleSignIn?" specifically, the [docs](https://developers.google.com/identity/protocols/oauth2/native-app#request-parameter-redirect_uri) list that as a valid value to use, and that's what their SDK does.

Thinking about it, "that's what their SDK does" is kinda the obvious path _for their SDK_, but it might be best for us to use a custom value, as you suggest. There's nothing in the docs that prevents it, so it might just be a matter of me figuring out where my early mistakes were.

WordPressAuthenticator/OAuth/OAuthError.swift
mokagio: For reference, [the Swift docs](https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html) use an example with `enum`:

```swift
enum VendingMachineError: Error {
    case invalidSelection
    case insufficientFunds(coinsNeeded: Int)
    case outOfStock
}
```

Another option I found which could be interesting is [using nested `enums`](https://vojtastavik.com/2017/04/23/organize-your-errors-swift/): 

```swift
enum Error: Swift.Error {
    case network(NetworkError)
    enum NetworkError {
        case notReachable
        case unknown
    }

    case device(DeviceError)
    enum DeviceError {
        case notEnoughSpace
        case unsupportedSystemVersion
    }
}
```

I like the idea of grouping all errors as `enum` cases, but I think I still prefer the encapsulation that `struct MyError: Error` brings. I would modify that example to:

```swift
protocol NetworkError: Swift.Error {}
protocol DeviceError: Swift.Error {}

enum Error: Swift.Error {
    case network(NetworkError)
    case device(DeviceError)

    enum Network {
        struct NotReachableError: NetworkError {}
        struct Unknown: NetworkError {}
    }

    enum Device: DeviceError {
        struct NotEnoughSpaceError: DeviceError {}
        struct UnsupportedSystemVersionError: DeviceError {}
    }
}

# Used in practice as

let e1 = Error.device(Error.Device.NotEnoughSpaceError())
let e2 = Error.network(Error.Network.NotReachableError())
```

Admittedly, one loses the ability to `switch` on the `NetworkError` and `DeviceError`, and it's unfortunate that the `NetworkError` and `DeviceError` protocols cannot be defined as nested in `enum Error`. Still, I think this would scale better.

Also, if really someone really wanted to switch between the leaves, they could do it via:

```swift
    switch error {
    case .network: break
    case .device(let deviceError):
        switch deviceError.self {
        case is Error.Device.UnsupportedSystemVersionError:
            print("Device error")
        case is Error.Device.NotEnoughSpaceError:
            print("Not enough space error")
        case _:
            fatalError("This code should not have been reached") // Yet another limitation of this approach
        }
    }
```

crazytonyli: In my experience, I only used error declaration similar to the `VendingMachineError` in your example. Yes, you do end up with a long switch case, but I think that's what you want in this instance, as you want _all_ errors to be handled, and Swift compiler can help you enforcing that (unless there is a `default` clause in the switch-case).

Regarding your second example, I don't think I have ever written a nested error type like. First of all, the associated value of the Error enum cases does not need to conform to `Swift.Error` (I don't see any benefit in doing that). Also, I feel like it's more readable to flatten the nested enum cases. In your example, there are really four possible errors (rather than two as declared in the `enum Error`): "network not reachable", unknown network issue, unsupported OS, and not enough disk space. I don't see the benefit of categorizing these four errors into "device" and "network".

mokagio: I've been trying out the `enum` collecting `struct E: LocalizedError` approach in my development branch, with [this](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/77fd7040680225cf92b0c12ec64569bcbca8bb0b/WordPressAuthenticator/OAuth/OAuthError.swift) being latest iteration at the time of writing.

I really dislike the spreading of information that we get with the `enum` approach, but I agree with the compile time benefit that you highlight.

> as you want all errors to be handled, and Swift compiler can help you enforcing that (unless there is a default clause in the switch-case).

None of the other approaches I tried is as secure and supported as what the compiler can do for us when using an `enum`.

As for nesting, leaving aside the oversight of having associated types not conforming to `Swift.Error`, I'd keep that as a technique up our sleeves to reduce the noise if we end up with a lot of errors, but I guess it might be best to have more specialized error types, e.g. `enum RequestError: Error` and `enum ResponseError: Error` that a catch-all error type with cases for both.

mokagio:
@crazytonyli I addressed your suggestion to use `enum` for the errors. As for [using a custom URI for the redirect](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/710#discussion_r1035358491), I'd like to get the properly written code and demo app using it done first, and then try the approach. What do you think?

crazytonyli (approved):

crazytonyli:
@mokagio Sounds perfect. Thanks!

PR #711, opened 2022-11-30
Make XMLRPC URL optional when verifying WP.com email and new config for skipping XMLRPC for site discovery

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/8258

## Description

Currently, when presenting the flow for verifying WP.com email, we are requiring that a valid XMLRPC URL should be provided. This is not necessary since this flow only handles authentication with WP.com. 

This PR fixes this by making XMLRPC optional for verifying WP.com email. It's better to remove it entirely, but I don't want to introduce a breaking change, so I'm still keeping XMLRPC as a parameter for the method. Please feel free to suggest another solution that's cleaner for this situation.

 A new config is also added to skip the XMLRPC check for the site discovery flow.

## Testing steps
Please follow the steps in https://github.com/woocommerce/woocommerce-ios/pull/8259 for testing.

---

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -28,7 +28,7 @@ PODS:
   - SVProgressHUD (2.2.5)
   - SwiftLint (0.49.1)
   - UIDeviceIdentifier (1.6.0)
-  - WordPressAuthenticator (4.2.0):
+  - WordPressAuthenticator (4.3.0-beta.1):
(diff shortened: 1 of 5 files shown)


selanthiraiyan (approved):
LGTM 🚢

PR #714, opened 2022-12-07
Add UI tests for demo app

mokagio:
We'll use them to test the new Google SignIn flow.

Currently checking two simple scenarios in the legacy one: pressing "Cancel" and "Continue" in the alert to start the flow.

Originally, I had the tests run in CI as well. Then they started failing for a reason I don't understand yet (the screenshot is of springboard... could the test app be crashing?) but they _work on my machine._

We don't need them in CI just yet, as I'm the only user for the moment. They could also remain only a development tool, but that would be a missed opportunity.

## To test

Run `bundle exec fastlane ui_test_demo_app` locally.

---


.buildkite/pipeline.yml
```diff
@@ -16,11 +16,14 @@ steps:
     env: *common_env
     plugins: *common_plugins
 
-  - label: ":calling: Build Demo App"
+  - label: ":calling: UI Test Demo App"
     key: "build-demo-app"
```
mokagio: This key should be updated, too, I guess.

.buildkite/ui-test-demo-app.sh
```diff
@@ -0,0 +1,28 @@
...
+set -e
+
+if [[ "$TESTS_EXIT_STATUS" -ne 0 ]]; then
+  # Keep the (otherwise collapsed) current "Testing" section open in Buildkite logs on error. See https://buildkite.com/do s/pipelines/managing-log-output#collapsing-output
+  echo "^^^ +++"
+  echo "UI Tests failed!"
+
+  echo "--- :camera_with_flash: Extracting Screenshots"
+  brew install chargepoint/xcparse/xcparse
+  xcparse screenshots fastlane/test_output/UITests.xcresult .build/screenshots/
```
mokagio: During development, the Google SignIn rendered in Safari was in Italian, because I'm currently in Italy, so I used _and commited_ Italian copy.

Then forgot all about it.

When the tests failed, I couldn't understand why, so I started working on uploading the failure screenshots.

By the time I was done, I had realized my mistake. Still, I'm sure this will be useful in the future.

Demo/AuthenticatorDemoUITests/PrologueScreenObject.swift
```diff
@@ -0,0 +1,37 @@
+import ScreenObject
+import XCTest
+
+/// Slimmed down version of the one found in the WordPress iOS codebased.
+///
+/// Eventually, this should end up in a UI testing helper framework that the package exports, so that consumers can use it to write their
+/// own UI tests if they see fit.
```
mokagio, line 7: I hope to get onto this sooner rather than later.

Because of the end of year vacation period coming up, "sooner" will likely end up being Jan-Feb 2023.

.buildkite/ui-test-demo-app.sh
mokagio: There are quite a few requirements and parameters that go into this approach, such as the prefix used in the code, the `.xcresult` name, and the fact that there are screenshots in the first place. Still, I can see this being extracted in our Buildkite plugin and adopted across the board.

Demo/AuthenticatorDemoUITests/AuthenticatorDemoUITests.swift
```diff
@@ -0,0 +1,55 @@
...
+        // - Add a custom prefix to identify these screenshots from those automatically taken
+        // - Need to append an extension for the image to be properly extracted
+        screenshot.name = "failure-\(name).png"
+
+        screenshot.lifetime = .keepAlways
+
+        add(screenshot)
+
+        super.record(issue)
+    }
```
mokagio: We can configure the test plan to take screenshots, of course, but it ends up with more screenshots than we'd like. This approach adds a bit of complexity, but gives us more control.

Demo/AuthenticatorDemoUITests/AuthenticatorDemoUITests.swift
```diff
@@ -0,0 +1,35 @@
+import XCTest
+
+final class AuthenticatorDemoUITests: XCTestCase {
```
mokagio, line 3: Xcode made this `final`. I don't know if there's a reason. I thought I'd leave it as is in case I find out why.

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+        static let next = "Next"
+    }
+
+    init(app: XCUIApplication) {
+        self.app = app
+    }
+
+    func authenticate(file: StaticString = #file, line: UInt = #line) {
+        // The device in which we're running the tests might have pre-existing accounts available.
+        let emailField = app.staticTexts[email]
```
pachlava: It took me some time to understand that `emailField` is about the tappable, non-edit element with the email address for the case of previously logged-in user (and not the input for email). WDYT of making the name sth like `loggedInTestAccount`, `alreadyLoggedIn`? (both are not perfect, but they reflect the general idea).

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+import XCTest
+
+struct GoogleSignInSafariScreen {
+
+    private let app: XCUIApplication
+    private let email = APICredentials.Tests.GoogleAccount.email
+    private let password = APICredentials.Tests.GoogleAccount.password
+
+    enum Labels {
+        static let selectDifferentAccount = "Use another account"
```
pachlava, line 19: For me, this part did not work, because it was Ukrainian localization for my case. I don't think this is critical - when it runs on CI, it should be permanent (probably English?). And I don't think it's worth it to locate this element differently (e.g. `boundBy` or something like this), because the UI might change there.

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+                selectDifferentAccount.tap()
+            }
+            typeEmail()
+            typePassword()
+        }
+
+        // To make sure the credentials input screen is gone, check none of the input elemets are
+        // on screen.
+        XCTAssertFalse(app.textFields.firstMatch.exists, file: file, line: line)
+        XCTAssertFalse(app.secureTextFields.firstMatch.exists, file: file, line: line)
```
pachlava: Just leaving a note that this area was a bit flaky for me.

1. First assertion: When I was using a lot of breakpoints (so there was plenty of time for the final screen to load, before the first assertion was executed), if failed, because actually this element below (the blurred email address) is also a TextField (non-editable, though):

Screenshot 2022-12-09 at 00 43 04

- Second assertion: Once I disabled the first assertion, the second one started to fail, when I was not in debug mode, because the password input was still on-screen after `Go` was tapped.

I'm not insisting on removing them, if they work for you, but this might be a source of flakiness.

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private func typeEmail() {
+        app.textFields.firstMatch.typeText(email)
```
pachlava: For some reason, unless I tapped into the input before entering text, the text was not entered (might this be the reason for fails on CI 🤔 ?):

```suggestion
        let emailInput = app.textFields.firstMatch
        emailInput.tap()
        emailInput.typeText(email)
```

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+        continueFromKeyboardOrScreen()
+    }
+
+    private func typePassword() {
+        app.secureTextFields.firstMatch.typeText(password)
```
pachlava: Same as for email input above.

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
```diff
@@ -0,0 +1,73 @@
...
+            emailField.tap()
+        } else {
+            if selectDifferentAccount.waitForExistence(timeout: 5) {
+                selectDifferentAccount.tap()
+            }
+            typeEmail()
+            typePassword()
+        }
+
+        // To make sure the credentials input screen is gone, check none of the input elemets are
```
pachlava: Nitpick:
```suggestion
        // To make sure the credentials input screen is gone, check none of the input elements are
```

Demo/AuthenticatorDemoUITests/ContinueWithGoogleAlert.swift
```diff
@@ -0,0 +1,38 @@
...
+    private let alertGetter: (XCUIApplication) -> XCUIElement = {
+        // There might be a more precise way to get the alert, such as looking at this label
+        $0.alerts.firstMatch
+    }
+
+    private var cancelButton: XCUIElement { alertGetter(app).buttons["Cancel"] }
+    private var continueButton: XCUIElement { alertGetter(app).buttons["Continue"] }
+
+    init() throws {
+        // Usually, one wuold use addUIInterruptionMonitor to handle the alert iOS shows to hand
```
pachlava: Typo:
```suggestion
        // Usually, one would use addUIInterruptionMonitor to handle the alert iOS shows to hand
```

pachlava (commented):
Hey @mokagio 👋 

I've reviewed the UI tests. This is my first experience with WordPressAuthenticator, so I might have missed something. There were no issues at all with `testCancelling...`, but I spent some time debugging / looking into the second one. There are two concerns:

1. Without preliminary tapping the email and password inputs, the text was not entered in my case.
2. There was some flakiness in the double `XCTAssertFalse` area. I'm not sure if it's necessary to change this part, if it works, but I think it's worth to keep it in mind.

Both are described in more details in separate comments. Thanks for working on this - it was interesting to take a look in the code 🙇

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
mokagio: I set a breakpoint and used `po app.staticTexts` to see if there was any other way to find items that was not language specific. 

Unfortunately, I couldn't find any.

When VPNed to Italy, I got the following object 

```
      StaticText, 0x129e28ce0, {{64.0, 351.0}, {156.0, 19.0}}, label: 'Utilizza un altro account', value: Utilizza un altro ..., Focused
```

That is, `value` and `label` are both localized. So no luck.

I added a comment about this in 0b6c76f1

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
mokagio: I renamed it to `existingAccountEntry` in 51b7a1e4. What do you think?

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
mokagio: I'm pretty sure the issue here was with the network connectivity being slower than the time the UI tests allowed for looking for the elements.

Making the test async solved it for me locally.

See 98a19911.

Demo/AuthenticatorDemoUITests/GoogleSignInSafariScreen.swift
mokagio: 🤦‍♂️ 

So what happened here is that I run the tests on the Simulator I use for regular development, where I don't have the software keyboard enabled so I can type in it directly from my Mac's keyboard without clicking the individual keys in the on screen one. However, that's not how Simulators run by default.

I realized it when testing on a different Simulator (TIL, always run the UI tests on a few different sims). Adding the tap, as you suggested, solved it for me, too.

See c7fae528

mokagio:
Thank you for all the feedback @pachlava.

Unfortunately, I hit a roadblock... Google wants to verify the test account identity which isn't really something we can do in the context of a UI test. 😞 

Screenshot 2023-03-15 at 5 25 27 pm

If found [this documentation](https://developers.google.com/google-ads/api/docs/first-call/test-accounts) page that mentions the ability to create test accounts in Google, but it might be limited to Google Ads only.

mokagio:
Closing as stale. 

Thanks @pachlava for all the help!

PR #717, opened 2022-12-08
SDK-less Google-SignIn – Part 2 of n

mokagio:
Adds objects to model OAuth token request and response, the pre-requisite Proof of Code Encryption Key object for the request that generates the URL to use for the token, and the logic to generate the token request URL.

@crazytonyli one disadvantage of using an incremental approach to building a big feature such as this is that the first PRs fail to tell the story of the system being developed, because they are only individual components not yet used together.

Is this format acceptable for you to review? Otherwise I could:

- Rework my WIP code to create a new PR that adds the higher level consumer of all these pieces (working name `NewGoogleAuthenticator`) so that we can do PRs that add a few bits and also show how they are used; or
- Dump what I have in a >800 lines PR, after which it might be easier to move adding little bits to the existing end-to-end setup (eg. more error handling, MFA, etc...)
- Keep working like this and share another maybe 2-3 PRs before we get to a working end-to-end bare bone setup

Let me know what would work best for you. Thanks!

## To test

If CI is green, we're good.

## See also

- #710 

---

### DIFF
--- a/WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
@@ -15,8 +15,6 @@ extension OAuthTokenRequestBody {
             clientSecret: "",
             code: authCode,
             codeVerifier: pkce.codeVerifier,
-            // TODO: This might be hardcoded...
-            //
             // As defined in the OAuth 2.0 specification, this field's value must be set to authorization_code.
             // – https://developers.google.com/identity/protocols/oauth2/native-app#exchange-authorization-code
             grantType: "authorization_code",
--- a/WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
@@ -1,0 +1,26 @@
+extension OAuthTokenRequestBody {
+
+    static func googleSignInRequestBody(
+        clientId: String,
+        authCode: String,
+        pkce: ProofKeyForCodeExchange
+    ) -> Self {
+        .init(
+            clientId: clientId,
+            // "The client secret obtained from the API Console Credentials page."
+            // - https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server
+            //
+            // There doesn't seem to be any secret for iOS app credentials.
+            // The process works with an empty string...
+            clientSecret: "",
+            code: authCode,
+            codeVerifier: pkce.codeVerifier,
+            // TODO: This might be hardcoded...
+            //
+            // As defined in the OAuth 2.0 specification, this field's value must be set to authorization_code.
+            // – https://developers.google.com/identity/protocols/oauth2/native-app#exchange-authorization-code
+            grantType: "authorization_code",
+            redirectURI: URL.redirectURI(from: clientId)
+        )
+    }
+}
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -30,16 +30,9 @@ extension URL {
         if #available(iOS 16.0, *) {
             return googleSignInBaseURL.appending(queryItems: queryItems)
         } else {
-            let baseURL = googleSignInBaseURL
-            guard var components = URLComponents(url: baseURL, resolvingAgainstBaseURL: false) else {
-                throw URLError(
-                    .unsupportedURL,
-                    userInfo: [
-                        NSLocalizedDescriptionKey: "Could not create `URLComponents` instance from \(baseURL)"
-                    ]
-                )
-            }
-
+            // Given `googleSignInBaseURL` is assumed as a valid URL, a `URLComponents` instance
+            // should always be available.
+            var components = URLComponents(url: googleSignInBaseURL, resolvingAgainstBaseURL: false)!
             components.queryItems = queryItems
             return try components.asURL()
         }
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -6,7 +6,6 @@ extension URL {
     // error, which we would catch because the unit tests would crash.
     static var googleSignInBaseURL = URL(string: "https://accounts.google.com/o/oauth2/v2/auth")!
 
-    // TODO: This is incomplete
     static func googleSignInAuthURL(clientId: String, pkce: ProofKeyForCodeExchange) throws -> URL {
         let queryItems = [
             ("client_id", clientId),
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -46,7 +46,7 @@ extension URL {
         }
     }
 
-    private static func redirectURI(from clientId: String) -> String {
+    static func redirectURI(from clientId: String) -> String {
         // Google's client id is in the form: 123-abc245def.apps.googleusercontent.com
         // The redirect URI uses the reverse-DNS notation.
         let reverseDNSClientId = clientId.split(separator: ".").reversed().joined(separator: ".")
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -31,8 +31,16 @@ extension URL {
         if #available(iOS 16.0, *) {
             return googleSignInBaseURL.appending(queryItems: queryItems)
         } else {
-            // FIXME: Throw error
-            var components = URLComponents(url: googleSignInBaseURL, resolvingAgainstBaseURL: false)!
+            let baseURL = googleSignInBaseURL
+            guard var components = URLComponents(url: baseURL, resolvingAgainstBaseURL: false) else {
+                throw URLError(
+                    .unsupportedURL,
+                    userInfo: [
+                        NSLocalizedDescriptionKey: "Could not create `URLComponents` instance from \(baseURL)"
+                    ]
+                )
+            }
+
             components.queryItems = queryItems
             return try components.asURL()
         }
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -10,7 +10,7 @@ extension URL {
         let queryItems = [
             ("client_id", clientId),
             ("code_challenge", pkce.codeCallenge),
-            ("code_challenge_method", pkce.mode.method),
+            ("code_challenge_method", pkce.method.urlQueryParameterValue),
             ("redirect_uri", redirectURI(from: clientId)),
             ("response_type", "code"),
             // TODO: We might want to add some of these or them configurable
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -3,11 +3,13 @@ import Foundation
 extension URL {
 
     // TODO: This is incomplete
-    static func googleSignInAuthURL(clientId: String) throws -> URL {
+    static func googleSignInAuthURL(clientId: String, pkce: ProofKeyForCodeExchange) throws -> URL {
         let baseURL = "https://accounts.google.com/o/oauth2/v2/auth"
 
         let queryItems = [
             ("client_id", clientId),
+            ("code_challenge", pkce.codeCallenge),
+            ("code_challenge_method", pkce.mode.method),
             ("redirect_uri", redirectURI(from: clientId)),
             ("response_type", "code")
         ].map { URLQueryItem(name: $0.0, value: $0.1) }
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -11,7 +11,19 @@ extension URL {
             ("code_challenge", pkce.codeCallenge),
             ("code_challenge_method", pkce.mode.method),
             ("redirect_uri", redirectURI(from: clientId)),
-            ("response_type", "code")
+            ("response_type", "code"),
+            // TODO: We might want to add some of these or them configurable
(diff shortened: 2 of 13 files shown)


WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
```diff
@@ -0,0 +1,26 @@
...
+            // "The client secret obtained from the API Console Credentials page."
+            // - https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server
+            //
+            // There doesn't seem to be any secret for iOS app credentials.
+            // The process works with an empty string...
+            clientSecret: "",
+            code: authCode,
+            codeVerifier: pkce.codeVerifier,
+            // TODO: This might be hardcoded...
+            //
```
mokagio:
```suggestion
```

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -2,26 +2,51 @@ import Foundation
...
 extension URL {
 
-    // TODO: This is incomplete
-    static func googleSignInAuthURL(clientId: String) throws -> URL {
-        let baseURL = "https://accounts.google.com/o/oauth2/v2/auth"
+    // It's acceptable to force-unwrap here because, for this call to fail we'd need a developer
+    // error, which we would catch because the unit tests would crash.
+    static var googleSignInBaseURL = URL(string: "https://accounts.google.com/o/oauth2/v2/auth")!
 
+    // TODO: This is incomplete
```
mokagio:
```suggestion
```

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -2,26 +2,50 @@ import Foundation
...
+            // The request we make with the SDK asks for:
+            //
+            // - email
+            // - profile
+            // - https://www.googleapis.com/auth/userinfo.email
+            // - https://www.googleapis.com/auth/userinfo.profile
+            // - openid
+            //
+            // See https://developers.google.com/identity/protocols/oauth2/scopes
+            ("scope", "https://www.googleapis.com/auth/userinfo.email")
```
mokagio, line 27: Spoiler: I updated this in my WIP branch but haven't seen any difference in the behavior, but that's likely due to my implementation being quite minimal at this point. Or, it's possible that only the backend would notice the difference, when completing the authentication using the OAuth token.

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -2,26 +2,50 @@ import Foundation
...
-            var components = URLComponents(string: baseURL)!
+            let baseURL = googleSignInBaseURL
+            guard var components = URLComponents(url: baseURL, resolvingAgainstBaseURL: false) else {
+                throw URLError(
+                    .unsupportedURL,
+                    userInfo: [
+                        NSLocalizedDescriptionKey: "Could not create `URLComponents` instance from \(baseURL)"
+                    ]
+                )
+            }
```
mokagio: Do you think this is acceptable as a `URLError`, or should I update it to another `OAuthError` `case` for consistency?

I'm leaning toward moving it to `OAuthError`.

WordPressAuthenticator/OAuth/OAuthTokenResponseBody.swift
```diff
@@ -0,0 +1,23 @@
...
+struct OAuthTokenResponseBody: Decodable {
+    let accessToken: String
+    let expiresIn: Int
+    /// This value is only returned if the request included an identity scope, such as openid, profile, or email.
+    /// The value is a JSON Web Token (JWT) that contains digitally signed identity information about the user.
+    let idToken: String?
+    let refreshToken: String?
+    let scope: String
+    /// The type of token returned. At this time, this field's value is always set to Bearer.
+    let tokenType: String
```
mokagio, line 13: The documentation comment I lifted from from the Google docs makes me wonder whether there's even a point in storing it at this time?

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -0,0 +1,41 @@
+// https://developers.google.com/identity/protocols/oauth2/native-app#step1-code-verifier
+//
+// FIXME: follow spec!
+//
+// A code_verifier is a high-entropy cryptographic random string using the unreserved
+// characters [A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~", with a minimum length of 43
+// characters and a maximum length of 128 characters.
```
mokagio, line 9: I plan to address the completion of this type in a dedicated PR. The process works for development purposes in the current, plain mode.

WordPressAuthenticatorTests/OAuth/ProofKeyForCodeExchangeTests.swift
```diff
@@ -0,0 +1,24 @@
...
+    func testCodeChallengeInPlainModeIsTheSameAsCodeVerifier() {
+        XCTAssertEqual(
+            ProofKeyForCodeExchange(codeVerifier: "abc", mode: .plain).codeCallenge,
+            "abc"
+        )
+    }
+
+    func testCodeChallengeInS256ModeIsEncodedAsPerSpec() {
+        // TODO:
+    }
```
mokagio, line 15: I put this down as a double reminder to finish the PKCE implementation. See also [previous comment](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/717/files#r1043267507).

crazytonyli:
@mokagio I'm okay with how ever you'd like to break down the end-to-end flow. I don't see any difficulties reviewing individual component, low-level or UI. 😄

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -2,26 +2,50 @@ import Foundation
...
 extension URL {
 
-    // TODO: This is incomplete
-    static func googleSignInAuthURL(clientId: String) throws -> URL {
-        let baseURL = "https://accounts.google.com/o/oauth2/v2/auth"
+    // It's acceptable to force-unwrap here because, for this call to fail we'd need a developer
+    // error, which we would catch because the unit tests would crash.
+    static var googleSignInBaseURL = URL(string: "https://accounts.google.com/o/oauth2/v2/auth")!
 
+    static func googleSignInAuthURL(clientId: String, pkce: ProofKeyForCodeExchange) throws -> URL {
```
crazytonyli, line 9: Since we are essentially implement OAuth 2, I think we can replace "Google Sign In" with a more generic term? Like, this function becomes `oauth2AuthenticationURL`, what do you think?

crazytonyli: I'm making assumption here, but I imagine the whole Google Sign In workflow can be used in "Sign In via Facebook", or even the WordPress OAuth2.

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
crazytonyli: I think it's safe to `!` here, since we already `!` the `googleSignInBaseURL`?

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -2,26 +2,50 @@ import Foundation
...
+                throw URLError(
+                    .unsupportedURL,
+                    userInfo: [
+                        NSLocalizedDescriptionKey: "Could not create `URLComponents` instance from \(baseURL)"
+                    ]
+                )
+            }
+
             components.queryItems = queryItems
             return try components.asURL()
```
crazytonyli: It's alsy safe to `!` here too?

Also, this `asURL` is an Alamofire function, maybe we shouldn't use it here since it throws an `AFError`?

WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
```diff
@@ -0,0 +1,46 @@
...
+            (CodingKeys.grantType.rawValue, grantType),
+            (CodingKeys.redirectURI.rawValue, redirectURI),
+        ]
+
+        let items = params.map { URLQueryItem(name: $0.0, value: $0.1) }
+
+        var components = URLComponents()
+        components.queryItems = items
+
+        guard let query = components.query else {
```
crazytonyli: I don't think it's possible to have a `nil` query in this case?

WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
```diff
@@ -0,0 +1,46 @@
...
+            throw OAuthError.failedToBuildURLQuery
+        }
+
+        guard let data = query.data(using: .utf8) else {
```
crazytonyli: I don't think it's possible to fail to utf8 encode the query string?

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+// The code verifier should have enough entropy to make it impractical to guess the value.
+struct ProofKeyForCodeExchange {
+
+    enum Mode {
```
crazytonyli: I'm not good at naming things, but what do you think about renaming this to `Method` and the `var method` to something like `var googleOAuthParameterName`?

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -0,0 +1,41 @@
...
+            case .plain: return "plain"
+            case .s256: return "S256"
+            }
+        }
+    }
+
+    let codeVerifier: String
+    let mode: Mode
+
+    init(codeVerifier: String, mode: Mode) {
```
crazytonyli: Considering we should randomly generate `codeVerifier`, do you think we can make this argument nil by default and generate a random string internally?

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
mokagio: Good point. I was thinking about this before going AFK and realized I may be using unclear boundaries here.

You might have noticed I made an `OAuth` folder and a `GoogleSignIn` one. My intent was to separate the components that are GoogleSignIn specific from those that are part of the OAuth spec. One issue with how I executed this is that I haven't been cross checking between the OAuth and Google documentations so there might be stuff in the OAuth folder that's still Google specific, and vice versa.

Thinking about it since, I came to the conclusion that I'd prefer to separate OAuth- from Google- specific code only once we'll need to add a new OAuth client (likely WordPress OAuth2). As such, I'm tempted to follow up this PR with one that moves all the `OAuth/` code into `GoogleSignIn/`, renaming some of the types if necessary.

My rationale for the suggestion above is that I'm more interested in having a working version of the SDK-less authentication than a general purpose OAuth client, even though building a client first might make it easier to add further authentication implementation in the future.

What do you think?

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
mokagio: That's a great observation! After all, what are the chances `URLComponents` will not initialize with a valid input URL?

mokagio: Addressed in d749c80ccc24d17e9c9040ec0fb1ee2ea4abfa24

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
mokagio: Another good point. Addressed in d5c6fd082497b9431cf9e71b716d18a7dae69eb7

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: Thank you for the suggestion. I'm keen on generating the string internally according to the spec but [I think that's work that could live in a dedicated PR](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/717/files#r1043267507). Does that sound okay?

What I have in mind is along the lines of your suggestion, with the difference of providing the random string as the default value.

```swift
init(codeVerifier: String = makeRandomPKCEString(), method: Method = .s256) {
```

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: Thank you for raising this. `Method` would definitely be a better name, as that's what [the spec](https://www.rfc-editor.org/rfc/rfc7636#page-17) uses 👍 

Addressed in 8abbb29ce3a4e795bcbb01890fb1826a7623e7f6.

As for the computed variable name, I went for `urlQueryParameterValue` to try and make it clearer where to use it.

WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
mokagio: Yes. That should be the case.

🤔 I guess I'm just uncomfortable force unwrapping stuff, but it does make for simpler code in a case like this one.

I tried to write a test that resulted in this code throwing, but I wasn't able to. That's not to say that there is _no way_ for the failure to occur, of course. Still, I feel a bit more confident about removing the errors here.

mokagio: Addressed in 8b6f6a89dd39665b903b73871e6f5c327f30770b

WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
mokagio: Addressed in 8b6f6a89dd39665b903b73871e6f5c327f30770b while working on [your previous comment](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/717/files#r1045479156).

Thanks!

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
crazytonyli: Sounds good to me. We probably don't need to "pre-abstract" stuff if there is no immediate plan to support another OAuth provider (service? server?). I trust you'll strike the right balance 👍

crazytonyli (approved):

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
mokagio: > I trust you'll strike the right balance 👍

Thanks for the confidence.

PR #720, opened 2023-01-06
Complete `CaptureLogs` conformance to `WordPressLoggingDelegate`

mokagio:
I had noticed this warning a while ago, but never made the time to address it.

The occasion came as part of The Code Quality Challenge day 3 "Get rid of a warning". See https://tuple.app/code-quality-challenge

---

### DIFF
--- a/WordPressAuthenticatorTests/Logging/LoggingTests.m
+++ b/WordPressAuthenticatorTests/Logging/LoggingTests.m
@@ -9,6 +9,10 @@
 
 @end
 
+// We are leaving some protocol methods intentionally unimplemented to then test that calling them
+// will not cause a crash.
+//
+// See https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/720#issuecomment-1374952619
 #pragma GCC diagnostic ignored "-Wprotocol"
 @implementation CaptureLogs
 
--- a/WordPressAuthenticatorTests/Logging/LoggingTests.m
+++ b/WordPressAuthenticatorTests/Logging/LoggingTests.m
@@ -13,7 +13,7 @@
 // will not cause a crash.
 //
 // See https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/720#issuecomment-1374952619
-#pragma GCC diagnostic ignored "-Wprotocol"
+#pragma clang diagnostic ignored "-Wprotocol"
 @implementation CaptureLogs
 
 - (instancetype)init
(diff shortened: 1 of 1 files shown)


crazytonyli:
I actually left those functions out intentionally, I should've added a comment about it though...

[The `testUnimplementedLoggingMethod` test](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/cb036688d89213b2464ecb8905a965a21e90e002/WordPressAuthenticatorTests/Logging/LoggingTests.m#L62) calls `LogVerbose`, whose corresponding logging function `logVerbose:` isn't implemented by the logger class. And this test function ensures calling an unimplemented logging function won't cause a crash.

Maybe we can `#pragma GCC diagnostic ignored "-Wprotocol"` to suppress the warning? What do you think?

WordPressAuthenticatorTests/Logging/LoggingTests.m
```diff
@@ -9,6 +9,11 @@ @interface CaptureLogs : NSObject<WordPressLoggingDelegate>
 
 @end
 
+// We are leaving some protocol methods intentionally unimplemented to then test that calling them
+// will not cause a crash.
+//
+// See https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/720#issuecomment-1374952619
+#pragma GCC diagnostic ignored "-Wprotocol"
```
crazytonyli: 🤦‍♂️ sorry, not sure why I wrote "gcc" in my previous comment. I think the right directive should be "clang".

Also, maybe we can add `#pragma clang diagnostic pop` and the end of the `CaptureLogs` implementation to stop ignoring the flag?

mokagio: No worries. I could have thought about it and checked.

However, that `#pragma` successfully disable the warning 🤔 

(I'm running out of time today so you might not see a commit to make the change after this comment)

crazytonyli: clang is compatible with gcc, so I guess it makes sense that the gcc directive works as well.

mokagio: Done. See [wordpress-mobile/WordPressAuthenticator-iOS@`8d7f99f` (#720)](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/720/commits/8d7f99f3b4aaa1c870395bcb1d268ebc63d5e1e6)

crazytonyli (approved):
Thanks for the fix! :shipit:

PR #724, opened 2023-01-18
Use configuration flag to log custom `step` event for `GetStartedViewController`

### LINKED ISSUE #723: Tracks: Add new configuration to log `enter_email_address` step
**Description**
Previously in #674 we introduced a new step `enter_email_address` to differentiate the start of the two flows: site credential login and WP.com login and use that to select the appropriate [help content for login failures](https://github.com/woocommerce/woocommerce-ios/blob/e52b46f43942aa61113584609830fbfd7a3a039d/WooCommerce/Classes/Model/CustomHelpCenterContent.swift#L34-L37) in WCiOS.

This has led to an issue with WPiOS Looker dashboard (ref: pe5sF9-FU-p2#comment-1542).

**Solution**
We need a new configuration and only track `enter_email_address` when that config is enabled. Otherwise, `start` should be use as before.

selanthiraiyan:
Closes #723 
### Description

Previously in #674 we introduced a new step value `enter_email_address` for `GetStartedViewController `. 
This was added to differentiate between `GetStartedViewController` and `SiteAddressViewController` in Woo mobile app. Those two screens had the same `flow` and `step` value.

This new step value change has led to an issue with the WPiOS Looker dashboard (ref: pe5sF9-FU-p2#comment-1542).
To fix the issue, we are introducing a configuration flag in `WordPressAuthenticatorConfiguration` to use custom step event only when the `useEnterEmailAddressAsStepValueForGetStartedVC` flag is enabled. 

### Testing steps

Follow the testing instructions from
1. https://github.com/woocommerce/woocommerce-ios/pull/8669
2. https://github.com/wordpress-mobile/WordPress-iOS/pull/19940

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ _None._
 ### Bug Fixes
 
 - Fix unresponsive issue in Onboading Questions screen. [#719]
+- Use configuration flag to log custom `step` event for `GetStartedViewController`. [#724]
 
 ### Internal Changes
 
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -22,7 +22,7 @@ PODS:
   - SVProgressHUD (2.2.5)
   - SwiftLint (0.49.1)
   - UIDeviceIdentifier (2.2.0)
-  - WordPressAuthenticator (5.0.0):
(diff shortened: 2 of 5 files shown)


itsmeichigo (approved):
LGTM 🚀

PR #725, opened 2023-01-18
New configuration for site address login only on the prologue screen

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/8679

## Description
This PR adds a new config to show only the site address login button on the prologue screen.

## Testing steps
Please follow the steps in https://github.com/woocommerce/woocommerce-ios/pull/8684.

---

### DIFF
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -271,7 +271,7 @@ class LoginPrologueViewController: LoginViewController {
                                                  style: primaryButtonStyle,
                                                  onTap: loginTapCallback())
         let enterYourSiteAddressButton = StackedButton(title: displayStrings.enterYourSiteAddressButtonTitle,
-                                                       isPrimary: configuration.enableSiteAddressLoginOnly,
+                                                       isPrimary: configuration.enableSiteAddressLoginOnlyInPrologue,
                                                        configureBodyFontForTitle: true,
                                                        accessibilityIdentifier: "Prologue Self Hosted Button",
                                                        style: secondaryButtonStyle,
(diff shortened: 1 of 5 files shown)


WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -288,18 +288,24 @@ class LoginPrologueViewController: LoginViewController {
...
                        enterYourSiteAddressButton,
                        createSiteButtonForBottomStackView]
         } else {
-            WPAuthenticatorLogError("Failed to create `StackedButton`s in login progue screen.")
+            WPAuthenticatorLogError("Failed to create `StackedButton`s in login prologue screen.")
             buttons = []
         }
 
-        let showDivider = configuration.enableWPComLoginOnlyInPrologue == false && configuration.enableSiteCreation == true
+        let showDivider = configuration.enableWPComLoginOnlyInPrologue == false &&
```
selanthiraiyan, line 306: Not related to this PR:

In a separate PR, I think we should remove this `enableWPComLoginOnlyInPrologue` flag as we no longer use that in Woo.

selanthiraiyan (approved):
LGTM 🚢

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
itsmeichigo: @selanthiraiyan  We also added a few other configs for the simplified login project that are no longer used: `enableSocialLogin`, `emphasizeEmailForWPComPassword`, `wpcomPasswordInstructions` - do you think we should remove them too? That would require reverting the logic changes too.

selanthiraiyan: I think yes. IMO, cleaning up the unused logic will help to maintain the project and make future changes.

itsmeichigo: Is there a safe way to test the removal to ensure it doesn't affect other host apps? We can't be sure that we are the only app that uses these configs.

PR #726, opened 2023-01-18
Run RuboCop via Danger in CI

mokagio:
~~Setup lifted from [WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/69ccc50a650b1794a25a7cd4308f89938f9d4696/.buildkite/pipeline.yml#L120-L126).~~

~~See it in action in #727:~~

**Update**: Working on this PR, I discovered we've been [running RuboCop via the CI-plugin `lint_pod`](https://github.com/Automattic/bash-cache-buildkite-plugin/blob/cd8b275e6c06fe51bffd6623d69605944671c4ed/bin/lint_pod#L9-L10) all along. Kudos @jkmassel

image

 ---


Gemfile
```diff
@@ -4,6 +4,8 @@ source 'https://rubygems.org'
 
 gem 'cocoapods', '~> 1.11'
 gem 'cocoapods-check', '~> 1.1'
+gem 'danger', '~> 8.6'
+gem 'danger-rubocop', '~> 0'
```
mokagio, line 8: The latest `danger-rubocpo` version is `0.10.0`. Using `~> 0` will resolve to the latest version `0.x` compatible with the repo.

Here's an example of the behavior in action using `~> 0` for `danger`:

image

PR #728, opened 2023-01-19
SDK-less GoogleSignIn - Part 3

mokagio:
This should be the last PR in the series that adds objects without using it. It turned out a bit long in term of SLOC, but a chunk of it is Xcode project cruft and tests boilerplate.

The core of this PR is providing an object responsible for sending a request to google for the token, `GoogleOAuthTokenGetter`. To make things neater, I also wrote a dedicated type for the Google client id `String` value.

## Previous PRs

- #710 
- #717 

## Next steps

- Merge the `OAuth` folder into `GoogleSignIn`
- Use the new types in the demo app to perform an existing user sign in
- Address the feedback in #714 and write UI tests
- Repeat but for the various other user scenarios—this is trickier because I don't yet know how to generate dummy Google account users with or without WordPress.com connection to test the flows

---


WordPressAuthenticator/GoogleSignIn/GoogleOAuthTokenGetter.swift
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+extension URLSession: DataGetting {
+
+    func data(for request: URLRequest) async throws -> Data {
+        let (data, _) = try await data(for: request)
+        return data
+    }
+}
```
mokagio: Note to self: Extract to dedicated file

WordPressAuthenticator/GoogleSignIn/DataGetting.swift
```diff
@@ -0,0 +1,4 @@
+protocol DataGetting {
+
+    func data(for request: URLRequest) async throws -> Data
+}
```
mokagio, line 4: Syntax sugar on top of `URLSession` that is also useful to create a test double for the unit tests.

WordPressAuthenticator/GoogleSignIn/GoogleClientId.swift
```diff
@@ -0,0 +1,26 @@
...
+            return root
+        }
+
+        return "\(root):/\(path)"
+    }
+
+    var defaultRedirectURI: String {
+        redirectURI(path: "oauth2callback")
+    }
+}
```
mokagio, line 26: I decided to create a dedicated type wrapping `String` for the Google Client Id value so that we could have validation and utility methods encapsulated in it.

What do you think of this approach? I might use it for other parts of this new functionality.

mokagio: By the way, @crazytonyli, I haven't forgotten your [suggestion](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/710#discussion_r1035358491) about using a different value for the redirect URI. But I'm keeping that as a later-on task to prioritize having something working end-to-end.

WordPressAuthenticator/GoogleSignIn/URLRequest+GoogleSignIn.swift
```diff
@@ -0,0 +1,17 @@
+extension URLRequest {
+
+    static func googleSignInTokenRequest(
```
mokagio, line 3: Below, I wrote:

```
// The concern with extending a Foundation type with special-purpose domain object getting ability
// is that it would pollute the namespace, offering the protocol methods as an option everywhere
// `URLSession` is used, even in part of the app that are unrelated with the resource.
```

I guess one could raise the same concern for this method on `URLRequest` 🤔 

One option could be to use a builder object `GoogleSignInRequestBuilder` `tokenRequest(body:) throws -> URLRequest`.

WordPressAuthenticator/OAuth/OAuthTokenResponseBody.swift
```diff
@@ -1,7 +1,7 @@
 /// Models the response to an OAuth token request.
 ///
 /// - Note: See documentation at https://developers.google.com/identity/protocols/oauth2/native-app#exchange-authorization-code
-struct OAuthTokenResponseBody: Decodable {
+struct OAuthTokenResponseBody: Codable, Equatable {
```
mokagio, line 4: I don't like that this is `Codable` and `Equatable` in the production code, when those capabilities are only needed in the tests. However, this is the only way to get the compiler to do the work of conforming to the protocols for us.

WordPressAuthenticatorTests/GoogleSignIn/GoogleOAuthTokenGettingStub.swift
```diff
@@ -0,0 +1,25 @@
...
+        self.result = result
+    }
+
+    func getToken(clientId: GoogleClientId, authCode: String, pkce: ProofKeyForCodeExchange) async throws -> OAuthTokenResponseBody {
+        switch result {
+        case .success(let response): return response
+        case .failure(let error): throw error
+        }
+    }
+}
```
mokagio, line 25: This test double is quite similar to the one for `DataGetting`. Both use a `Result<T, Error>` under the hood to stub a `throws -> T` method. 

If I'll need to create another `throws -> T` types, I might spend some time to write an abstraction to use to create more of these.

crazytonyli (approved):

PR #729, opened 2023-01-19
Depend on WordPressKit ~> 6.0

crazytonyli:
Bump the WordPressKit dependency from v5 to v6, so that it can be upgraded on WordPress-iOS (which I have [a draft PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/19908) for).

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #731, opened 2023-01-25
Update the version in the podspec file

crazytonyli:
I missed this update during merging the latest changes in PR #729

---

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -22,7 +22,7 @@ PODS:
   - SVProgressHUD (2.2.5)
   - SwiftLint (0.49.1)
   - UIDeviceIdentifier (2.2.0)
-  - WordPressAuthenticator (5.1.0):
+  - WordPressAuthenticator (5.2.0):
     - GoogleSignIn (~> 6.0.1)
     - Gridicons (~> 1.0)
     - "NSURL+IDN (= 0.4)"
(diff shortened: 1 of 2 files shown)


mokagio (approved):
> I missed this update during merging the latest changes in PR https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/729

You are not alone! This gets me pretty much all the times. We ought to implement a CI check for this. `if <get tag from Buildkite environment> != <parse version from podspec> then fail`.

crazytonyli:
@mokagio It's a good idea to prevent this "version mismatch" issue in CI 👍 .

I think `pod publish` command already does the check you mentioned, see [this CI job](https://buildkite.com/automattic/wordpress-authenticator-ios/builds/572#0185e66b-52c6-4d83-ade3-5c553b736774) output:
> Tag 5.2.0 does not match version 5.1.0 from WordPressAuthenticator.podspec.

Since the "publishing pod" step only runs when a new git tag is created, we probably can bring this check forward to the PR stage, by checking if the latest version in CHANGELOG.md matches the version in podspec.

mokagio:
> Since the "publishing pod" step only runs when a new git tag is created, we probably can bring this check forward to the PR stage, by checking if the latest version in CHANGELOG.md matches the version in podspec.

Right! Maybe we can extract the `CHANGELOG.md` parsing logic from release-toolkit for this purpose. That's something we'd want to do anyway to automate creating releases / release notes.

PR #735, opened 2023-01-30
SDK-less Google SignIn part 4

mokagio:
Just a few standalone changes and additions to the components for the Google SignIn SDK-less flow. 

See also #728 which also includes a list of next step items. 

---

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
@@ -55,6 +55,8 @@ public struct WordPressAuthenticatorConfiguration {
         return clientId
     }
 
+    let googleLoginWithoutSDK: Bool
+
     /// UserAgent
     ///
     let userAgent: String
@@ -200,7 +202,9 @@ public struct WordPressAuthenticatorConfiguration {
                  skipXMLRPCCheckForSiteAddressLogin: Bool = false,
                  enableManualSiteCredentialLogin: Bool = false,
                  useEnterEmailAddressAsStepValueForGetStartedVC: Bool = false,
-                 enableSiteAddressLoginOnlyInPrologue: Bool = false) {
+                 enableSiteAddressLoginOnlyInPrologue: Bool = false,
+                 googleLoginWithoutSDK: Bool = false
+    ) {
 
         self.wpcomClientId = wpcomClientId
         self.wpcomSecret = wpcomSecret
@@ -212,6 +216,7 @@ public struct WordPressAuthenticatorConfiguration {
         self.googleLoginClientId =  googleLoginClientId
         self.googleLoginServerClientId = googleLoginServerClientId
         self.googleLoginScheme = googleLoginScheme
+        self.googleLoginWithoutSDK = googleLoginWithoutSDK
         self.userAgent = userAgent
         self.showLoginOptions = showLoginOptions
         self.enableSignUp = enableSignUp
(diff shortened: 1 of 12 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
```diff
@@ -47,6 +47,14 @@ public struct WordPressAuthenticatorConfiguration {
     ///
     let googleLoginScheme: String
 
+    internal var googleClientId: GoogleClientId {
+        guard let clientId = GoogleClientId(string: googleLoginClientId) else {
+            fatalError("Could not init GoogleClientId from developer provided value.")
+        }
+
+        return clientId
+    }
```
mokagio, line 56: This way, we won't need instantiate a `GoogleClientId` when passing the configuration options to the new SDK-less flow. It will keep the code more concise.

mokagio: I'm not sure why it's explicitly `internal`, given that's the default...

WordPressAuthenticator/GoogleSignIn/GoogleOAuthTokenGetter.swift
```diff
@@ -8,11 +8,17 @@ class GoogleOAuthTokenGetter: GoogleOAuthTokenGetting {
 
     func getToken(
         clientId: GoogleClientId,
+        audience: String,
```
mokagio, line 11: This is not in the spec at https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server but we'll get an idToken that our backend considers invalid if omitted.

It is, however, in another piece of documentation that I didn't think of writing down. Interestingly, wherever I found out about this used the name "audience" but in our existing code the value assigned to the SDK equivalent is `googleLoginServerClientId`.

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -19,17 +19,8 @@ extension URL {
...
-            //
-            // - email
-            // - profile
-            // - https://www.googleapis.com/auth/userinfo.email
-            // - https://www.googleapis.com/auth/userinfo.profile
-            // - openid
-            //
             // See https://developers.google.com/identity/protocols/oauth2/scopes
-            ("scope", "https://www.googleapis.com/auth/userinfo.email")
+            ("scope", "openid https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile")
```
mokagio, line 23: I haven't tried it yet, but I think we could remove `https://www.googleapis.com/auth/userinfo.profile`. The only information we need other than the id token from Google is the email adderss.

crazytonyli: It appears [GoogleSignIn SDK uses "email profile" as the scopes](https://github.com/google/GoogleSignIn-iOS/blob/7.0.0/GoogleSignIn/Sources/GIDScopes.m#L58-L61).

crazytonyli (approved):

WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
mokagio: Great find, thanks!

The code also [suggests](https://github.com/google/GoogleSignIn-iOS/blob/7932d33686c1dc4d7df7a919aae47361d1cdfda4/GoogleSignIn/Sources/GIDScopes.m?rgh-link-date=2023-02-07T22%3A08%3A56Z#L19-L22) that the "https://..." scopes are no longer needed as "old".

I'll try this in a PR later on.

PR #736, opened 2023-02-01
Add new config to remove XMLRPC check for site address login

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/8770.

**Description**
This PR adds a new config to remove XMLRPC checks for site address login. When this is enabled, the library will trigger a new delegate method `handleSiteCredentialLogin`, so that the host app will need to handle the login with the input credentials. This allows the login to be handled without the need of using XMLRPC API.

**Testing steps**
Please follow https://github.com/woocommerce/woocommerce-ios/pull/8790 to test this new change.

---

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
@@ -142,7 +142,7 @@ public struct WordPressAuthenticatorConfiguration {
 
     /// If enabled, the library will trigger the delegate method `handleSiteCredentialLogin`
     /// instead of using the XMLRPC API for handling site credential login.
-    let manualSiteCredentialLogin: Bool
+    let enableManualSiteCredentialLogin: Bool
 
     /// Used to determine the `step` value for `unified_login_step` analytics event in `GetStartedViewController`
     ///
@@ -190,7 +190,7 @@ public struct WordPressAuthenticatorConfiguration {
                  wpcomPasswordInstructions: String? = nil,
                  skipXMLRPCCheckForSiteDiscovery: Bool = false,
                  skipXMLRPCCheckForSiteAddressLogin: Bool = false,
-                 manualSiteCredentialLogin: Bool = false,
+                 enableManualSiteCredentialLogin: Bool = false,
                  useEnterEmailAddressAsStepValueForGetStartedVC: Bool = false,
                  enableSiteAddressLoginOnlyInPrologue: Bool = false) {
 
@@ -224,7 +224,7 @@ public struct WordPressAuthenticatorConfiguration {
         self.wpcomPasswordInstructions = wpcomPasswordInstructions
         self.skipXMLRPCCheckForSiteDiscovery = skipXMLRPCCheckForSiteDiscovery
         self.skipXMLRPCCheckForSiteAddressLogin = skipXMLRPCCheckForSiteAddressLogin
-        self.manualSiteCredentialLogin = manualSiteCredentialLogin
+        self.enableManualSiteCredentialLogin = enableManualSiteCredentialLogin
         self.useEnterEmailAddressAsStepValueForGetStartedVC = useEnterEmailAddressAsStepValueForGetStartedVC
         self.enableSiteAddressLoginOnlyInPrologue = enableSiteAddressLoginOnlyInPrologue
     }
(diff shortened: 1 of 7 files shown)


WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -444,7 +444,8 @@ private extension SiteAddressViewController {
         checkSiteExistence(url: url) { [weak self] in
             guard let self = self else { return }
             // skips XMLRPC check for site discovery if needed
```
selanthiraiyan: Nit: Should we update this comment to add that we skip the check for site address login case as well?

WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
```diff
@@ -136,6 +136,12 @@ public struct WordPressAuthenticatorConfiguration {
     ///
     let skipXMLRPCCheckForSiteDiscovery: Bool
 
+    /// If enabled, site address login will not check for XMLRPC URL or use XMLRPC API to handle login.
+    /// Important: make sure to implement the delegate method `handleSiteCredentialLogin` when enabling this config.
+    ///
+    ///
+    let skipXMLRPCCheckForSiteAddressLogin: Bool
```
selanthiraiyan, line 141: Nit:   

This changes two things.

1. After entering the site address, we skip the XMLRPC check
1. After entering the site credentials, we skip the whole login process and delegate the work to the client app.

WDYT think about naming it as `delegateSiteAddressLogin` or `shouldDelegateSiteAddressLogin`? 

IMHO, if we use `skipXMLRPCCheck...` pattern, it might convey that we are only skipping XMLRPC check just like `skipXMLRPCCheckForSiteDiscovery`.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
```diff
@@ -515,7 +556,7 @@ extension SiteCredentialsViewController {
     override func displayRemoteError(_ error: Error) {
         configureViewLoading(false)
         let err = error as NSError
-        if err.code == 403 {
+        if err.code == 403 || err.code == 401 {
```
selanthiraiyan: Nit: Would it help to add a comment to explain these error codes? Or moving into a constants enum would be even better. 😅

selanthiraiyan (approved):
LGTM 🚢 

I left a few non-blocking nits.

WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
itsmeichigo: Good point! I added an extra config for manual site address login in ff454bb, let me know what you think.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
itsmeichigo: Actually I think it makes more sense to let the host app decide whether to display the wrong credential alert. Updated in d094005.

WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
selanthiraiyan: Looks great. 👍

WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
```diff
@@ -136,6 +136,14 @@ public struct WordPressAuthenticatorConfiguration {
     ///
     let skipXMLRPCCheckForSiteDiscovery: Bool
 
+    /// If enabled, site address login will not check for XMLRPC URL.
+    ///
+    let skipXMLRPCCheckForSiteAddressLogin: Bool
+
+    /// If enabled, the library will trigger the delegate method `handleSiteCredentialLogin`
+    /// instead of using the XMLRPC API for handling site credential login.
+    let manualSiteCredentialLogin: Bool
```
selanthiraiyan: Nit: To follow the other names from this file, can we prefix `enable`? i.e. `enableManualSiteCredentialLogin`?

selanthiraiyan (approved):
Thank you for the updates. 🚀

WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
itsmeichigo: Updated in c4d2dcf60d470bb407fc07e16dd8cebd6c16eb4f.

PR #737, opened 2023-02-03
Disable credential picker for site address login only

itsmeichigo:
Related to https://github.com/woocommerce/woocommerce-ios/issues/8816.

**Description**
In WooCommerce, when the config `enableSiteAddressLoginOnlyInPrologue` is enabled, we allow users to log in only with site address. If the credentials are available for WPCom in Keychain, we'll let them login in with WPCom automatically.

To avoid this undefined behavior, this PR disables that case by not presenting the credential picker if  `enableSiteAddressLoginOnlyInPrologue` is on.

**Testing steps**
Please follow the steps in https://github.com/woocommerce/woocommerce-ios/pull/8817.

---

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -22,7 +22,7 @@ PODS:
   - SVProgressHUD (2.2.5)
   - SwiftLint (0.49.1)
   - UIDeviceIdentifier (2.2.0)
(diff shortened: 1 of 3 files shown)


jaclync (approved):
Works as expected :shipit:

PR #739, opened 2023-02-06
Add type modeling Google SignIn "ID token"

mokagio:
The OAuth process with Google returns, among other info, an [`id_token` field](https://developers.google.com/identity/openid-connect/openid-connect#obtainuserinfo). This is a JSON web token (JWT) with in its payload the information accessible by the `scope` request parameter. In our case, what we want is only the `email`.

The next step will be to return this type from the new flow in `GoogleAuthenticator`, so that `email` can be passed together with the OAuth token to the WordPress.com backend.

---

### DIFF
--- a/WordPressAuthenticator/GoogleSignIn/Data+Base64URL.swift
+++ b/WordPressAuthenticator/GoogleSignIn/Data+Base64URL.swift
@@ -1,0 +1,24 @@
+extension Data {
+
+    /// "base64url" is an encoding that is safe to use with URLs.
+    /// It is defined in RFC 4648, section 5.
+    ///
+    /// See:
+    /// - https://tools.ietf.org/html/rfc4648#section-5
+    /// - https://tools.ietf.org/html/rfc7515#appendix-C
+    init(base64URLDecode: String) {
+        let base64 = base64URLDecode
+            .replacingOccurrences(of: "-", with: "+")
+            .replacingOccurrences(of: "_", with: "/")
+
+        let length = Double(base64.lengthOfBytes(using: String.Encoding.utf8))
+        let requiredLength = 4 * ceil(length / 4.0)
+        let paddingLength = requiredLength - length
+        if paddingLength > 0 {
+            let padding = "".padding(toLength: Int(paddingLength), withPad: "=", startingAt: 0)
+            self.init(base64Encoded: base64 + padding, options: .ignoreUnknownCharacters)!
+        } else {
+            self.init(base64Encoded: base64, options: .ignoreUnknownCharacters)!
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


WordPressAuthenticator/GoogleSignIn/Data+Base64URL.swift
```diff
@@ -0,0 +1,24 @@
+extension Data {
+
+    /// "base64url" is an encoding that is safe to use with URLs.
+    /// It is defined in RFC 4648, section 5.
+    ///
+    /// See:
+    /// - https://tools.ietf.org/html/rfc4648#section-5
+    /// - https://tools.ietf.org/html/rfc7515#appendix-C
+    init?(base64URLEncoded: String) {
```
mokagio, line 9: We need this type of decoding because of JWT. 

I considered making the `init` `throw`, but decided against because:

1. Existing `Data` initializers, such as `Data(base64Encoded:, options:)` are `init?`
2. I don't think we need a refined error granularity at this level.

WordPressAuthenticator/GoogleSignIn/JWToken.swift
```diff
@@ -0,0 +1,47 @@
+/// Represents a JSON Web Token (JWT)
+///
+/// See https://jwt.io/introduction
+struct JWToken {
```
crazytonyli: What do you think about using the full name as the type name here, like `JSONWebToken`?

WordPressAuthenticator/GoogleSignIn/JWToken.swift
```diff
@@ -0,0 +1,47 @@
...
+    let encodedValue: String
```
crazytonyli: A minor suggestion: Swift uses the term `rawValue` in lots of its APIs, I feel like we can use the same variable name here?

WordPressAuthenticator/GoogleSignIn/IDToken.swift
```diff
@@ -0,0 +1,16 @@
+/// See https://developers.google.com/identity/openid-connect/openid-connect#obtainuserinfo
+public struct IDToken {
+
+    public let token: String
```
crazytonyli: My preference is using a strong type if possible, like using `JWToken` instead of `String` here. So that the call site can be sure it's a valid ID token, also makes it impossible to create a `IDToken` instance without a valid JWT.

crazytonyli (approved):

WordPressAuthenticator/GoogleSignIn/JWToken.swift
mokagio: Great suggestion. It's a bit inconsistent as J (short) W (short) Token (long).

The alternative would be `JWT`, similar to `URL`. But I've gone with the more verbose option you suggested in 2e99cef925a2c6aa42ff13db37e205ab2a704c1b as it seems clearer.

WordPressAuthenticator/GoogleSignIn/JWToken.swift
mokagio: Sounds good. 

Addressed in 99ac474f5cb293168ff256d43fdc850ced46db53

WordPressAuthenticator/GoogleSignIn/IDToken.swift
mokagio: 💯 can't believe I missed this. Glad you spotted it.

Addressed in de81593d4407f9cf0bbbac387741dfc981b58b5d

PR #740, opened 2023-02-13
Implement PKCE S256 code challenge mode

mokagio:
Follows up on a few `TODO` and `FIXME` to implement the secure and spec-compliant PKCE hashing.

---


WordPressAuthenticator/GoogleSignIn/Data+Base64URL.swift
```diff
@@ -21,4 +21,12 @@ extension Data {
             self.init(base64Encoded: base64, options: .ignoreUnknownCharacters)
         }
     }
+
+    // Funcion name to match the standar library's `base64EncodedString()`
```
mokagio: Note to self: Add link to spec.

WordPressAuthenticator/GoogleSignIn/Data+SHA256.swift
```diff
@@ -0,0 +1,35 @@
...
+
+extension Data {
+
+    func sha256Hashed() -> Data {
+        Data(sha256Hash())
+    }
+
+    func sha256Hashed() -> String {
+        sha256Hash().map { String(format: "%02x", $0) }.joined(separator: "")
+    }
```
mokagio: I like the overload here, but it comes with the downside that some usages might require the type to be explicitly specified, `let x: String = data.sha256Hashed()` vs `let x = data.sha256Hashed()`.

The reason I needed both `Data` and `String` was for ease of testing.

WordPressAuthenticator/GoogleSignIn/Data+SHA256.swift
```diff
@@ -0,0 +1,35 @@
...
+// class Tests: XCTestCase {
+
+//     func testHashing() {
+//         // test sha256Hashed()
+//         XCTAssertEqual(
+//             "foo".data(using: .utf8)!.sha256Hashed(),
+//             "2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae".data(using: .utf8)!
+//         )
+//     }
+// }
```
mokagio: Wooops. Remove this.

WordPressAuthenticator/GoogleSignIn/Data+Base64URL.swift
mokagio:
```suggestion
    /// See https://tools.ietf.org/html/rfc4648#section-5
    ///    
    /// Function name to match the standard library's `base64EncodedString()`.
```

WordPressAuthenticator/GoogleSignIn/String+SHA256.swift
```diff
@@ -0,0 +1,13 @@
...
+
+    var sha256Hashed: String {
+        let data = Data(self.utf8)
+        var hash = [UInt8](repeating: 0, count: Int(CC_SHA256_DIGEST_LENGTH))
+        data.withUnsafeBytes {
+            _ = CC_SHA256($0.baseAddress, CC_LONG(data.count), &hash)
+        }
+        return hash.map { String(format: "%02x", $0) }.joined()
+    }
+}
```
mokagio: This is no longer needed. Its tests should also be removed.

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -25,21 +25,66 @@ struct ProofKeyForCodeExchange {
...
         case .s256:
-            // TODO: code_challenge = BASE64URL-ENCODE(SHA256(ASCII(code_verifier)))
-            fatalError()
+            // The spec defines code_challenge for the s256 mode as:
+            //
+            // code_challenge = BASE64URL-ENCODE(SHA256(ASCII(code_verifier)))
+            //
+            // We don't need the ASCII conversion, because we build `CodeVerifier` from URL safe
+            // characters.
+            let rawData = codeVerifier.rawValue.data(using: .utf8)!
```
mokagio: Is it safe to force unwrap?

WordPressAuthenticatorTests/OAuth/ProofKeyForCodeExchangeTests.swift
```diff
@@ -4,14 +4,22 @@ import XCTest
...
         XCTAssertEqual(
-            ProofKeyForCodeExchange(codeVerifier: "abc", method: .plain).codeCallenge,
-            "abc"
+            ProofKeyForCodeExchange(codeVerifier: codeVerifier, method: .plain).codeCallenge,
+            codeVerifier.rawValue
         )
     }
 
     func testCodeChallengeInS256ModeIsEncodedAsPerSpec() {
         // TODO:
```
mokagio: Remove this `TODO`!

WordPressAuthenticator/GoogleSignIn/Data+SHA256.swift
jkmassel: I've found the compiler generally can figure it out these days without too much `couldn't resolve this in reasonable time` – I like this overload pattern too!

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
jkmassel: You shouldn't need to!

```suggestion
            let rawData = Data(codeVerifier.rawValue.utf8)
```

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -25,21 +25,72 @@ struct ProofKeyForCodeExchange {
...
+
+        let rawValue: String
+
+        // From the docs: using the unreserved characters [A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~"
+        // That is, URL safe characters.
+        //
+        // Notice that Swift offers `CharacterSet.urlQueryAllowed` to represent this set of characters.
+        // However, there is no straightforward way to convert a `CharacterSet` to a `Set<Character>`.
+        // See for example https://nshipster.com/characterset/.
+        private let allowedCharacters = Set("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._~")
```
jkmassel: Would it be work yoinking this into an extension somewhere?

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -25,21 +25,72 @@ struct ProofKeyForCodeExchange {
...
+    }
+}
+
+// This is a helper for the tests.
+//
+// Unfortunately, it needs to be part of the production code because Swift doesn't allow adding
+// non-convenience initializers outside the module.
+extension ProofKeyForCodeExchange.CodeVerifier {
+
+    init?(value: String) {
```
jkmassel: I wonder if it'd be worth putting this in the body of the type, and _possibly_ as the default initializer? Then your "make new PKCE" method would be like `ProofKeyForCodeExchange.create(withLength:)`? 

The reason I ask is that do we know for sure that we'll never need to parse an existing one? I've found that an initializer than only ever creates a new value but can't process an existing one eventually gets me in trouble 😅

WordPressAuthenticatorTests/GoogleSignIn/CodeVerifierTests.swift
```diff
@@ -0,0 +1,46 @@
+@testable import WordPressAuthenticator
+import XCTest
+
+class CodeVerifierTests: XCTestCase {
+
+    func testCodeVerifierIsRandomString() {
+        XCTAssertNotEqual(
+            ProofKeyForCodeExchange.CodeVerifier().rawValue,
```
jkmassel: nit – could we make this `equatable` so that `rawValue` doesn't need to be exposed?

Or is that too much window dressing?

WordPressAuthenticatorTests/GoogleSignIn/Data+Base64URLTests.swift
```diff
@@ -37,4 +39,45 @@ class DataBase64URLDecoding: XCTestCase {
             Data()
         )
     }
+
+    // MARK: - Encoding
+
+    func testBase64URLEncoding() {
+        XCTAssertEqual(
+            "hello world".data(using: .utf8)!.base64URLEncodedString(),
```
jkmassel: Same note about string -> data here, unless you want to deal with optionality, in which case I'd suggest `XCTUnwrap`?

WordPressAuthenticatorTests/GoogleSignIn/Data+Base64URLTests.swift
```diff
@@ -37,4 +39,45 @@ class DataBase64URLDecoding: XCTestCase {
...
+        )
+    }
+
+    // Verify against non UTF-8 characters to ensure it's okay to force unwrapping internally,
+    // as suggested by the following discussion in the forums
+    // https://forums.swift.org/t/can-encoding-string-to-data-with-utf8-fail/22437/4
+    func testBase64URLEncodingNonUTF8() {
+        // The Chinese character for "world" cannot be represented in the UTF-8 encoding.
+        XCTAssertNotNil(
+            "Hello, 世界".data(using: .utf8)!.base64URLEncodedString(),
```
jkmassel: Same note about string -> data here

WordPressAuthenticatorTests/GoogleSignIn/Data+Base64URLTests.swift
```diff
@@ -37,4 +39,45 @@ class DataBase64URLDecoding: XCTestCase {
...
+
+        // Notice that if you paste the input strings in the following examples
+        // in an online encoder, it'll return a different value. I think that
+        // has to do with the kind of UTF-8 allowances `Data` makes internally.
+        // For the scope in which we expect to use this code, i.e. only UTF-8
+        // encodable characters, that doesn't matter.
+
+        // \u{FF}, Unicode scalar 0xFF, is not a valid UTF-8
+        XCTAssertEqual(
+            "Hello, \u{FF}".data(using: .utf8)!.base64URLEncodedString(),
```
jkmassel: Same note about string -> data here

WordPressAuthenticatorTests/GoogleSignIn/Data+Base64URLTests.swift
```diff
@@ -37,4 +39,45 @@ class DataBase64URLDecoding: XCTestCase {
...
+            "SGVsbG8sIMO_"
+        )
+        // \0, null character, is not a valid UTF-8
+        XCTAssertEqual(
+            "Hello, \0 World!".data(using: .utf8)!.base64URLEncodedString(),
```
jkmassel: Same note about string -> data here

WordPressAuthenticatorTests/GoogleSignIn/Data+SHA256Tests.swift
```diff
@@ -0,0 +1,22 @@
+@testable import WordPressAuthenticator
+import XCTest
+
+class DataSHA256Tests: XCTestCase {
+
+    func testSHA256Hasing() {
+        // It's fiddly to test the hashing `Data` against `Data`, so we use `String` against
+        // `String` and we trust/hope that the underlying implementation is shared.
+        XCTAssertEqual(
+            "foo".data(using: .utf8)!.sha256Hashed(),
```
jkmassel: Same note about string -> data here

WordPressAuthenticatorTests/GoogleSignIn/Data+SHA256Tests.swift
```diff
@@ -0,0 +1,22 @@
...
+            "2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae"
+        )
+        XCTAssertEqual(
+            "Hello, World!".data(using: .utf8)!.sha256Hashed(),
```
jkmassel: Same note about string -> data here

WordPressAuthenticatorTests/GoogleSignIn/Data+SHA256Tests.swift
```diff
@@ -0,0 +1,22 @@
...
+            "dffd6021bb2bd5b0af676290809ec3a53191dd81c7f70a4b28688a362182986f"
+        )
+        XCTAssertEqual(
+            "abcABC-=/?".data(using: .utf8)!.sha256Hashed(),
```
jkmassel: Same note about string -> data here

jkmassel (changes requested):
Lotta nits, but good to go after so I'm approving to unblock

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: Thank you for the suggestion. Addressed in e23de84e

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: > I wonder if it'd be worth putting this in the body of the type,

See 0361bb24

mokagio: > I've found that an initializer than only ever creates a new value but can't process an existing one eventually gets me in trouble 😅

Ah! Thank you fort the suggestion. I don't think I wrote it but I didn't like the state of initialization either, but at the time I decided to ship as it was rather than spending extra time on it.

You suggestion was just what I needed to come up with the mix of idea and motive to improve the implementation.

See ff9f23d5

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: 😍 

Thanks!

See 1adc8564

WordPressAuthenticatorTests/GoogleSignIn/Data+Base64URLTests.swift
mokagio: Excellent, thanks!

See caa86091

WordPressAuthenticatorTests/GoogleSignIn/CodeVerifierTests.swift
mokagio: I don't think it's window dressing. 

See 07a3b2b6

The compiler does all the work and the resulting code is more compact. Win for the tests and win for future consumers of the production code.

mokagio:
@jkmassel thank you for the great suggestions.

From your message, I'm guessing the intent of the review was not to request changes and I feel I could admin merge this.

image

However, I would have asked you kindly for another look given the number of changes I introduced.

WordPressAuthenticatorTests/GoogleSignIn/Character+URLSafeTests.swift
```diff
@@ -0,0 +1,15 @@
+@testable import WordPressAuthenticator
+import Foundation
+import XCTest
+
+class Character_URLSafeTests: XCTestCase {
+
+    func testURLSafeCharacters() {
```
jkmassel: nit: this took me a minute or two to grok – the combination of `XCTAssertFalse` and `.inverted` messed with my brain – basically we're saying "everything in `Character.urlSafeCharacters` should also be in `CharacterSet.urlQueryAllowed`", right?

If so, could we `XCTAssertTrue(CharacterSet.urlQueryAllowed.contains(character.unicodeScalars.first!))`? Or does that not properly capture what we're after? 

Also, can we `XCTUnwrap` `character.unicodeScalars.first!` to avoid the tests crashing?

WordPressAuthenticator/GoogleSignIn/Data+SHA256.swift
```diff
@@ -0,0 +1,20 @@
...
+        Data(sha256Hash())
+    }
+
+    func sha256Hashed() -> String {
+        sha256Hash().map { String(format: "%02x", $0) }.joined(separator: "")
+    }
+
+    private func sha256Hash() -> [UInt8] {
+        var hash = [UInt8](repeating: 0, count: Int(CC_SHA256_DIGEST_LENGTH))
+        withUnsafeBytes {
```
jkmassel: This is probably fine, but anytime I see `unsafe` in the code I like to think about if there's a better way.

[CryptoKit has a `SHA256` implementation](https://developer.apple.com/documentation/cryptokit/sha256) (that probably just does this under the hood 😂), but that wouldn't require us to use any `unsafe` code. I believe it's within our minimum iOS version so we should be ok to use it. WDYT?

WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
```diff
@@ -16,7 +16,7 @@ extension OAuthTokenRequestBody {
             clientSecret: "",
             audience: audience,
             code: authCode,
-            codeVerifier: pkce.codeVerifier,
+            codeVerifier: pkce.codeVerifier.rawValue,
```
jkmassel, line 19: nit: Could this be opaque, and only require passing the `pkce.codeVerifier`, and not having to see the `.rawValue`?

WordPressAuthenticator/GoogleSignIn/String+Random.swift
```diff
@@ -0,0 +1,12 @@
+extension String {
+
+    static func randomString(using characters: Set<Character>, withLength length: Int) -> String {
+        let allowedCharactersCount = UInt32(characters.count)
+
+        return (0..<length).reduce("") { accumulator, _ in
+           let randomOffset = Int(arc4random_uniform(allowedCharactersCount))
+           let randomCharacter = characters[characters.index(characters.startIndex, offsetBy: randomOffset)]
+           return accumulator + String(randomCharacter)
+        }
```
jkmassel: Nit, but we can probably golf away `arc4random_uniform` in favour of the Swift standard library _if this doesn't need to be cryptographically secure_. If it _does need_ to be cryptographically secure (which I think is the case because it's called in the PKCE implementation) , I wonder if we should name the method to `randomSecureString` or similar?

Assuming it _does_ need to be cryptographically secure, WDYT about using `SecRandomCopyBytes` to do this in O(1) time instead of O(n)?

```suggestion
       (0..<length).reduce("") { accumulator, _ in
           let randomOffset = Int.random(in: 0...characters.count)
           let randomCharacter = characters[characters.index(characters.startIndex, offsetBy: randomOffset)]
           return accumulator + String(randomCharacter)
        }
```

jkmassel (dismissed):
Ok, so I hit `Request Changes` last time by accident, which I feel bad about, but this time I really do think there are a few more things we could do here – keen to hear your thoughts!

WordPressAuthenticatorTests/GoogleSignIn/Character+URLSafeTests.swift
mokagio: > Also, can we XCTUnwrap character.unicodeScalars.first! to avoid the tests crashing?

Done. See ca5d0a00.

I was at peace with force unwrapping there because, given the static nature of the inputs, `unicodeScalars` would "never™" have been empty. Still, I do appreciate the consistency of handling the possibility just in case.

mokagio: @jkmassel 

> basically we're saying "everything in `Character.urlSafeCharacters` should also be in `CharacterSet.urlQueryAllowed`", right?
>
> If so, could we `XCTAssertTrue(CharacterSet.urlQueryAllowed.contains(character.unicodeScalars.first!))`? Or does that not properly capture what we're after?

Yes, what I'm after here is to make sure that the two collections map 1:1.

I went for the `inverted` approach to cover the case were `CharacterSet.urlQueryAllowed` is a subset of `Character.urlSafeCharacters`. But with fresh eyes I now see that the straightforward approach you suggest has the same result. 😅 👍

mokagio: See a3af7d4f

mokagio: Actually, I just realized that the test can be made simpler and maybe clearer by using `SetAlgebra`.

See 811955fc

WordPressAuthenticator/GoogleSignIn/Data+SHA256.swift
mokagio: 🤦‍♂️ an to think CryptoKit had been available all along...

Great idea, thanks. The implementation also ends up more straightforward.

See 3051ce42

WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
mokagio: Yes, that would be a neat refinement.

WordPressAuthenticator/GoogleSignIn/String+Random.swift
mokagio: @jkmassel 

>  If it _does need_ to be cryptographically secure (which I think is the case because it's called in the PKCE implementation)

This is what the [spec](https://www.rfc-editor.org/rfc/rfc7636#section-4.1) says:

> code_verifier = high-entropy cryptographic random STRING using the unreserved characters [A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~"  from [Section 2.3 of [RFC3986]](https://www.rfc-editor.org/rfc/rfc3986#section-2.3), with a minimum length of 43 characters and a maximum length of 128 characters.

"cryptographic random STRING" makes me think it should be cryptographically secure.

mokagio: I rewrote to use `SecRandomCopyBytes` in 4629e575. However, I'm not sure how to from O(n) to O(1) without losing the part of code that picks characters from the allowed set.

WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
mokagio: > Yes, that would be a neat refinement.

...which I might tackle in a dedicated PR because it would require changing the shape of a few types. The `codeVerifier` input in the request object currently maps to the `code_verifier` JSON string. Changing that will require some plumbing.

WordPressAuthenticator/GoogleSignIn/String+Random.swift
mokagio: Looking over the [spec](https://www.rfc-editor.org/rfc/rfc7636#section-4.1) once more, I found this

> It is RECOMMENDED that the output of a suitable random number generator be used to create a 32-octet sequence.  The octet sequence is then base64url-encoded to produce a 43-octet URL safe string to use as the code verifier.

Adopting this approach would remove a lot of custom code on our end. I'll try it in a dedicated PR stacked on this one.

_(RTFM Gio...)_

WordPressAuthenticator/GoogleSignIn/String+Random.swift
```diff
@@ -0,0 +1,22 @@
...
+
+        var randomBytes = [UInt8](repeating: 0, count: length)
+        // Use `SecRandomCopyBytes` to generate cryptographically secure random bytes to use as
+        // offsets to create the random string.
+        let status = SecRandomCopyBytes(kSecRandomDefault, length, &randomBytes)
+
+        // FIXME: Handle errors by returning nil
+        guard status == errSecSuccess else {
+            fatalError()
+        }
```
mokagio: The `SecRandomCopyBytes` function can fail for a variety of reasons, including:

<details>
<summary>See some way it can fail.</summary>
Insufficient entropy: The function may fail if there is not enough entropy available to generate the requested number of random bytes. This can happen if the system's random number generator has not collected enough entropy from hardware events or other sources. In this case, the function may return a errSecInsufficientEntropy error code.

Invalid input parameters: The function may fail if the input parameters are invalid or out of range. For example, if the output buffer pointer is null, the function will return a errSecParam error code.

Security framework errors: The function may fail due to errors within the Security framework itself. For example, if the Security framework is not initialized properly or encounters an internal error, the function may return a errSecUnimplemented error code.

System errors: The function may fail due to system-level errors, such as memory allocation failures or I/O errors. In this case, the function may return a errSecIO error code.

To handle possible errors, you should always check the return value of the `SecRandomCopyBytes` function and handle any errors appropriately. If the function returns an error code, you can use the `SecCopyErrorMessageString` function to get a human-readable error message that can help you diagnose the problem.

</details>

(_via ChatGPT_).

I haven't handled returning `nil` from here yet because of the downstream usages.

But, I think we can handle this without propagating the error outside the library by falling back to a less secure but runtime-guaranteed random value in case of failure, which I think it's reasonable to assume would be incredibly rare IRL.

WordPressAuthenticator/GoogleSignIn/OAuthRequestBody+GoogleSignIn.swift
mokagio: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/745

WordPressAuthenticator/GoogleSignIn/String+Random.swift
jkmassel: Yeah, I'm not sure I've _ever_ seen `SecRandomCopyBytes` fail in real life, so if it did I'd take that as a sign that something else on the system is catastrophically wrong – my suggestion there would be to `throw` all the way back up to the UI and tell the user "we couldn't log you in, please try again" or whatever

WordPressAuthenticator/GoogleSignIn/String+Random.swift
```diff
@@ -0,0 +1,39 @@
...
+    /// - Complexity: O(n) where n is the given `length`.
+    static func randomString(using characters: Set<Character>, withLength length: Int) -> String {
+        let allowedCharactersCount = characters.count
+        let startIndex = characters.startIndex
+
+        return (0..<length).reduce("") { accumulator, _ in
+            let randomOffset = Int.random(in: 0..<allowedCharactersCount)
+            let randomCharacter = characters[characters.index(startIndex, offsetBy: randomOffset)]
+            return accumulator + String(randomCharacter)
+        }
```
crazytonyli: Not sure if I missed anything, but can the implementation be simplified as below?
```swift
guard !characters.isEmpty else { return "" }

return (0..<length).reduce(into: "") { str, _ in
    str.append(characters.randomElement()!)
}
```

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -25,21 +25,92 @@ struct ProofKeyForCodeExchange {
...
+
+extension ProofKeyForCodeExchange {
+
+    struct CodeVerifier: Equatable {
+
+        let rawValue: String
+
+        static let allowedCharacters = Character.urlSafeCharacters
+        static let minimumLength = 43
+        static let maximumLength = 128
```
crazytonyli: These two constants could be combined into one: `static let acceptableLength = 43...128`?

WordPressAuthenticator/GoogleSignIn/String+Random.swift
```diff
@@ -0,0 +1,39 @@
+extension String {
+
+    /// Returns a cryptographically secure string generated with characters from the given `Set<Character>` and with length
+    /// `length`.
+    ///
+    /// - Complexity: O(n) where n is the given `length`.
+    static func secureRandomString(using characters: Set<Character>, withLength length: Int) -> String? {
```
crazytonyli: I did a quick search and this function appears not being used anywhere, do we need it?

crazytonyli (commented):
Sorry I missed this PR. I only have a few minor comments.

WordPressAuthenticator/GoogleSignIn/String+Random.swift
mokagio: Fair enough. We'll need to do _some_ error handling at the UI level anyway, so is not like adding a specific error for this one comes at a high cost.

mokagio: See 60181fa0

WordPressAuthenticator/GoogleSignIn/String+Random.swift
mokagio: That is correct.

I removed it in d83f7462 after your feedback. [_Kill your darlings_](https://www.goodreads.com/quotes/371112-kill-your-darlings-kill-your-darlings-even-when-it-breaks)

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: Good point. Addressed in 93f584d6

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -25,21 +25,92 @@ struct ProofKeyForCodeExchange {
...
+        }
+
+        let data = Data(bytes)
+
+        // Base64url-encoding a 32-octect sequence should always result in a 43-length string,
+        // string, but let's cap it just in case.
+        //
+        // Also notice that by base64url-encoding, we ensure the characters are in the allowed
+        // set.
+        return String(data.base64URLEncodedString().prefix(43))
```
crazytonyli: If this function is moved to `makeRandomCodeVerifier`, we can avoid duplicating the magic number 43 and use `acceptableLenght.lowerBound` instead.

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
```diff
@@ -25,21 +25,92 @@ struct ProofKeyForCodeExchange {
...
+            //
+            // We don't need the ASCII conversion, because we build `CodeVerifier` from URL safe
+            // characters.
+            let rawData = Data(codeVerifier.rawValue.utf8)
+            let hashedData: Data = rawData.sha256Hashed()
+            return hashedData.base64URLEncodedString()
         case .plain:
-            return codeVerifier
+            return codeVerifier.rawValue
+        }
```
crazytonyli: Currently the `CodeVerifier` type doesn't too much, I guess it's like a glorified string (with some additional checks). However, I think we can make use of it by moving this code challenge calculation (BTW, there is a typo in the variable name: C"h"allenge) into the CodeVerifier, for example `CodeVerifier.codeChallenge(using method: Method)`. Then, the hashing and base 64 encoding can be encapsulated inside the `CodeVerifier` type, along with some specification around the code verifier. What do you think?

crazytonyli (changes requested):

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: Great idea!

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: I thought about using the lower bound, but I though it would be confusing to use in relation to the "32-octed sequence base64url-encoded = 43 characters" assertion.

But I guess one could updated the comment a bit to explain that.

Will take a shot.

mokagio: 330bed67

WordPressAuthenticator/OAuth/ProofKeyForCodeExchange.swift
mokagio: 02cfa0a2

crazytonyli (approved):
Looks good to me!

Issue #741, opened 2023-02-13
Google SignIn without SDK – What's left to do


mokagio:
- [x] Merge the `OAuth` folder into `GoogleSignIn` – _Do this at a point when there are no other files likely to be added, to avoid conflicts_ – See #750
- [x] [Update `OAuthRequestBody` to init with a `CodeVerifier` parameter](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/740#discussion_r1110440264) - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/745
- [x] Finish wiring the pieces together. See https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/743
- [x]  Experiment to find the minimal valid `scope` to use. – https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/756
- [x] Address the feedback in #714 and write UI tests – _Blocked because sign in now requires MFA_
  - [ ] ~~Add UI tests for the various other user scenarios—this is trickier because I don't yet know how to generate dummy Google account users with or without WordPress.com connection to test the flows~~
- [x] [Validate the JWT used to create the ID token](https://developers.google.com/identity/openid-connect/openid-connect#validatinganidtoken) – Not needed. See internal ref paaHJt-4H3-p2
- [x] Ship behind remote feature flan in Jetpack/WordPress, see https://github.com/wordpress-mobile/WordPress-iOS/pull/20128
  - [x] Fix new accounts not having email in `WPAccount` object - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/759
  - [x] Add tracks logs – N.A. Logs are already available

mokagio:
All done now. The client apps are already pointing to the 7.0.0 beta version.

- https://github.com/wordpress-mobile/WordPress-iOS/pull/21344
- https://github.com/woocommerce/woocommerce-ios/pull/10483

PR #743, opened 2023-02-17
SDK-less Google SignIn – Part 5 – Connect the pieces

mokagio:
There's still work to do (#741) but, at last, here's the PR that puts all the pieces together.

You can test the behavior by running the demo app and going through the various login options.

Screenshot 2023-02-17 at 4 48 18 pm

---


Demo/AuthenticatorDemo/AppDelegate.swift
```diff
@@ -1,4 +1,6 @@
 import UIKit
+import WordPressKit
+import WordPressShared
```
mokagio: This can probably be removed.

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticationDelegate.swift
```diff
@@ -1,4 +1,6 @@
+import WebKit
 import WordPressAuthenticator
+import WordPressKit
```
mokagio, line 3: ~~Not sure why we need this. Maybe it's a leftover from early iterations.~~

We need it to use `WordPressComRestApi` and friends below.

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
```diff
@@ -76,5 +78,24 @@ extension ViewController {
                 navTitleTextColor: .white
             )
         )
+
+        WordPressAuthenticator.shared.delegate = self
+    }
+
+    // TODO: Need to handle new user flow
```
mokagio: I need to double check this... The WordPress backend should be handling this for us already. Not sure whether it makes sense to handle creating an account in the context of the demo app.

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
```diff
@@ -0,0 +1,97 @@
+import AuthenticationServices
+
+public class NewGoogleAuthenticator: NSObject {
```
mokagio, line 3: I created a new object to keep things tidy and made it `public` to use it in the demo app.

However, this code could, and probably should, be folded into `GoogleAuthenticator`.

I like the idea of exposing the building blocks for authentication in the library, so that consumers can decide whether to use the UI the library offers, or roll their own. But one could argue that's a bigger design change in the library and it ought to be considered and implemented outside the scope of the SDK-less Google SignIn work.

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
```diff
@@ -0,0 +1,97 @@
...
+            )
+
+            session.presentationContextProvider = contextProvider
+            // FIXME: should this be configurable and/or defaulting to what?
+            session.prefersEphemeralWebBrowserSession = false
+
+            // FIXME: Need this to avoid TSAN error in demo app. Can we move the call elsewhere?
+            DispatchQueue.main.async {
+                session.start()
+            }
```
mokagio: Keen on ideas on this. But it could also be that it's okay to start the session on the main thread from here.

WordPressAuthenticator/Signin/LoginEmailViewController.swift
```diff
@@ -632,3 +633,9 @@ extension LoginEmailViewController: GoogleAuthenticatorLoginDelegate {
     }
 
 }
+
+extension LoginEmailViewController: ASWebAuthenticationPresentationContextProviding {
+    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        return view.window!
+    }
+}
```
mokagio: You'll see a number of these. They became necessary when I update `showFrom(viewController:` to required `viewController` to be `UIViewController & ASWebAuthenticationPresentationContextProviding`. 

It surprised me that so many `UIViewController` subclasses in were passed as parameters to that call and required the protocol conformance but haven't researched it further.

mokagio:
@jkmassel @crazytonyli this is still a draft because it's based on a not yet merged branch, but I'd love your feedback anyway.

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
jkmassel: FWIW the demo app shouldn't have the capability to create a new user anyway – that's restricted based on the OAuth token granted to the WP/JP apps. 

If this flow allows it then sure, I guess, but I'd say for the purposes of the demo app it's ok to say "you need to have an account" instead of implementing the signup flow? 

I know this project also contains signup code, and I think for that case we could modify the back-end a little to make a public set of OAuth credentials and special-case the server to respond with something like "if this were a real app, this would've worked, but it's not so here's a real-looking response you can use to test the flow but please don't try to do anything useful with it"

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
```diff
@@ -76,5 +78,24 @@ extension ViewController {
...
+    func newGoogleSignInFlow() {
+        Task.init {
```
jkmassel: I might just be spoiled by only working on SPM stuff, but could we just do:
```suggestion
        Task {
```

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
```diff
@@ -76,5 +78,24 @@ extension ViewController {
...
+            do {
+                let token = try await self.googleAuthenticator.getOAuthToken()
+
+                presentAlert(
+                    title: "🎉",
+                    message: "Successfully authenticated with Google.\n\nEmail in received token: \(token.email)",
+                    onDismiss: {}
+                )
+            } catch {
+                presentAlert(title: "❌", message: error.localizedDescription, onDismiss: {})
```
jkmassel: Does `presentAlert` give us any guarantees that it runs on the main thread?

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
jkmassel: > But one could argue that's a bigger design change in the library and it ought to be considered and implemented outside the scope of the SDK-less Google SignIn work.

I don't mind if this is a special case tbh – there are very very few consumers of this library, so I think to say "third party auth providers are different" is reasonable – especially since SIWA is like one button and a callback 😂

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
```diff
@@ -0,0 +1,97 @@
...
+                url: url,
+                callbackURLScheme: scheme,
+                completionHandler: { result in
+                    continuation.resume(with: result)
+                }
+            )
+
+            session.presentationContextProvider = contextProvider
+            // FIXME: should this be configurable and/or defaulting to what?
+            session.prefersEphemeralWebBrowserSession = false
```
jkmassel, line 82: I think for now we can leave it as-is – it wouldn't be too hard to make it configurable, but as a user I can't see a reason I'd want to have to re-login to Google when my Safari is already signed in – that's kind of annoying, especially if I have to 2FA again

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
jkmassel: I think the only way to get around this would be to write a callback-based version of this method that uses an assertion to ensure it's running on the main thread. Then from there, wrap _that_ in `withCheckedThrowingContinuation` and mark the parent method as `@MainThread` – I'm not 100% sure that'll work because I don't know if `withCheckedThrowingContinuation` guarantees that it runs its code on the parent Task context though.

Alternatively, you could look at a delegate-based approach which would let you make guarantees about which thread that call runs on, but it'd be more complex.

jkmassel (commented):
Left some thoughts – I hope they're helpful?

Demo/AuthenticatorDemo/ViewController.swift
```diff
@@ -7,24 +8,43 @@ class ViewController: UIViewController {
...
+        CellConfiguration(text: "Show Login - With Google SDK") { [weak self] in
             guard let self else { fatalError() }
+
+            self.initializeWordPressAuthenticator(withoutGoogleSDK: false)
+            WordPressAuthenticator.showLoginFromPresenter(self, animated: true)
+        },
+        CellConfiguration(text: "Show Login - Without Google SDK") { [weak self] in
+            guard let self else { fatalError() }
+
+            self.initializeWordPressAuthenticator(withoutGoogleSDK: false)
```
crazytonyli: This argument value should be `true`, right?

Demo/AuthenticatorDemo/ViewController.swift
```diff
@@ -7,24 +8,43 @@ class ViewController: UIViewController {
...
             WordPressAuthenticator.showLoginFromPresenter(self, animated: true)
+        },
+        CellConfiguration(text: " Google SignIn - Standalone, Wihout SDK") { [weak self] in
+            guard let self else { fatalError() }
+
+            self.initializeWordPressAuthenticator(withoutGoogleSDK: false)
```
crazytonyli: The argument value here should also be true?

WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
```diff
@@ -120,11 +121,36 @@ class GoogleAuthenticator: NSObject {
     ///                  The values are updated during the Google process,
     ///                  and returned to the calling view controller via delegate methods.
     ///   - authType: Indicates the type of authentication (login or signup)
-    func showFrom(viewController: UIViewController, loginFields: LoginFields, for authType: GoogleAuthType = .login) {
+    func showFrom(
+        viewController: UIViewController & ASWebAuthenticationPresentationContextProviding,
```
crazytonyli: Considering we almost always want to use the `viewController.window` in `ASWebAuthenticationPresentationContextProviding`'s implemenation, what do you think implementing the protocol inside this library rather than making it a requirement for the client app?

Demo/AuthenticatorDemo/ViewController.swift
mokagio: 🤦‍♂️

Demo/AuthenticatorDemo/ViewController.swift
mokagio: 🤦‍♂️

WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
mokagio: Could you expand on this? Are you suggesting to do something like making `GoogleAuthenticator` conform to `ASWebAuthenticationPresentationContextProviding`, store the input `viewController` somehow, then forwarding `viewController.window` in the protocol implementation?

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
mokagio: Fair enough. I might leave a note about it in the code, then.

Demo/AuthenticatorDemo/ViewController.swift
mokagio: Address in 80072fd8, while also using more straightforward "with Google SDK".

Demo/AuthenticatorDemo/ViewController.swift
mokagio: Address in 80072fd8, while also using more straightforward "with Google SDK".

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
mokagio: @jkmassel 

> FWIW the demo app shouldn't have the capability to create a new user anyway – that's restricted based on the OAuth token granted to the WP/JP apps.

With the above, do you mean that our backend should only allow creating new users with a token that has been generated using the Google apps (i.e. the client id, secret, and audience) for WordPress and Jetpack?

I hadn't considered that, but I think it makes sense. Also, I think we can safely test the new user creation in the context of Jetpack/WordPress and add a notice in the demo app that says "you should have a valid user already", as you suggested.

We can always revisit later on in case we decided to do some work around signup.

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
mokagio: For sure.

2704cd1c

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
mokagio: I couldn't find anything in the docs about whether the UIKit method runs on the main thread.

This is a custom method so we can definitely force it to run on the main thread. 

However, I'm not sure why we would go to "such effort" in the context of the demo app? What am I missing?

FWIW, TSan doesn't complain in the successful branch of the `do catch`.

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
mokagio: Done via 85a3c271

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
mokagio: I just tried this without the `DispatchQueue.main.async` [in WordPress](https://github.com/wordpress-mobile/WordPress-iOS/pull/20128) and got the same kind of error I got with the demo app

```
=================================================================
Main Thread Checker: UI API called on a background thread: -[UIViewController view]
PID: 25962, TID: 52550464, Thread name: (none), Queue name: com.apple.root.user-initiated-qos.cooperative, QoS: 25
Backtrace:
4   WordPress                           0x0000000105261db0 $s22WordPressAuthenticator24GoogleAuthViewControllerC18presentationAnchor3forSo8UIWindowCSo26ASWebAuthenticationSessionC_tF + 132
5   WordPress                           0x0000000105261f24 $s22WordPressAuthenticator24GoogleAuthViewControllerC18presentationAnchor3forSo8UIWindowCSo26ASWebAuthenticationSessionC_tFTo + 52
6   AuthenticationServices              0x000000011033a558 -[ASWebAuthenticationSession _startDryRun:] + 92
7   WordPress                           0x00000001052d8fc0 $s22WordPressAuthenticator09NewGoogleC0C6getURL8clientId6scheme4pkce10Foundation0G0VAA0e6ClientI0V_SSAA23ProofKeyForCodeExchangeVtYaKFyScCyAJs5Error_pGXEfU_ + 512
```

I'd say for the time being it should be safer to keep the dispatch call in here and document why we did so.

WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
crazytonyli: Yes, something like that. We can have an appropriate existing type, or create a new type, to implement the protocol.

So, an alternative option would be, creating a new type:
```
class WebAuthenticationPresentationContext: ASWebAuthenticationPresentationContextProviding {
    let window: UIWindow
}
```

And then, in the `requestAuthorization()` below, using `WebAuthenticationPresentationContext(viewController.window)` as value to the `contextProvidier` argument.

mokagio: Thanks Tony!

Addressed in d5ed7ebd

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
mokagio: Dropped a comment about this with ecd5888d

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
mokagio: I tweaked the presentation and method name in the demo app to make it clearer this code path only tests getting the token from Google.

See 485b9065

mokagio:
Thank you @jkmassel and @crazytonyli for the great feedback along the way. This should now be ready for one final look. 🙇‍♂️

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
jkmassel: > our backend should only allow creating new users with a token that has been generated using the Google apps (i.e. the client id, secret, and audience) for WordPress and Jetpack?

IINM, do the OAuth dance and get a token from Google, which we can pass to WP.com to authenticate as the given user. When we perform that request, we use the WP.com OAuth clientID/secret to authenticate as the app (but not as any user because we aren't logged in at that point). That WP.com clientID/secret is what is allowed to create new users.

Does that explain it better? 😅

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticationDelegate.swift
```diff
@@ -90,3 +88,46 @@ extension ViewController: WordPressAuthenticatorDelegate {
...
+extension ViewController {
+
+    // This is just so we can avoid nesting within a dismiss block and the weak self dance.
+    //
+    // See WordPress iOS
+    //
+    // - WordPressAuthenticationManager sync(credentials:, onCompletion:)
+    // - WordPressAuthenticationManager syncWPCom(authToken:, isJetpackLogin:, onCompletion:)
+    // - AccountService createOrUpdateAccountWithAuthToken:success:failure:
+    private func sync(credentials: AuthenticatorCredentials, onSynced: @escaping () -> Void) {
```
crazytonyli: This `onSynced` is not called anywhere.

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
```diff
@@ -0,0 +1,121 @@
...
+    ) {
+        self.clientId = clientId
+        self.scheme = scheme
+        self.audience = audience
+        self.oauthTokenGetter = oautTokenGetter
+        self.contextProvider = contextProvider
+    }
+
+    /// Get the user's OAuth token from their Google account. This token can be used to authenticate with the WordPress backend.
+    public func getOAuthToken() async throws -> IDToken {
```
crazytonyli, line 59: I'm wondering if it'd be better to move the `viewController` initializer argument to here as a function argument. It appears we don't need to hold on to the view controller instance across the whole life of the `NewGoogleAuthenticator` instance? And something like `googleAuthenticator.getOAuthToken(from: viewController)` seems like better indicating that a view controller will be used to present the authentication flow.

WordPressAuthenticator/GoogleSignIn/OAuthError.swift
```diff
@@ -5,12 +5,20 @@ enum OAuthError: LocalizedError {
 
     case failedToGenerateSecureRandomCodeVerifier(status: Int32)
 
+    // OAuth token response
+    case urlDidNotContainCodeParameter(url: URL)
+    case tokenResponseDidNotIncludeIdToken
+
     var errorDescription: String {
```
crazytonyli: We probably should make the `OAuthError` public, so that when the app captures the error thrown by the `NewGoogleAuthenticator.getOAuthToken`, they cast the error to the correct type and handle it accordingly.

WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
```diff
@@ -120,11 +120,36 @@ class GoogleAuthenticator: NSObject {
...
         self.authType = authType
-        requestAuthorization(from: viewController)
+
+        guard authConfig.googleLoginWithoutSDK else {
+            // Use method that depends on SDK
+            requestAuthorization(from: viewController)
+            return
+        }
+
+        Task {
```
crazytonyli: Similar to https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/743/files#r1110446829, it's probably safer to run this Task in the main actor (i.e. adding a `@MainActor` annotation: `Task { @MainActor in `), so that the `didSignIn` function below are guaranteed to be called from the main thread.

crazytonyli (changes requested):

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticator.swift
mokagio: Gotcha 👍

Demo/AuthenticatorDemo/ViewController+WordPressAuthenticationDelegate.swift
mokagio: Good eyes!

Addressed in d1b6c91f

Thanks 🙇‍♂️

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
mokagio: That's along what I was thinking here:

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/d1b6c91f5ac76a5456da669b7d74bc6d145c01af/WordPressAuthenticator/Unified%20Auth/GoogleAuthenticator.swift#L310-L322

If that's okay with you, I'd like to tackle it in a dedicated PR

WordPressAuthenticator/Unified Auth/GoogleAuthenticator.swift
mokagio: Good point. 

Addressed in 76fd3c4e. Thanks!

WordPressAuthenticator/GoogleSignIn/OAuthError.swift
mokagio: Good point. I addressed it in 3c49abdd.

> so that when the app captures the error thrown by the `NewGoogleAuthenticator.getOAuthToken`

Worth pointing out that, for the time being, no app will call this directly. Still, it's handy to be ready for when / if they'll want to.

crazytonyli (approved):

WordPressAuthenticator/GoogleSignIn/NewGoogleAuthenticator.swift
mokagio: Addressed in #755

PR #745, opened 2023-02-24
Use `CodeVerifier` in `OAuthTokenRequestBody` for better encapsulation

mokagio:
See discussion at https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/740#discussion_r1110440264

---

### DIFF
--- a/WordPressAuthenticator.xcodeproj/project.pbxproj
+++ b/WordPressAuthenticator.xcodeproj/project.pbxproj
@@ -16,6 +16,7 @@
 		1A21EE9822832BC300C940C6 /* WordPressComOAuthClientFacade+Swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1A21EE9722832BC200C940C6 /* WordPressComOAuthClientFacade+Swift.swift */; };
 		1A4095182271AEFC009AA86D /* WPAuthenticator-Swift.h in Headers */ = {isa = PBXBuildFile; fileRef = 1A4095152271AEFC009AA86D /* WPAuthenticator-Swift.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		3108613125AFA4830022F75E /* PasteboardTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3108613025AFA4830022F75E /* PasteboardTests.swift */; };
+		3F107B0529A87AF0009B3658 /* CodeVerifier+Fixture.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F107B0429A87AF0009B3658 /* CodeVerifier+Fixture.swift */; };
 		3F30A6B4299F10BD0004452F /* Character+URLSafe.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F30A6B3299F10BD0004452F /* Character+URLSafe.swift */; };
 		3F30A6BA299F12E30004452F /* Character+URLSafeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F30A6B9299F12E30004452F /* Character+URLSafeTests.swift */; };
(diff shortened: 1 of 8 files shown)


WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
```diff
@@ -6,16 +6,34 @@ struct OAuthTokenRequestBody: Encodable {
     let clientSecret: String
     let audience: String
     let code: String
-    let codeVerifier: String
+    let rawCodeVerifier: String
```
mokagio: I considered keeping this called `codeVerifier`, but though it be clearer to have "raw" in the name when the type is `String`.

I also considered adding a computed `codeVerifier: PKCE.CodeVerifier` property. I decided not to because the `init?(value:)` method can fail and it was either force unwrap under the assumption that the Google backend will never send back an invalid one (because it wouldn't allow us to send it an invalid one) or return an `Optional`. Moreover, there is no existing reader of that value outside of the tests... YAGNI.

WordPressAuthenticator.xcodeproj/project.pbxproj
```diff
@@ -587,6 +589,8 @@
 		3FE8072329365FC20088420C /* GoogleSignIn */ = {
 			isa = PBXGroup;
 			children = (
+				3F30A6B9299F12E30004452F /* Character+URLSafeTests.swift */,
```
mokagio, line 587: This addition is because I sorted the group alphabetically. See the matching removal below.

WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
jkmassel: If the type were instead `ProofKeyForCodeExchange.CodeVerifier`, could you keep `codeVerifier` and just access the `rawValue` exclusively in `asURLEncodedData`? 

Not a huge deal – mostly just a nit 🙂

WordPressAuthenticatorTests/GoogleSignIn/CodeVerifier+Fixture.swift
```diff
@@ -0,0 +1,12 @@
+@testable import WordPressAuthenticator
+
+extension ProofKeyForCodeExchange.CodeVerifier {
+
+    /// A code verifier for testing purposes that is guaranteed to be valid and deterministic.
+    ///
+    /// The reason we care about it being deterministic is because we don't want implicit randomness test.
+    /// The only place were we want to use random values in the `CodeVerifier` tests which explicitly check the random generation.
+    static func fixture() -> Self {
+        .init(value: (0..<minimumLength).map { _ in "a" }.joined())!
```
jkmassel: Another nit, but if we want to avoid `!` here (in case SwiftLint or something wants to yell at us for it) we could add a private initializer like `init(knownGoodValue:` that uses a `precondition` or similar to crash if this API is misused. 

Not sure if we care, but it's an option 🤷

jkmassel (approved):
Looks good, left a couple of minor (potential) improvement opportunities in case you agree they are _actually_ improvements and not just overcomplications 🙂

WordPressAuthenticator/OAuth/OAuthTokenRequestBody.swift
mokagio: Your comment made me realize I was approaching the type in the context of `Decodable` (where it's handy to keep `String` properties to leverage the compiler-generated decoding) but this type is actually used for encoding data. 😅 

See 03f377c8.

WordPressAuthenticatorTests/GoogleSignIn/CodeVerifier+Fixture.swift
mokagio: I'd like to park this suggestion for the moment. We don't have a rule to flag force unwraps in the test code ATM and there are arguments in favor of leaving this as an option, trading off the possibility of test-runtime crashes in case of developer error for more compact code.

PR #747, opened 2023-03-01
Add VS Code settings to `.gitignore` and track project settings

mokagio:
Notice how the path for SwiftLint is to the Pod-installed version, to ensure consistency.

Been working with Copilot via VS Code a lot recently. These are useful `.gitignore` and `.vscode/settings.json` configurations for this project.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. – N.A.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "swiftlint.path": "../Pods/SwiftLint/swiftlint"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,17 @@ fastlane/test_output
 !.configure-files/*.enc
 # If the user created a local secrets file, ignore it
 Demo/Secrets.swift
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "swiftlint.path": "../Pods/SwiftLint/swiftlint"
```
crazytonyli, line 2: Is this setting used by SwftLint plugin? If so, should we also check in the VS Code setting (presumably the `.vscode/extensions.json` file?) where it says the SwiftLint plugin is required for this project?

mokagio: Oh, good point! Thanks. I'm still a VS Code noob.

crazytonyli (approved):

PR #748, opened 2023-03-01
Track `signed_in` event during delegated site credential login

selanthiraiyan:
## Description

This PR starts tracking the `signed_in` tracks event when the client handles the site credential login. 

**Why?**
- When the client WooCommerce-iOS handles the site credential login, the user doesn't have to login using WPCOM credentials if the self-hosted site doesn't have Jetpack. 
- So, we must consider that the user has been logged in and track the `signed_in` event.

Internal discussion about the impact of this event on Tracks funnels - p1677669246532249/1677588582.228569-slack-C046HDZL87J

## Testing steps

Please follow the testing instructions from https://github.com/woocommerce/woocommerce-ios/pull/9014 to test this PR.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
+- Track `signed_in` event when the site credential login is handled by the client. [#748]
 
 ### Internal Changes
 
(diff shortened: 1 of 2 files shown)


itsmeichigo (commented):
👋  As commented in https://github.com/woocommerce/woocommerce-ios/pull/9014, I don't think we should track this right after site credential finishes because there are still a few checks following the sign-in.

selanthiraiyan:
Thank @itsmeichigo! Closing this PR as this change isn't needed anymore. More details - https://github.com/woocommerce/woocommerce-ios/pull/9014#issuecomment-1450511106

PR #756, opened 2023-03-15
Simplify `scope` parameter used for asking OAuth token to Google

mokagio:
See discussion at https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/735#discussion_r1094342966

To test this, I run the app and verified the authentication was still successful and that my email was sent back from Google in the token to pass to the WordPress.com backend.

---

### DIFF
--- a/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
+++ b/WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
@@ -18,11 +18,9 @@ extension URL {
             ("code_challenge_method", pkce.method.urlQueryParameterValue),
             ("redirect_uri", clientId.defaultRedirectURI),
             ("response_type", "code"),
-            // The Google Sign-In SDK use "profile email", but we only need the user's email address.
-            //
-            // See:
+            // See what the Google SDK does:
             // https://github.com/google/GoogleSignIn-iOS/blob/7.0.0/GoogleSignIn/Sources/GIDScopes.m#L58-L61
-            ("scope", "email")
+            ("scope", "profile email")
         ].map { URLQueryItem(name: $0.0, value: $0.1) }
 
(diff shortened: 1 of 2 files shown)


WordPressAuthenticator/GoogleSignIn/URL+GoogleSignIn.swift
```diff
@@ -18,9 +18,11 @@ extension URL {
             ("code_challenge_method", pkce.method.urlQueryParameterValue),
             ("redirect_uri", clientId.defaultRedirectURI),
             ("response_type", "code"),
-            // TODO: We might want to add some of these or them configurable
-            // See https://developers.google.com/identity/protocols/oauth2/scopes
-            ("scope", "openid https://www.googleapis.com/auth/userinfo.email https://www.googleapis.com/auth/userinfo.profile")
+            // The Google Sign-In SDK use "profile email", but we only need the user's email address.
```
crazytonyli: The scopes are reduced from `profile + email` (when using the GoogleSignIn SDK), to `email` (when using our own implementation). Are you sure this won't cause any issue in the backend? Like, (I'm purely making assupmtion here) maybe the backend will use Google account's username to create a WP account?

mokagio: Oh, right. Good point.

Given the shape of the API, which explicitly requires the `email` alongside the token, I'd say that's the only thing we need.

But better safe than sorry hey. I could and should confirm this with the backend and API folks but in the meantime better be backwards compatible.

Addressed in [2480904](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/756/commits/24809046f5b7f32dee76048baf7c7ffe2d56b573)

crazytonyli (approved):
:shipit:

PR #763, opened 2023-03-28
Replace `AppleUser` with `SocialService.User`

mokagio:
Instead of using a type with "Apple" in its name as the associated type for `SocialService` `case .apple`, we can use a general purpose type, `SocialService.User`. After all, there is nothing Apple-specific in the email, full name pair and the information that the service used was Google or Apple is already encoded in the `case` name.

The next step for this will be to make `case .google` use `User`, too, as described in #759. But I need #761 first.

The bulk of this PR is actually code to enable testing the change.

@ScoutHarris @jaclync, I asked for your review as suggested by GitHub. I assume that's because you worked on Sing In with Apple? If so, your input would be much appreciated 🙇‍♂️ 

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #764, opened 2023-03-28
Replace `GIDGoogleUser` with `SocialService.User` in case .google

### LINKED ISSUE #759: Google sign in without SDK – New accounts lack email in `WPAccount`
I tested creating a new WordPress.com account by signing in without the Google SDK and WordPress iOS crashed. [This](https://github.com/wordpress-mobile/WordPress-iOS/pull/20128) is the branch I was working on.

 Screenshot 2023-03-24 at 3 52 00 pm

The issue, as far as I could trace it, was that when the app created a `LoginEpilogueUserInfo` with the `WPAccount` instance it just loaded from the DB (?!) via `WPAccount.lookupDefaultWordPressComAccount(_:)`, the instance had no `email`.

mokagio:
Closes #759.

As described in #759, the idea with this PR is to remove the Google SDK-dependent type from the `SocialService` `google` `case` associated value and use the pure Swift `SocialService.User` instead. The new type was introduced in #763.

## To test

The code changes here are consumed via unit tests alone. The demo app doesn't read the `loginFields` metadata where `SocialService` lives.

The "real" test for this new code is in the Jetpack/WordPress iOS that uses it https://github.com/wordpress-mobile/WordPress-iOS/pull/20128

---

### DIFF
--- a/WordPressAuthenticator/Model/LoginFields.swift
+++ b/WordPressAuthenticator/Model/LoginFields.swift
@@ -152,7 +152,7 @@ public class LoginFieldsMeta: NSObject {
 
     @objc public var socialServiceIDToken: String?
 
-    var googleUser: GIDGoogleUser?
+    var googleUser: SocialService.User?
 
     var appleUser: SocialService.User?
 
@@ -165,7 +165,7 @@ public class LoginFieldsMeta: NSObject {
          requiredMultifactor: Bool = false,
          socialService: SocialServiceName? = nil,
          socialServiceIDToken: String? = nil,
-         googleUser: GIDGoogleUser? = nil,
+         googleUser: SocialService.User? = nil,
          appleUser: SocialService.User? = nil) {
         self.emailMagicLinkSource = emailMagicLinkSource
         self.jetpackLogin = jetpackLogin
--- a/WordPressAuthenticator/Services/SocialService.swift
+++ b/WordPressAuthenticator/Services/SocialService.swift
@@ -9,6 +9,13 @@ public enum SocialService {
         public let fullName: String
     }
 
+    public var user: User {
+        switch self {
+        case .google(let user): return user
+        case .apple(let user): return user
+        }
+    }
+
     /// Google's Signup Linked Account
     ///
     case google(user: User)
--- a/WordPressAuthenticator/Services/SocialService.swift
+++ b/WordPressAuthenticator/Services/SocialService.swift
@@ -11,7 +11,7 @@ public enum SocialService {
 
     /// Google's Signup Linked Account
     ///
-    case google(user: GIDGoogleUser)
+    case google(user: User)
 
     /// Apple's Signup Linked Account
     ///
--- a/WordPressAuthenticator/Signin/LoginViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginViewController.swift
@@ -132,8 +132,8 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
             fatalError()
         }
 
-        let service = loginFields.meta.googleUser.flatMap {
-            return SocialService.google(user: $0)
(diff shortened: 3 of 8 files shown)


WordPressAuthenticator/Services/SocialService.swift
```diff
@@ -9,9 +9,16 @@ public enum SocialService {
...
+        switch self {
+        case .google(let user): return user
+        case .apple(let user): return user
+        }
+    }
+
     /// Google's Signup Linked Account
     ///
-    case google(user: GIDGoogleUser)
+    case google(user: User)
```
mokagio, line 21: `apple` and `google` both have a `User` associated type now.

This kind of makes it backward to have an enum with associated type, we should have a type with an enum property

```swift
struct SocialUser {

  let fullName: String
  let email: String
  let source: SocialUser.Source // .apple, .google, ...
}
```

I've come across other bits of information in some places that could go into this type, such as the token.

I'm skeptical of working on "such a big" rewrite, though. We should be making changes to the auth layer here carefully. Also, I'd like to first remove all the Google-SDK cruft first. The downside is that we'll have to go through at least two major versions. I think it's worth it, though.

WordPressAuthenticator/Model/LoginFields.swift
```diff
@@ -165,7 +165,7 @@ public class LoginFieldsMeta: NSObject {
          requiredMultifactor: Bool = false,
          socialService: SocialServiceName? = nil,
          socialServiceIDToken: String? = nil,
-         googleUser: GIDGoogleUser? = nil,
+         googleUser: SocialService.User? = nil,
          appleUser: SocialService.User? = nil) {
```
mokagio, line 169: I wrote it elsewhere already, but it seems to me like these parameters are ripe for consolidation into a single type (`SocialUser`?).

I'm not keen on introducing such a change right now because I'd like to stay focused on removing the Google SDK only. The tradeoff is that we'll have to cycle through two major version changes, but that seems acceptable. It's just a number after all.

mokagio: > I wrote it elsewhere

Elsewhere = [the file after this](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/764/files#r1152489604), 14 hours ago 🤦‍♂️ 🙃 

But hey, at least I'm consistent in the naming suggestion I'm leaving...

WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -425,10 +425,19 @@ extension LoginViewController {
...
-        loginFields.username = googleEmail
+        updateLoginFields(
+            email: googleEmail,
+            username: googleEmail,
+            fullName: googleUser.profile?.name ?? "",
+            googleToken: googleToken
+        )
+    }
+
+    func updateLoginFields(email: String, username: String, fullName: String, googleToken: String) {
```
crazytonyli: This `username` argument is not used, should we get rid of it? Also, any reason adding another `updateLoginFields` function here, rather than do the assignments in the one above?

crazytonyli (changes requested):

WordPressAuthenticator/Signin/LoginViewController.swift
mokagio: What I should have done with that `username` parameter was using it in the method and then passing `.email` at callsite. I guess I was trying to cover my basis.

But your observation made me realize the version with the `GIDGoogleUser` input is unused, and this new one is only used by it. So, I removed both of them.

crazytonyli (approved):

WordPressAuthenticator/Signin/LoginViewController.swift
crazytonyli: Yay, more deletions!

PR #768, opened 2023-04-12
Site credentials: Always trigger `completionHandler` when login finishes if possible

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/9427

**Description**
This PR fixes the issue for custom site credential login by always triggering `completionHandler` if it's available, regardless of the configurations.

**Testing**
Please follow https://github.com/woocommerce/woocommerce-ios/pull/9440 for the testing steps.

---

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
@@ -479,6 +479,38 @@ private extension SiteCredentialsViewController {
             displayError(error as NSError, sourceTag: sourceTag)
         }
     }
+
+    func syncDataOrPresentWPComLogin(with wporgCredentials: WordPressOrgCredentials) {
+        if configuration.isWPComLoginRequiredForSiteCredentialsLogin {
+            presentWPComLogin(wporgCredentials: wporgCredentials)
+            return
+        }
+        // Client didn't explicitly ask for WPCOM credentials. (`isWPComLoginRequiredForSiteCredentialsLogin` is false)
+        // So, sync the available credentials and finish sign in.
+        //
+        let credentials = AuthenticatorCredentials(wporg: wporgCredentials)
+        WordPressAuthenticator.shared.delegate?.sync(credentials: credentials) { [weak self] in
+            NotificationCenter.default.post(name: Foundation.Notification.Name(rawValue: WordPressAuthenticator.WPSigninDidFinishNotification), object: nil)
+            self?.showLoginEpilogue(for: credentials)
+        }
+    }
+
+    func presentWPComLogin(wporgCredentials: WordPressOrgCredentials) {
+        // Try to get the jetpack email from XML-RPC response dictionary.
+        //
+        guard let loginFields = makeLoginFieldsUsing(xmlrpc: wporgCredentials.xmlrpc,
+                                                     options: wporgCredentials.options) else {
+            WPAuthenticatorLogError("Unexpected response from .org site credentials sign in using XMLRPC.")
+            let credentials = AuthenticatorCredentials(wporg: wporgCredentials)
+            showLoginEpilogue(for: credentials)
+            return
+        }
+
+        // Present verify email instructions screen. Passing loginFields will prefill the jetpack email in `VerifyEmailViewController`
+        //
+        presentVerifyEmail(loginFields: loginFields)
+    }
+
     // MARK: - Private Constants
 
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
```diff
@@ -541,6 +541,11 @@ extension SiteCredentialsViewController {
         let wporg = WordPressOrgCredentials(username: username, password: password, xmlrpc: xmlrpc, options: options)
         let credentials = AuthenticatorCredentials(wporg: wporg)
 
+        /// If `completionHandler` is available, return early with the credentials.
+        if let completionHandler = completionHandler {
+            return completionHandler(wporg)
```
jaclync: ❓ since it early returns, do we not need any of the logic below like `delegate.sync(credentials: credentials)` and its completion block since `configuration.isWPComLoginRequiredForSiteCredentialsLogin` is `false` for WCiOS?

When I looked into the usage of `SiteCredentialsViewController` (not a public class), it always has a non-nil completion block from `WordPressAuthenticator.showSiteCredentialLogin` at the moment. The completion block type is nullable now only for the initial `init?(coder: NSCoder)` if I understand correctly. This means that none of the code below this would be needed 😅 If this is the case, what do you think about deleting the code below so that it's clear for future developers that this is the end of the function?

jaclync (commented):
I had a question, as I'm quite new to this part of the code 😅 What's the issue in https://github.com/woocommerce/woocommerce-ios/issues/9427 that this change aims to fix?

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
itsmeichigo: @jaclync  As you can see, the initializer with the non-optional `completionHandler` is only used in `WordPressAuthenticator.showSiteCredentialLogin` - a method to present site credential login directly and triggers the completion block immediately after login completes. The goal is to send the credentials back and not to authenticate. In Woo, this flow is only used during the Jetpack connection flow for users authenticated with WPCom.

`SiteCredentialsViewController` has another initializer  `init?(coder: NSCoder)` without the completion handler, which is used in all the other flows through `SiteCredentialsViewController.instantiate(from: .siteAddress)`. In these flows, we trigger the delegate when the login completes.

itsmeichigo:
> What's the issue in https://github.com/woocommerce/woocommerce-ios/issues/9427 that this change aims to fix?

Sorry for not clarifying this. This PR tries to fix the following issue in https://github.com/woocommerce/woocommerce-ios/issues/9427:

> The site credential login step leads to application password login instead of Jetpack connection.

Since we were triggering the delegate instead of the `completionHandler` after site credential completes, Woo was handling application password login instead of returning the site credentials for fetching Jetpack connection URL. By prioritizing the `completionHandler` when it's available, we ensure that the delegate method is not triggered when it's not needed.

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteCredentialsViewController.swift
jaclync: What do you think if this `finishedLogin` calls two separate functions, one with a completion block and the other without, with `wporg` and `credentials` passed to them from the beginning of the function?

For developers new to this code, I think it can be quite confusing reading about or not knowing these two code paths.

itsmeichigo: Good point! Separate this method into 3 parts in 075d308a039525f119a717641b03d2f74a8815af - lmk what you think!

jaclync (approved):
Thanks for the changes, this looks more clear to me 👍 

Also wanted to ask, might this change affect WPiOS/JPiOS if they call the site credentials with a completion block? Since this triggers the completion block and skips the delegate etc., it can have some impact on certain usage.

I encountered some issues in the testing of the main PR, but approving this first since the changes seem fine.

itsmeichigo:
Thanks for the review @jaclync! I did a quick check and found that WPiOS/JPiOS [doesn't use](https://github.com/wordpress-mobile/WordPress-iOS/search?q=showSiteCredentialLogin) the method `showSiteCredentialLogin`. This method was introduced by me for Woo so I hope it's safe to merge this change. By the way, this is the intended behavior so it feels right to me to fix it (hence the versioning for bug-fix instead of new feature).

PR #775, opened 2023-07-12
Break retain cycles between button onTap action and view controllers

staskus:
**Part of:** https://github.com/wordpress-mobile/WordPress-iOS/issues/21075

Assigning a function directly to a closure creates a retain cycle between buttonViewController and viewController. 

Mmory Leak - WordPressAuthenticator Pod

**Solution**

Pass a closure and use a `weak self` instead. Explaining each fix in the commits.

---

- [x] Bump version.

### DIFF
--- a/WordPressAuthenticator/NUX/NUXViewControllerBase.swift
+++ b/WordPressAuthenticator/NUX/NUXViewControllerBase.swift
@@ -286,12 +286,44 @@ extension NUXViewControllerBase where Self: UIViewController, Self: UIViewContro
     private func setupNotificationsIndicator() {
         helpNotificationIndicator.isHidden = true
 
-        NotificationCenter.default.addObserver(forName: .wordpressSupportNotificationReceived, object: nil, queue: nil) { [weak self] _ in
+        wordpressSupportNotificationReceivedObserver = NotificationCenter.default.addObserver(forName: .wordpressSupportNotificationReceived, object: nil, queue: nil) { [weak self] _ in
             self?.refreshSupportNotificationIndicator()
         }
-        NotificationCenter.default.addObserver(forName: .wordpressSupportNotificationCleared, object: nil, queue: nil) { [weak self] _ in
+        wordpressSupportNotificationClearedObserver = NotificationCenter.default.addObserver(forName: .wordpressSupportNotificationCleared, object: nil, queue: nil) { [weak self] _ in
             self?.refreshSupportNotificationIndicator()
         }
+    }
+
+    private var wordpressSupportNotificationReceivedObserver: NSObjectProtocol? {
+        get {
+            
+            objc_getAssociatedObject(self, &wordpressSupportNotificationReceivedKey) as? NSObjectProtocol
+        }
+        set {
+            objc_setAssociatedObject(self, &wordpressSupportNotificationReceivedKey, newValue, .OBJC_ASSOCIATION_RETAIN)
+        }
+    }
+
+    private var wordpressSupportNotificationClearedObserver: NSObjectProtocol? {
+        get {
+            objc_getAssociatedObject(self, &wordpressSupportNotificationClearedKey) as? NSObjectProtocol
+        }
+        set {
+            objc_setAssociatedObject(self, &wordpressSupportNotificationClearedKey, newValue, .OBJC_ASSOCIATION_RETAIN)
+        }
+    }
+
+    func removeNotificationObservers() {
+        if let wordpressSupportNotificationReceivedObserver {
+            NotificationCenter.default.removeObserver(wordpressSupportNotificationReceivedObserver)
+        }
+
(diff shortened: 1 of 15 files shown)


staskus:
@guarani 

I'm thinking about how to better test it for the reviewer.

Do you think it's better to expand the testing suite here (add unit tests, add more example UI) or is it enough to build and test https://github.com/wordpress-mobile/WordPress-iOS/pull/21092 locally together with these changes?

Most of the changes are providing `[weak self]` inside closures, not sure if anything could break in this instance.

staskus:
@guarani 
I added unit tests confirming that most of the VCs in the suite don't have to retain cycles, at least the most obvious ones. I found such a test both beneficial and easy to write, so I think it's a win win. I will guard against the regressions.

To save the time, here's the test in action showing it failing when reverting some of the changes I made:

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/4062343/c4d18742-45a0-452c-a8a8-1d35b9eb2fa0

guarani:
👋 Hey @staskus, these tests for memory leaks look promising. 
I'll review this today (I couldn't get to it yesterday).

WordPressAuthenticator/NUX/NUXViewControllerBase.swift
```diff
@@ -286,14 +286,46 @@ extension NUXViewControllerBase where Self: UIViewController, Self: UIViewContro
...
+    }
+
+    private var wordpressSupportNotificationClearedObserver: NSObjectProtocol? {
+        get {
+            objc_getAssociatedObject(self, &wordpressSupportNotificationClearedKey) as? NSObjectProtocol
+        }
+        set {
+            objc_setAssociatedObject(self, &wordpressSupportNotificationClearedKey, newValue, .OBJC_ASSOCIATION_RETAIN)
+        }
+    }
```
guarani: nitpick: Another option that came to mind is letting the classes that implement this protocol store the observers. These variables would move to the protocol:
```
    var wordpressSupportNotificationReceivedObserver: NSObjectProtocol? { get set }
    var wordpressSupportNotificationClearedObserver: NSObjectProtocol? { get set }
```
then the protocol could be class-bound:
```
public protocol NUXViewControllerBase: AnyObject {
...
```
Then each implementing class would need to implement these variables:
```
class NUXTableViewController {
    public var wordpressSupportNotificationReceivedObserver: NSObjectProtocol?
    public var wordpressSupportNotificationClearedObserver: NSObjectProtocol?
...
```

However, neither option seems clearly better than the other, so happy to leave it as-is.

guarani (approved):
Changes look good! I first checked that memory usage was going up when navigating the demo app on `trunk` and then verified that it stays the same on this branch. I also checked the Memory Graph and saw that none of the view controllers beyond the demo's home screen stay in memory.

WordPressAuthenticator/NUX/NUXViewControllerBase.swift
staskus: Thanks! Yes, I don't like adding this Obj-C magic to the code, I didn't want to refactor the structure of the code too much though. I will check your approach, maybe it will be a sweet spot between avoiding the Obj-C runtime and having to refactor too much in the original code.

guarani: I looked at the change and it seems like a good approach (not one I would have thought of) 👍 
`addNotificationObserver` caught my eye because it's storing the observer (after its added), despite the name being being **add**. Definitely not an issue, just an observation and no need (from my side) to change it.

staskus: I forgot to mention, the approach you suggested wasn't acceptable to the compiler. If we wanted to set observer value, we had to make all the methods inside the extension mutable. Also, it prevented from calling those methods in`deinit`.  

> addNotificationObserver caught my eye because it's storing the observer (after its added), despite the name being being add. Definitely not an issue, just an observation and no need (from my side) to change it.

Yes, probably not the best naming. Due to the `mutable` issue I described above, I need to make the value-changing operations from inside the viewControllers. I thought using array instead of separate observer vars will be a cleaner approach.

guarani: Do you recall if you made the protocol class-bound? I also ran into this when I was trying out the above last week, and then realized that the error only happens because the compiler thinks the protocol could be used on structs as well. Adding `: AnyObject` to the protocol fixed it for me.

staskus: > Do you recall if you made the protocol class-bound? 

No, I haven't, that's a good point, I missed that from the mutable hint of the compiler 🥲 Yes, then that would've worked.

guarani: Ok 👍

PR #777, opened 2023-07-25
Remove GoogleSignIn dependency

mokagio:
Notice this points to release/7.0.0 instead of `trunk`. I have at least a followup PR I want to open on top of this, and I want to merge a single breaking change into `trunk`.

---

### DIFF
--- a/WordPressAuthenticator/Services/SocialService.swift
+++ b/WordPressAuthenticator/Services/SocialService.swift
@@ -1,7 +1,4 @@
-import GoogleSignIn
-
-// MARK: - Social Services Metadata
-//
+/// Social service metadata.
 public enum SocialService {
 
     public struct User {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ _None._
 
 ## 7.0.0
 
+### Breaking Changes
+
+- Removed dependency GoogleSignIn SDK and flags to configure it [#777]
+
 ## 6.3.0
 
 _Note: This should have been 6.2.1 because it contained only a bug fix. Unfortunately we currently don't have automation in place to enfore SemVer. Given a beta had already been released, we went with 6.3.0 stable._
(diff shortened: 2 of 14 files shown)


WordPressAuthenticator/Services/SocialService.swift
```diff
@@ -1,7 +1,4 @@
-import GoogleSignIn
-
-// MARK: - Social Services Metadata
-//
+/// Social service metadata.
 public enum SocialService {
```
mokagio, line 2: Given removing the library will be a breaking change, we might as well update this type to be 

1. Surface the metadata and `User` information
2. Have an `enum` with just the "type", Google vs Apple

mokagio: See #778 – currently in progress

crazytonyli: My understanding is removing a library dependency doesn't count as a breaking change. If the app needs the removed library, it makes sense to declare the dependency on the app side. That means the app can simply add a `pod 'GoogleSingleIn'` to their Podfile, which should be there in the first place.

crazytonyli (approved):
Love to see some deleted code and libraries 👍

WordPressAuthenticator/Services/SocialService.swift
mokagio: Correct.

What I think is the breaking change here is the removal of the `googleLoginWithoutSDK` parameter from the `public` initializer:

Screenshot 2023-07-28 at 10 39 36 am

If there was code that called that `init` passing an explicit value, e.g. [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/2bed9cc895efacf30eb4b1cf8046738ad9537a31/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift#L93), it would break.

crazytonyli: Ah of course. That makes sense!

PR #778, opened 2023-07-26
Streamline metadata after GoogleSignIn removal

mokagio:
This PR does some tidy up:

- Make `LoginFieldsMeta` `internal` so that we can more easily modify it in the future
- Restructures `SocialService` into `SocialUser`, removing the `SocialServiceName` `SocialService` `enum` cases duplicity
- Uses `SocialUser` instead of `SocialService` in the `presentSignupEpilogue` `WordPressAuthenticatorDelegateProtocol` method

The rationale for those changes was to improve encapsulation and remove duplication. I hope it'll help us further simplify the library in the future.

You can see the code in action in:

- https://github.com/woocommerce/woocommerce-ios/pull/10483
- https://github.com/wordpress-mobile/WordPress-iOS/pull/21344

---


WordPressAuthenticator/Model/LoginFieldsMeta.swift
```diff
@@ -0,0 +1,83 @@
...
+    ///
+    var requiredMultifactor: Bool
+
+    /// Identifies a social login and the service used.
+    ///
+    var socialService: SocialServiceName?
+
+    var socialServiceIDToken: String?
+
+    var socialUser: SocialUser?
```
mokagio, line 45: Something I'm particularly keen to get rid of is having all these `social*` properties. 

Now that the type is `internal`, we should be able to consolidate them (maybe into `SocialUser`) without new breaking changes.

WordPressAuthenticatorTests/Mocks/WordPressAuthenticatorDelegateSpy.swift
```diff
@@ -1,4 +1,5 @@
 @testable import WordPressAuthenticator
+import WordPressKit
```
mokagio, line 2: This import is no longer needed.

WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -299,11 +299,15 @@ extension LoginViewController {
             return
         }
 
+        // FIXME: Tests this
```
mokagio: Woops. I shall followup on this ASAP...

WordPressAuthenticator/Model/LoginFields.swift
```diff
@@ -36,7 +36,22 @@ public class LoginFields: NSObject {
...
+    }
+
+    @objc public var requiredMultifactor: Bool {
+        meta.requiredMultifactor
+    }
+
+    @objc public var xmlrpcURL: NSURL? {
+        get { meta.xmlrpcURL }
+        set { meta.xmlrpcURL = newValue }
+    }
```
crazytonyli, line 53: I'm guessing the reason these properties are public is they are used by the apps?

WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -299,11 +299,15 @@ extension LoginViewController {
...
         let appleConnectParameters: [String: AnyObject]? = {
-            if let appleUser = loginFields.meta.appleUser {
-                return AccountServiceRemoteREST.appleSignInParameters(email: appleUser.email, fullName: appleUser.fullName)
-            }
-            return nil
+            guard let user = loginFields.meta.socialUser else { return nil }
+            guard case .apple = user.service else { return nil }
```
crazytonyli: Nitpick: merging these two in to one guard statement?

WordPressAuthenticator/Model/LoginFields.swift
mokagio: Correct. At least, I think that's why I set them like that, but I shall double check. Thank you for asking this 🙌

WordPressAuthenticator/Signin/LoginViewController.swift
mokagio: Nice. Addressed in 7d91349a

WordPressAuthenticator/Model/LoginFields.swift
mokagio: Actually, here's why the were `public`: They are required to be so that they can be seen in `LoginFacade`:

image

WordPressAuthenticator/Signin/LoginViewController.swift
mokagio: Done in [`ccdc610` (#778)](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/778/commits/ccdc6109870dd23c8e7ba9586fdd917c61c96831)

mokagio:
@crazytonyli this is ready for another look at your earliest convenience. Thanks

WordPressAuthenticator/Model/LoginFields.swift
crazytonyli: Interesting... I didn't know that you have to mark these property as public for them to be available to `LoginFacade.m` which sits in the same module.

crazytonyli (approved):

PR #780, opened 2023-08-24
Bump deployment target to iOS 15.0

mokagio:
What it says in the title.

Both current known consumers of the library target iOS 15 and upwards.

This, unfortunately, introduced various deprecation warnings. Fixing them is out of scope at the moment, given the possibility that soon some of the affected classes will be removed as part of a UX refresh.

Screenshot 2023-08-24 at 1 24 42 pm


crazytonyli:
Oh no, those warnings [failed CI build](https://buildkite.com/automattic/wordpress-authenticator-ios/builds/839#018a2596-4589-4b83-9524-6de076d75aa0/6-1519).
>  [!] WordPressAuthenticator did not pass validation, due to 16 warnings (but you can use `--allow-warnings` to ignore them).

mokagio:
@crazytonyli thank you for checking!

😳 ouch... 🤔 I like how our pod validation doesn't allow warnings. So, I think it might be best to bench this PR. 

As per the PR description, I don't think it's worth spending time on those warnings at the moment:

> Fixing them is out of scope at the moment, given the possibility that soon some of the affected classes will be removed as part of a UX refresh.

What do you think?

mokagio:
Closing as obsolete

PR #784, opened 2023-10-03
Support WCiOS app login links with WPCOM and site credentials navigation

jaclync:
Related to https://github.com/woocommerce/woocommerce-ios/pull/10831

## Description
Introduces two navigations entry points for both the Site Credentials view and the Password view. Those will be used for the login deep link, where the Site Credentials view will be called when a non-jetpack installed store is trying to log in, and the Password view for WPCOM accounts trying to log in.

## Testing steps

Please follow the steps at https://github.com/woocommerce/woocommerce-ios/pull/10831

---

### DIFF
--- a/WordPressAuthenticator/Navigation/NavigateToEnterSiteCredentials.swift
+++ b/WordPressAuthenticator/Navigation/NavigateToEnterSiteCredentials.swift
@@ -4,25 +4,32 @@ import Foundation
 ///
 public struct NavigateToEnterSiteCredentials: NavigationCommand {
     private let loginFields: LoginFields
+    private let onDismiss: () -> Void
 
-    public init(loginFields: LoginFields) {
+    public init(loginFields: LoginFields, onDismiss: (() -> Void)?) {
         self.loginFields = loginFields
+        self.onDismiss = onDismiss ?? {}
     }
     public func execute(from: UIViewController?) {
         let navigationController = (from as? UINavigationController) ?? from?.navigationController
         presentSiteCredentialsView(navigationController: navigationController,
-                                   loginFields: loginFields)
+                                   loginFields: loginFields,
+                                   onDismiss: onDismiss)
     }
 }
 
 private extension NavigateToEnterSiteCredentials {
-    func presentSiteCredentialsView(navigationController: UINavigationController?, loginFields: LoginFields) {
+    func presentSiteCredentialsView(navigationController: UINavigationController?, loginFields: LoginFields, onDismiss: @escaping () -> Void) {
         guard let controller = SiteCredentialsViewController.instantiate(from: .siteAddress) else {
             WPAuthenticatorLogError("Failed to navigate to SiteCredentialsViewController")
             return
         }
 
         controller.loginFields = loginFields
+        controller.dismissBlock = { _ in
+            onDismiss()
+        }
+
         navigationController?.pushViewController(controller, animated: true)
     }
 }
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Navigation/NavigateToEnterSiteCredentials.swift
```diff
@@ -0,0 +1,28 @@
...
+}
+
+private extension NavigateToEnterSiteCredentials {
+    func presentSiteCredentialsView(navigationController: UINavigationController?, loginFields: LoginFields) {
+        guard let controller = SiteCredentialsViewController.instantiate(from: .siteAddress) else {
+            WPAuthenticatorLogError("Failed to navigate to SiteCredentialsViewController")
+            return
+        }
+
+        controller.loginFields = loginFields
```
ThomazFB, line 25: Jaclyn left some comments regarding the need to initialize or not both the `dismissBlock` and `errorToPresent`, I'm unsure what possible scenarios could require the `errorToPresent` value since we're jumping directly into them from the deep link data and how we could feed the error from WCiOS's AuthenticationManager. From my tests, network errors are still being handled correctly.

Also, it seems we can't allow those views to become dismissable from those `NavigateTo` events.

With that said, I just went without initializing the `errorToPresent` and `dismissBlock` value. Please let me know if you think we should do otherwise.

WordPressAuthenticator/Navigation/NavigateToEnterWPCOMPassword.swift
```diff
@@ -0,0 +1,28 @@
...
+}
+
+private extension NavigateToEnterWPCOMPassword {
+    func presentPasswordView(navigationController: UINavigationController?, loginFields: LoginFields) {
+        guard let controller = PasswordViewController.instantiate(from: .password) else {
+            WPAuthenticatorLogError("Failed to navigate to PasswordViewController from GetStartedViewController")
+            return
+        }
+
+        controller.loginFields = loginFields
```
ThomazFB, line 25: Same point as https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/784/files#r1350930454

WordPressAuthenticator/Navigation/NavigateToEnterSiteCredentials.swift
itsmeichigo: From what I understand, `errorToPresent` is internal to the authenticator library, it is used to handle login errors only not any other errors from outside.

`dismissBlock` is only called when the `SiteCredentialsViewController` is initialized with `isDismissible: true`, which will show a dismiss button instead of the back button on the top left. This is not relevant for our case as we are pushing the view into the navigation stack now.

WordPressAuthenticator/Navigation/NavigateToEnterWPCOMPassword.swift
itsmeichigo: Similar to the above reply, we don't need to update those fields in this case.

itsmeichigo (approved):
LGTM :shipit:

WordPressAuthenticator/Navigation/NavigateToEnterSiteCredentials.swift
ThomazFB: Thank you for this clarification, @itsmeichigo! 🙇  I came to a similar conclusion but got worried I was missing something I didn't understand completely. 😅

PR #792, opened 2023-10-23
Fix an issue where `guessXMLRPCURL` is called with an URL without a scheme

staskus:
Related issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/20985
Can be tested with this PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/21848

## Technical details

`guessXMLRPCURL` can be called with a URL without a scheme due to a race condition which results in an error.

When user submits self-hoste site url, `validateForm()` updates `loginFields.siteAddress` to have appropriate scheme. However, in between asynchronuous `checkSiteExistence()` and `guessXMLRPCURL()` calls `loginFields.siteAddress` can be changed by textField delegate methods back to an URL without a scheme.

### Solution

Make sure `guessXMLRPCURL` is called with an url with a scheme

---

### DIFF
--- a/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
@@ -450,7 +450,7 @@ private extension SiteAddressViewController {
                 return
             }
             // Proceeds to check for the site's WordPress
-            self.guessXMLRPCURL(for: self.loginFields.siteAddress)
+            self.guessXMLRPCURL(for: url.absoluteString)
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ _None._
 
 -->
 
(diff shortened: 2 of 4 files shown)


WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewController.swift
```diff
@@ -450,7 +450,7 @@ private extension SiteAddressViewController {
                 return
             }
             // Proceeds to check for the site's WordPress
-            self.guessXMLRPCURL(for: self.loginFields.siteAddress)
+            self.guessXMLRPCURL(for: url.absoluteString)
```
staskus, line 453: By the time `self.guessXMLRPCURL` gets called, `self.loginFields.siteAddress` can be already changed. Better to rely on an `url` variable that belongs to a local scope.

guarani (approved):
This fix makes sense and it works.

Thanks for tackling this, @staskus! 🙇

PR #788, opened 2023-10-09
Add Security Keys support

Ecarrion:
closes https://github.com/woocommerce/woocommerce-ios/issues/10859

# Depends On

-  https://github.com/wordpress-mobile/WordPressKit-iOS/pull/629

# Merged Into

- https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/785
- https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/786
- https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/787
- https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/789

# Implemented On

- https://github.com/woocommerce/woocommerce-ios/pull/10904
- https://github.com/wordpress-mobile/WordPress-iOS/pull/21739

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #793, opened 2023-10-24
Fix: Submitting wrong code for SMS 2FA can result in a vague error message

staskus:
Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/20983
Related PR with testing steps: https://github.com/wordpress-mobile/WordPress-iOS/pull/21863

When a user enters an incorrect length SMS code, `authTypeAndNonce` is constructed as empty. There's no point attempting to log in with an empty nonce since the user would receive a non-descriptive "two_step_nonce required" message.

Changes made https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/788 have slightly changed the originally reported issue. Now it incorrectly displays: "Please fill out all the fields" after entering too short an SMS code. 

Before we would call `loginWithNonce` with an empty nonce resulting in "two_step_nonce required"  error.
Now we called `validateFormAndLogin` with an empty nonce resulting in "Please fill out all the fields" error.

### Changes

Explicitly show "invalid 2fa code message" when entering an incorrect SMS code.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ _None._
 
 ### Bug Fixes
 
-- Fix an issue where `guessXMLRPCURL` was called with an URL without a scheme resulting in an error [#792]
+- Fix an issue where `guessXMLRPCURL` was called with an URL without a scheme resulting in an error. [#792]
+- Fix an issue that leads to an ambiguous error message when an incorrect SMS 2FA code is submitted. [#793]
 
 ## 7.2.0
 
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -9,7 +9,7 @@ PODS:
   - SVProgressHUD (2.2.5)
   - SwiftLint (0.49.1)
   - UIDeviceIdentifier (2.3.0)
-  - WordPressAuthenticator (7.2.1-beta.1):
+  - WordPressAuthenticator (7.2.1-beta.2):
     - Gridicons (~> 1.0)
     - "NSURL+IDN (= 0.4)"
     - SVProgressHUD (~> 2.2.5)
@@ -71,7 +71,7 @@ SPEC CHECKSUMS:
   SVProgressHUD: 1428aafac632c1f86f62aa4243ec12008d7a51d6
   SwiftLint: 32ee33ded0636d0905ef6911b2b67bbaeeedafa5
   UIDeviceIdentifier: 442b65b4ff1832d4ca9c2a157815cb29ad981b17
-  WordPressAuthenticator: cbf4306acc96a09296603f650e85600fbf74f2d4
+  WordPressAuthenticator: 1d73abee8fdd87032e987c40f140423a6aa0e577
   WordPressKit: a5432c2e3c2247c2b83b3ebf0acec75ae00782ef
   WordPressShared: 0aa459e5257a77184db87805a998f447443c9706
   WordPressUI: 1cf47a3b78154faf69caa18569ee7ece1e510fa0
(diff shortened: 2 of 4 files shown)


guarani (approved):
> When a user enters an incorrect length SMS code, `authTypeAndNonce` is constructed as empty. There's no point attempting to log in with an empty nonce since the user would receive a non-descriptive "two_step_nonce required" message.

It looks like this screen only enables the Continue button if the entered code is of length 6, 7, or 8 characters. When an SMS code is needed (7 digits) but I enter 6 digit code, that's the only scenario I'm aware of that reproduces this issue ([via this code](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/9e3af4c6cd414957bc5fbb60f308b921aa139be1/WordPressKit/SocialLogin2FANonceInfo.swift#L35-L39)). Is there a way to know what length is expected and disable the Continue button if the entered code length doesn't match the expected code length? If so, we could avoid showing an error message since the Continue button would be un-tappable. 

The backend returns `two_step_notification_sent` and should populate it with `sms` if an SMS is sent. This seemed promising but when my test account has SMS-based 2FA enabled, the backend will still send a push notification (I still got it after logging out of my account on other devices). Did you see this also?

So `two_step_notification_sent` returns `push` even though I have SMS-based 2FA enabled. 

I also looked at what happens if I tap the button to send an SMS. `two_step_notification_sent` isn't returned (I'm not sure if this is intentional) and instead we just get:
```
▿ 2 elements
  ▿ 0 : 2 elements
    - key : "two_step_nonce"
    - value : xyz
  ▿ 1 : 2 elements
    - key : "phone_number"
    - value : 12
 ```

So it seems that if `two_step_notification_sent` returned the 2FA type, we could address this issue differently by not allowing the form to be submitted if the user enters 6 characters and has SMS-based 2FA enabled. I'm not sure if it doesn't always return the expected value on purpose (e.g. due to security reasons) or if it's a bug.

Either way, the proposed solution here makes sense and I tested it and it works 👍 Thanks for making the fix, @staskus!

staskus:
> Is there a way to know what length is expected and disable the Continue button if the entered code length doesn't match the expected code length? If so, we could avoid showing an error message since the Continue button would be un-tappable.

@guarani, this is an excellent idea. I also thought about it, but as you noticed, I couldn't determine if we can consistently know which source we're expecting, so I left the logic itself unchanged. 

>  I'm not sure if it doesn't always return the expected value on purpose (e.g. due to security reasons) or if it's a bug.

Yes, this is a great question 🤔 I will time-box a bit more time to gain more understanding.

guarani:
👋 Hi @staskus, this change caused a regression with passkey support and it was reverted in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/802. I haven't investigated an alternative solution to fix this issue without affecting passkeys.

It's worth noting that @enejb and I looked at replacing the login endpoint in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/646. This allows the app to get an accurate list of supported 2FA methods and might give us the info we need to approach this issue differently (e.g. we'll know what length to expect in the 2FA text field and be able to avoid the error message entirely).

staskus:
> 👋 Hi @staskus, this change caused a regression with passkey support and it was reverted in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/802. I haven't investigated an alternative solution to fix this issue without affecting passkeys.

@guarani thanks for fixing it 😓  I couldn't have imagined it would break passkey support. My assumption here was wrong:

> There's no point attempting to log in with an empty nonce since the user would receive a non-descriptive "two_step_nonce required" message.

guarani:
I reviewed this and didn't catch it either. It crossed my mind that this could be related on Nov 2 (p1698934406482319/1698884760.893009-slack-C0180B5PRJ4) but I didn't follow up on this particular lead until a week had passed.

Yeah, the empty nonce was it.

PR #794, opened 2023-10-24
Fix two 2FA controllers being opened when logging in at the same time

staskus:
Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/21864
Related PR: 

Fix two 2fa controllers being opened at the same time
- [self.delegate needsMultifactorCode] is executed when nonceInfo == nil
- [self.delegate needsMultifactorCodeForUserID:userID andNonceInfo:nonceInfo] is executed always

```
        if (nonceInfo == nil && [self.delegate respondsToSelector:@selector(needsMultifactorCode)]) {
            [self.delegate needsMultifactorCode]; // If nonceInfo == nil, executed
        }

        if ([self.delegate respondsToSelector:@selector(needsMultifactorCodeForUserID:andNonceInfo:)]) {
            [self.delegate needsMultifactorCodeForUserID:userID andNonceInfo:nonceInfo];  // If nonceInfo == nil, executed as well
        }
```

We only need these paths to be executed once.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ _None._
 ### Bug Fixes
 
 - Fix an issue where `guessXMLRPCURL` was called with an URL without a scheme resulting in an error. [#792]
+- Fix an issue that leads to an ambiguous error message when an incorrect SMS 2FA code is submitted. [#793]
 - Fix an issue where two 2FA controllers were being opened at the same time when logging in. [#794]
 
 ## 7.2.0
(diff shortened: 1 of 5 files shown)


guarani (approved):
Tested 2FA login using a non-passkey account and confirmed that the 2FA screen is now shown only once. 

Would it be good to loop in WC devs in case this fix needs to be shipped in their next release?

staskus:
> Would it be good to loop in WC devs in case this fix needs to be shipped in their next release?

Good idea, I'll loop them in.

PR #795, opened 2023-10-25
Fix issue checking site existence for a number of sites

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/11005

**Description**
As discussed in peaMlT-d1-p2, some users reported that their sites cannot pass the site check on the site address login screen on iOS, but it works for them on Android.

The code on the Android app is using GET to check the site info whereas we're using HEAD for iOS. I could not find any documentation about why we decided to use HEAD, but since GET helps unstuck users and make us align with Android, this PR updates the check to use GET for the check.

**Testing**

Please follow the steps in https://github.com/woocommerce/woocommerce-ios/pull/11006.

---


staskus (approved):
✅ Tested with https://github.com/woocommerce/woocommerce-ios/pull/11006

PR #796, opened 2023-10-27
Confirm that site is indeed not a WordPress site after XML-RPC validation fails

staskus:
More context: https://github.com/wordpress-mobile/WordPress-iOS/issues/18497#issuecomment-1779361108
Related PR https://github.com/wordpress-mobile/WordPress-iOS/pull/21904

In rare cases, XML-RPC validation can fail and indicate that the site is not a WordPress site when it actually is. 

To avoid this corner case, I added an additional `isWPSite` check after `guessXMLRPCURL` and before showing errors to the users:
- If WordPressOrgXMLRPCValidator determines that it's not a valid WordPress site and `isWPSite` (`connect/site-info`) API call confirms it, then we inform the user that the site is invalid
- If WordPressOrgXMLRPCValidator determines that it's not a valid WordPress site but `isWPSite` (`connect/site-info`) API suggests otherwise, we inform users about XML-RPC issues which they can then look for further assistance

### Implementation details

I didn't want to change this fragile code before writing tests to it and to write tests I needed to split business logic from UI. I recommend reviewing commit by commit:

1. Extract error handling code - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/796/commits/9c14098b737397095420c4bd68cc39ebe1ce8778
2. Move it to ViewModel - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/796/commits/b62db739cd3f6d6b20e50947ccc6821c8309e32b
3. Add unit tests - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/796/commits/76925ba83e7a6712b7412fc7d849bbb6de457ab8
4. Add isWPSite check - https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/796/commits/a27ac06167677a7303a4d84d633c0daa6b82fb45


staskus:
@guarani the code can be reviewed but ideally, I would do some more manual testing myself, so I'm not targeting 23.6 for sure. If you'll have time to review the code or critique the solution overall - great, I'll finish after I come back. Thanks! 🙇

SiobhyB (approved):
@staskus, thank you for working on this! 🙇‍♀️ I verified that the misleading "not a WordPress site" error is resolved with these changes applied:

| Before  | After |
| ------------- | ------------- |
|  |  |

In addition, I went through the steps at https://github.com/wordpress-mobile/WordPress-iOS/pull/17724 and tried a few ways to see if I could break login, but couldn't. 

The only note I have is that, although less confusing than before, I feel some users won't know what next steps to take or perhaps even what XML-RPC is. How would you feel about adding a sentence like `Please check with your hosting provider to find out if they are blocking XML-RPC access`? Or perhaps even linking to [our support doc](P7jrAc-1tl-p2)? I understand that this might be out of the scope of this PR though. Let me know if I should open a separate issue regarding the wording!

I went through each commit as well as your descriptions, and the logic seems sound to me. That said, I'm not an iOS developer, so it might be wise to have @guarani or another iOS developer do another pass.

Approving based on my testing and understanding of the logic. :)

WordPressAuthenticator/Unified Auth/View Related/Site Address/SiteAddressViewModel.swift
```diff
@@ -0,0 +1,136 @@
...
+
+                loading(false)
+
+                guard self.isSiteDiscovery == false else {
+                    completion(.troubleshootSite)
+                    return
+                }
+
+                let err = self.originalErrorOrError(error: error as NSError)
+                self.handleGuessXMLRPCURLError(error: err, loading: loading, completion: completion)
```
guarani: nitpick: I think this should be outdented. It would be nice if something like SwiftLint could catch this earlier.

guarani (approved):
Approved via https://github.com/wordpress-mobile/WordPress-iOS/pull/17724. 

The code change looks good and I ran this repos tests locally.

PR #801, opened 2023-11-09
Passkeys: Don't show error if users cancels Passkey authentication

momo-ozawa:
## Description

Fixes an issue where an error was displayed when you exit out of Passkey authentication

### Before

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/6711616/c76c9fd1-311e-4d53-bee4-7c1eb71b0e75

### After

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/6711616/951a4b74-04a7-4163-8405-5ec95c508998

## How to test
Test via https://github.com/wordpress-mobile/WordPress-iOS/pull/22001

---

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -28,7 +28,7 @@ def wordpress_authenticator_pods
   ## These should match the version requirement from the podspec.
   pod 'Gridicons', '~> 1.0'
   pod 'WordPressUI', '~> 1.7-beta'
-  pod 'WordPressKit', '~> 8.7-beta'
+  pod 'WordPressKit', '~> 9.0-beta'
   pod 'WordPressShared', '~> 2.1-beta'
 
   third_party_pods
--- a/WordPressAuthenticator.podspec
+++ b/WordPressAuthenticator.podspec
@@ -40,6 +40,6 @@ Pod::Spec.new do |s|
   # Use a loose restriction that allows both production and beta versions, up to the next major version.
   # If you want to update which of these is used, specify it in the host app.
   s.dependency 'WordPressUI', '~> 1.7-beta'
-  s.dependency 'WordPressKit', '~> 8.7-beta'
+  s.dependency 'WordPressKit', '~> 9.0-beta'
   s.dependency 'WordPressShared', '~> 2.1-beta'
 end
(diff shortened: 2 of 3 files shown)


momo-ozawa:
Hey @Ecarrion 👋 
I don't think we should show an error when canceling the Passkey authentication flow. Wdyt?

guarani (approved):
After canceling, if I try the passkey option again it fails with the message "Your session has expired, please go back to the login screen.". Is this expected? It's not a great user experience but the message is helpful & actionable, so I don't think it's a blocker.

guarani:
I updated WordPressKit to point to the next major version since this library and WPiOS both depend on it. I hope that's ok 🙇

Ecarrion:
Hi @momo-ozawa @frosty,

I initially decided to present the error flow when canceling for simplicity, because we can't cancel the flow and start again the passkeys process without regenerating the server nonces.

The "ideal" solution would be to have some state on the client side to track if the error was due to a cancelation to not start the passkeys flow from scratch but only from the signature. But at the time I didn't make that a priority 😅 

What do you think about changing the error message(but we still reset the flow/stack upon tapping "ok") to explain a little bit better what is happening. This would prevent the user to end up on a faulty state and having to tap on CTAs twice to still end up with error screens.

momo-ozawa:
> I initially decided to present the error flow when canceling for simplicity, because we can't cancel the flow and start again the passkeys process without regenerating the server nonces.

@Ecarrion Ah I see! So showing an error when you cancel is expected at this point in time.

> What do you think about changing the error message(but we still reset the flow/stack upon tapping "ok") to explain a little bit better what is happening. This would prevent the user to end up on a faulty state and having to tap on CTAs twice to still end up with error screens.

Given that the current flow is intentional, and our priority is to merge https://github.com/wordpress-mobile/WordPress-iOS/pull/22001 and https://github.com/wordpress-mobile/WordPressKit-iOS/pull/646 into WPiOS 23.6, I think we can leave out this change for now and address it in a future release. Wdyt, @guarani? 

We can change the copy as Ernesto suggested, or consider a more robust solution that restarts the passkeys flow from the signature.

momo-ozawa:
I'm closing this for now - we can revisit this in a different PR

PR #807, opened 2023-11-27
Avoid casting to NSError when handling HTTP API errors

crazytonyli:
We want to preserve the type information of the original error, instead of making them all `NSError` type. 

This code change doesn't have any real runtime impact at the moment, because the errors are still converted to `NSError`. But later, I'll create PRs to refactor error handling, and at that point the `alertForError` function will be dealing with concrete error types, like:

```swift
func alertForError(error: Error, ...) {
  if let oauthError = error as? OAuthError {
    // ...
  }

  if let xmlrpcError = error as? XMLRPCError {
    // ...
  }

  // ...
}

```

---

### DIFF
--- a/WordPressAuthenticator/NUX/NUXViewControllerBase.swift
+++ b/WordPressAuthenticator/NUX/NUXViewControllerBase.swift
@@ -71,9 +71,9 @@ extension NUXViewControllerBase where Self: UIViewController, Self: UIViewContro
 
     /// Displays a login error in an attractive dialog
     ///
-    func displayError(_ error: NSError, sourceTag: WordPressSupportSourceTag) {
+    func displayError(_ error: Error, sourceTag: WordPressSupportSourceTag) {
         let presentingController = navigationController ?? self
-        let controller = FancyAlertViewController.alertForError(error as NSError, loginFields: loginFields, sourceTag: sourceTag)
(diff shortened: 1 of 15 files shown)


WordPressAuthenticator/NUX/NUXViewControllerBase.swift
```diff
@@ -71,9 +71,9 @@ extension NUXViewControllerBase where Self: UIViewController, Self: UIViewContro
 
     /// Displays a login error in an attractive dialog
     ///
-    func displayError(_ error: NSError, sourceTag: WordPressSupportSourceTag) {
+    func displayError(_ error: Error, sourceTag: WordPressSupportSourceTag) {
```
mokagio, line 74: Inspired by one of your PRs that I reviewed earlier where `authenticateWithUsername(` became `authenticate(username:`, what do you think of making this 

```suggestion
    func display(error: Error, sourceTag: WordPressSupportSourceTag) {
```

Or even simply

```suggestion
    func display(_ error: Error, sourceTag: WordPressSupportSourceTag) {
```

mokagio (approved):

WordPressAuthenticator/NUX/NUXViewControllerBase.swift
crazytonyli: 👍 Sounds good! There are more changes to come. I'll leave them as they are for now, and change them in my next PRs.

PR #809, opened 2023-11-28
Add an option to disable Passkeys

kean:
This PR adds a temporary option to disable passkeys.

To test: see https://github.com/wordpress-mobile/WordPress-iOS/pull/22113. 

---

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #810, opened 2023-11-29
Add unit tests for how the rows in the TwoFAVC are computed

mokagio:
Builds on top #809 . While looking at that PR, I noticed that there are three essentially boolean parameters that drive the rows behavior (either actual `Bool` or `Optional` `.some` vs `.none`). That results into eight _possible_ runtime input configurations. Too many to test manually, but easily testable via unit tests.

In truth, there's an additional boolean parameter, `#available(iOS 16)` but I decided to ignore that under the assumption that it's most of the user base.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. — N.A.

### DIFF
--- a/WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
+++ b/WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
@@ -7,19 +7,19 @@ import XCTest
 class TwoFAViewControllerTests: XCTestCase {
 
     func testLoadRowsWithNoErrorAndNoPasskeys() {
-        let defualtRows: [TwoFAViewController.Row] = [.instructions, .code, .alternateInstructions, .sendCode]
+        let defaultRows: [TwoFAViewController.Row] = [.instructions, .code, .alternateInstructions, .sendCode]
 
         XCTAssertEqual(
             TwoFAViewController.computeRows(errorMessage: nil, nonceWebauthn: nil, passkeysEnabled: false),
-            defualtRows
+            defaultRows
         )
         XCTAssertEqual(
             TwoFAViewController.computeRows(errorMessage: nil, nonceWebauthn: nil, passkeysEnabled: true),
-            defualtRows
+            defaultRows
         )
         XCTAssertEqual(
             TwoFAViewController.computeRows(errorMessage: nil, nonceWebauthn: "some error", passkeysEnabled: false),
-            defualtRows
+            defaultRows
         )
     }
 
--- a/WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
+++ b/WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
@@ -1,0 +1,56 @@
+@testable import WordPressAuthenticator
+import XCTest
+
+// The logic depends, in part, on iOS 16 features.
+// We could test for the earlier versions but that would optimizing for a small minority and not worth at this point in time.
+@available(iOS 16, *)
+class TwoFAViewControllerTests: XCTestCase {
+
+    func testLoadRowsWithNoErrorAndNoPasskeys() {
+        let defualtRows: [TwoFAViewController.Row] = [.instructions, .code, .alternateInstructions, .sendCode]
+
+        XCTAssertEqual(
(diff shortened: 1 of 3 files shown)


WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        )
+    }
+
+    func testLoadRowsWithErrorAndPasskeys() {
+        XCTAssertEqual(
+            TwoFAViewController.computeRows(errorMessage: "some error", nonceWebauthn: "some nonce", passkeysEnabled: true),
+            [.instructions, .code, .errorMessage, .alternateInstructions, .sendCode, .enterSecurityKey]
+        )
+    }
+}
```
mokagio, line 56: Note for the future. `nonceWebauthn` and `passkeysEnabled` are always tied together, so we could simplify this code by adding a middle layer.

Something like

```swift
enum ShouldUsePasskeys {
  case false
  case true(nonce: String)
}
```

We would then have tests for

- Computing a `ShouldUsePasskeys` given `nonceWebauthn` and `passkeysEnabled` (four possibilties)
- Computing rows given `String?` error messages and `ShouldUsePasskeys` (four possibilities)

The number of possibilities would be the same as now, but each step would be individually easier to understand

mokagio: I acknowledge that `ShouldUsePasskeys` looks a lot like `Optional`, but I find it's clearer in expressing its meaning.

WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
```diff
@@ -0,0 +1,56 @@
+@testable import WordPressAuthenticator
+import XCTest
+
+// The logic depends, in part, on iOS 16 features.
+// We could test for the earlier versions but that would optimizing for a small minority and not worth at this point in time.
+@available(iOS 16, *)
+class TwoFAViewControllerTests: XCTestCase {
+
+    func testLoadRowsWithNoErrorAndNoPasskeys() {
+        let defualtRows: [TwoFAViewController.Row] = [.instructions, .code, .alternateInstructions, .sendCode]
```
jkmassel:
```suggestion
        let defaultRows: [TwoFAViewController.Row] = [.instructions, .code, .alternateInstructions, .sendCode]
```

etc, below

jkmassel (changes requested):

WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        XCTAssertEqual(
+            TwoFAViewController.computeRows(errorMessage: nil, nonceWebauthn: nil, passkeysEnabled: false),
+            defualtRows
```
mokagio:
```suggestion
            defaultRows
```

WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        )
+        XCTAssertEqual(
+            TwoFAViewController.computeRows(errorMessage: nil, nonceWebauthn: nil, passkeysEnabled: true),
+            defualtRows
```
mokagio:
```suggestion
            defaultRows
```

WordPressAuthenticatorTests/TwoFAViewControllerTests.swift
```diff
@@ -0,0 +1,56 @@
...
+        )
+        XCTAssertEqual(
+            TwoFAViewController.computeRows(errorMessage: nil, nonceWebauthn: "some error", passkeysEnabled: false),
+            defualtRows
```
mokagio:
```suggestion
            defaultRows
```

jkmassel (approved):
Thanks for indulging my silly typo fix!

mokagio:
> Thanks for indulging my silly typo fix!

Not at all silly! As you know, _tpyos_ matter here. Ref pb6Nl-dHI-p2

PR #816, opened 2023-12-19
Fix an issue where when passkey flow is cancelled an error is shown

kean:
---


Ecarrion (commented):
Hi @kean,

I'm not sure if this is ready for review(I don't see any PR description) but I got notified as a reviewer.

I don't think we can merge these changes as they are right now. If we don't `exit` the flow when the user `cancels` we get new errors if the user taps "security keys" again as nonces used in the network requests can only be used one time.

```
Failed to request webauthn challenge endpointError(WordPressKit.AuthenticationFailure(kind: WordPressKit.AuthenticationFailure.Kind.unknown, localizedErrorMessage: Optional("Your session has expired, please go back to the login screen."), newNonce: nil, originalErrorJSON: ["data": {
    errors =     (
                {
            code = "invalid_two_step_nonce";
            message = "Your session has expired, please go back to the login screen.";
        }
    );
}, "success": 0]))
```

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/562080/af6a6de3-2a6f-4a66-9c4d-43c0b227ed76


----

If we want to improve this experience I think we need to handle more state for refreshing nonces or knowing when a nonce is still valid.

kean:
Nice catch. Yeah, makes sense.

As a user, I would expect that when I tap "Cancel", I would be returned to the 2FA screen so that I could try a different 2FA option.

PR #818, opened 2023-12-22
Add Accessbility identifier to passkeys button

Ecarrion:
As the title says, this PR adds an accessibility identifier to the "Use security key button". This is needed to enhance the WooCommerce iOS App UI Tests. 

This changed is used here: 
- https://github.com/woocommerce/woocommerce-ios/pull/11538


jostnes (approved):
Thanks for adding accessibility ID to the new element, making the UI more testable! Validated via the WCiOS PR and the `accessibilityIdentifier` works as expected:

image

PR #820, opened 2023-12-28
Fixes issue where "Use Security Key" option should not be present

### LINKED ISSUE #817: The Security Key option is displayed for an account that does not have a Security Key setup
**Setup:**

- Account A has a security key set up
- Account B has two-factor authentication enabled, but no security key

**Steps:**

- Log in with account A (using password autofill)
- Tap "Back" on the second factor screen
- Log in with account B

**Expected Result**

The second factor screen does **not** show "Use a security key" option.

**Actual Result**

- The second factor screen shows a "Use a security key" option.
- The `nonce` value and other parameters are the same as after the login with the account A

Ecarrion:
closes #817 

# Why

This PR fixes an issue where the Security Key option is displayed for an account that does not have a Security Key setup.

This happened because there were some nonce state that weren't being cleared properly when logging in with a new account.

# How

- Clean the nonce state when the 2Factor code screen is presented.
- For extra safety, improve the check before rendering the security key option.

# Demo

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/562080/69de42dd-437f-4149-949d-2e8668f81151

# Testing Steps

You can follow the steps on this PR 

- https://github.com/woocommerce/woocommerce-ios/pull/11579

### DIFF
--- a/WordPressAuthenticator/Signin/LoginViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginViewController.swift
@@ -216,6 +216,10 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
             return
         }
 
+        // Make sure we don't provide any old nonce information when we are required to present only the multi-factor code option.
+        loginFields.nonceInfo = nil
+        loginFields.nonceUserID = 0
+
         presentUnified2FA()
     }
 
--- a/WordPressAuthenticator/Unified Auth/View Related/2FA/TwoFAViewController.swift
+++ b/WordPressAuthenticator/Unified Auth/View Related/2FA/TwoFAViewController.swift
@@ -491,7 +491,7 @@ private extension TwoFAViewController {
         rows.append(.spacer(4))
         rows.append(.sendCode)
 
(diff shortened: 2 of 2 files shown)


kean:
I tested the change in WP-iOS https://github.com/wordpress-mobile/WordPress-iOS/pull/22309 and the issue was fixed.

kean (approved):

WordPressAuthenticator/Signin/LoginViewController.swift
```diff
@@ -216,6 +216,10 @@ open class LoginViewController: NUXViewController, LoginFacadeDelegate {
             return
         }
 
+        // Make sure we don't provide any old nonce information when we are required to present only the multi-factor code option.
+        loginFields.nonceInfo = nil
+        loginFields.nonceUserID = 0
```
kean, line 221: (Nit) It doesn't seem like the optimal place to clear the state. It should probably be cleared right after you are returned to the login screen after a failed attempt.

PR #822, opened 2024-01-04
Bump version to 8.0.1

alpavanoglu:
---

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -9,7 +9,7 @@ PODS:
   - SVProgressHUD (2.2.5)
   - SwiftLint (0.54.0)
   - UIDeviceIdentifier (2.3.0)
-  - WordPressAuthenticator (8.0.0):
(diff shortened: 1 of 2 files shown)


salimbraksa:
@alpavanoglu My understanding is that this branch is unnecessary. The release manager will take care of updating the version. The process that I'm familiar with works like this:
- Merge pod PRs into `trunk`.
- In WordPress-iOS, change the pod branch to reference `trunk`. 
- Then, the release manager will take care of updating updating pod references during the code freeze. 
cc @mokagio

mokagio:
Thanks @salimbraksa @alpavanoglu.

The only thing to notice is that

> In WordPress-iOS, change the pod branch to reference trunk.

Should be 

> In WordPress-iOS, change the pod branch to reference **the latest commit in your feature branch, or latest on `trunk` after merging**.

The rationale is that a commit is a precise snapshot in time while a branch reference might end up changing over time or even disappear as some point.

PR #824, opened 2024-01-09
Relax WordPressKit dependency from `~> 9.0.0` to `~> 9.0`

mokagio:
A dependency on `~> 9.0.0` results in only versions `9.0.x` being compatible. Under the assumption WordPressKit always respects semantic versioning, that approach limits the versions WordPressAuthenticator can be compatible with. In particular, prevents seamlessly using something like WordPressKit 9.1.0 in a client of both WordPressAuthenticator and WordPressKit.

Case in point, I just created WordPressKit 9.0.3,
https://github.com/wordpress-mobile/WordPressKit-iOS/pull/677, in place of 9.1.0 in order to avoid having to ship a new WordPressAuthenticator version in the middle of a release cycle.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-_None._
+- Relax WordPressKit dependency to `~> 9.0` [#824]
 
 ## 8.0.1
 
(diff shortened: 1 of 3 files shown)


crazytonyli (approved):
Good call! It should have been `9.0` in the first place.

PR #830, opened 2024-01-29
Implement needsMultifactorCode with userId and nonceinfo delegate in StoredCredentialsAuthenticator

staskus:
Fix login being stuck in a loading state when the account has 2FA with passkeys enabled.

Fix implements `needsMultifactorCode(forUserID userID: Int, andNonceInfo nonceInfo: SocialLogin2FANonceInfo)` delegate which is called when `nonceInfo != nil` which presents 2FA login.

Context: p1706533264603049-slack-C012H19SZQ8

---

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/4062343/9874afb2-188d-424d-b656-da03b9f6b9d2

### DIFF
--- a/WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
+++ b/WordPressAuthenticator/Unified Auth/StoredCredentialsAuthenticator.swift
@@ -155,6 +155,13 @@ extension StoredCredentialsAuthenticator: LoginFacadeDelegate {
         presentTwoFactorAuthenticationView()
     }
 
+    func needsMultifactorCode(forUserID userID: Int, andNonceInfo nonceInfo: SocialLogin2FANonceInfo) {
+        loginFields?.nonceInfo = nonceInfo
+        loginFields?.nonceUserID = userID
+
+        needsMultifactorCode()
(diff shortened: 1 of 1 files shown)


staskus:
@Ecarrion 👋 Letting you know. We've noticed on the JP app that login from the home screen doesn't work if the account has 2FA with passkey enabled.

guarani (commented):
Tested and working well. I tested the JP app:
- WP.com login with OTP in 1Password
- WP.com login with passkey in 1Password
- WP.com login with SMS
- WP.com login with no 2FA at all

 I couldn't find another instance where `LoginFacadeDelegate` is conformed to, but `needsMultifactorCode` is not implemented.

I'd love to have @Ecarrion's review as well to make sure this change is safe for WooMobile.

guarani (approved):

Ecarrion (approved):
This flow does not seem to be used on the woo app, but still tested all other login flows and they work correctly! 

Thanks for letting us know!

staskus:
Thank you for reviews, @guarani @Ecarrion!

PR #836, opened 2024-02-21
Add `useURLSession` configuration to pass to WordPressKit

mokagio:
This will allow clients that do no need access to WordPressKit (i.e. WooCommerce) to explicitly depend only on WordPressAuthenticator and still configure whether to use `URLSession`.

See it in use in https://github.com/woocommerce/woocommerce-ios/pull/12063

For context: I discovered this possibility while looking around the apps and the libraries searching for ways to redistribute concerns. I wanted to know how many of the WordPressKit APIs Woo used, go I grepped `import WordPressKit`. I found a single usage, which this PR plus https://github.com/woocommerce/woocommerce-ios/pull/12063 remove.

---

### DIFF
--- a/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
+++ b/WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
@@ -174,6 +174,11 @@ public struct WordPressAuthenticatorConfiguration {
     ///
     let enableSiteAddressLoginOnlyInPrologue: Bool
 
+    /// Whether to use `URLSession` or Alamofire for certain requests.
+    ///
+    /// Defaults to `false` but will soon become `true` as we are actively working to remove Alamofire.
+    let useURLSession: Bool
+
     /// Designated Initializer
     ///
     public init (wpcomClientId: String,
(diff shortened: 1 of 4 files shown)


WordPressAuthenticator/Authenticator/WordPressAuthenticatorConfiguration.swift
```diff
@@ -210,7 +215,8 @@ public struct WordPressAuthenticatorConfiguration {
                  enableManualSiteCredentialLogin: Bool = false,
                  enableManualErrorHandlingForSiteCredentialLogin: Bool = false,
                  useEnterEmailAddressAsStepValueForGetStartedVC: Bool = false,
-                 enableSiteAddressLoginOnlyInPrologue: Bool = false
+                 enableSiteAddressLoginOnlyInPrologue: Bool = false,
+                 useURLSession: Bool = false
```
crazytonyli, line 219: The `useURLSession` configuration is a global configuration, using it in an initialiser _feels_ weird to me...

Also, we'll soon (maybe in a couple of weeks) deprecate `useURLSession` and remove Alamofire from WordPressKit. Maybe not worth adding it to another framework?

mokagio: Fair enough. I'm okay to wait, in particular because the `useURLSession` deprecation should be very near: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/736

PR #841, opened 2024-03-08
Detect invalid WP site and return error

itsmeichigo:
**Description**
Previously, we are ignoring when a user enters an incorrect WP address. The URL after redirect would then be `https://wordpress.com/blog`, and they would be directed to the next screen for WPCom credential input.

This PR fixes this by checking for the redirected URL and returning an error when it's `https://wordpress.com/blog`.

**Testing steps**
- Run the demo app.
- Select Show Login > Enter your existing site address.
- Enter an invalid WP site that ends with "wpcomstaging.com" or "wordpress.com".
- Confirm that an alert is displayed and you cannot reach the WPCom login flow.

**Screencast**

https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/assets/5533851/28e8bea5-33ff-46b6-9887-4d2b90c853bb

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
+- Detect invalid WordPress site address. [#841]
 
 ### Internal Changes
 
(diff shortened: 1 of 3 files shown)


selanthiraiyan (approved):
LGTM :shipit:

PR #851, opened 2024-04-12
Release 9.0.7

itsmeichigo:
Releasing a new version containing a fix for the site credential login to show social login options.
I updated the CHANGELOG to make this fix a bug fix instead of breaking, because it doesn't make the hosting apps fails to compile.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-- Show social login options in site credentials login mode [#849]
+_None._
 
 ### New Features
 
@@ -47,6 +47,12 @@ _None._
 ### Internal Changes
 
 _None._
+
+## 9.0.7
+
+### Bug Fixes
+
+- Show social login options in site credentials login mode [#849]
 
 ## 9.0.6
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ _None._
 
 _None._
 
+## 9.0.7
+
+### Bug Fixes
```
selanthiraiyan, line 53: I think this title should be "Breaking Changes". 
```suggestion
### Breaking Changes
```

selanthiraiyan (approved):
🚀

CHANGELOG.md
itsmeichigo: I explained in the description: 

> I updated the CHANGELOG to make this fix a bug fix instead of breaking, because it doesn't make the hosting apps fails to compile.

If we want to keep it a breaking change, we would need to bump the version to a major number (10.0). I don't think this break the hosting apps, so I updated it.

selanthiraiyan: > I explained in the description:

Sorry, I missed this.

> I don't think this break the hosting apps, so I updated it.

This makes sense. 👍

PR #855, opened 2024-08-15
Hide site credential login button from Get Started screen after entering a WPCom site address

itsmeichigo:
Part of https://github.com/woocommerce/woocommerce-ios/issues/13622

**Description**
On WCiOS we support site credential login for self-hosted sites with JP connection. In a previous attempt to add the site credential login option, we accidentally enabled that option for WPCom sites. This option is not functional as it's impossible to log in to WPCom sites with site credentials, so this PR fixes this by hiding that option for WPCom sites.

**Testing steps**
Test the demo app to ensure that the credential login button is not available in the default case.
Additionally, follow the steps in https://github.com/woocommerce/woocommerce-ios/pull/13623 to check the cases for WCiOS.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,11 +42,17 @@ _None._
 
 ### Bug Fixes
 
-- Hide site credential login option from Get Started screen after entering a WPCom site address [#855]
+_None._
 
 ### Internal Changes
 
 _None._
+
+## 9.0.9
+
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


hafizrahman (approved):
Looks good to me 👍🏼

itsmeichigo:
Thanks Hafiz for the review! I've added 2 comments to bump the library version and will release the new version after this PR is merged.

PR #856, opened 2024-09-10
Fix compiler issues for Xcode 16 Support

staskus:
This is a fix for compiler errors for Xcode 16 support.

I used WordPress-iOS as an example https://github.com/wordpress-mobile/WordPress-iOS/pull/23497. They have already made this fix since they copied these dependencies on the repository.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-_None._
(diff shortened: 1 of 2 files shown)


itsmeichigo (approved):
As tested in https://github.com/woocommerce/woocommerce-ios/pull/13911, this compiles on both Xcode 15.3 and 16.1 Beta. LGTM :shipit:

PR #858, opened 2024-10-30
Allow login for WPCOM suspended sites

selanthiraiyan:
Part of https://github.com/woocommerce/woocommerce-ios/issues/13348

**Description**
When users attempt to log into the app with a site suspended by WPCOM, they receive an error message stating the URL is invalid and cannot proceed with login.

This PR adds a new configuration and updates the login flow to handle this case by directing those users to use site credentials instead of WordPress.com.

**Testing steps**

Follow the testing steps at https://github.com/woocommerce/woocommerce-ios/pull/14257 to check the cases for WCiOS.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Add support for logging in into WPCOM suspended sites.
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


hichamboushaba (approved):
Nice work @selanthiraiyan, tested along the WCiOS PR, and the changes work as expected 👏

PR #861, opened 2024-11-14
Add support for requesting account creation when sending auth magic link

hichamboushaba:
⚠️ Please don't merge until the Podfile is updated with a trunk version of WordPressKit.

Part of: https://github.com/woocommerce/woocommerce-ios/issues/14419

This PR adds support for requesting the creation of the account when using `WordPressComAccountService/requestAuthenticationLink`, this allows using the same WPCom endpoint for sending magic link to both existing and new accounts.

### Testing
Check https://github.com/woocommerce/woocommerce-ios/pull/14431

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #862, opened 2024-11-20
[DO NOT MERGE] Magic link create account test

twstokes:
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ _None._
 ### New Features
 
 - Add support for logging in into WPCOM suspended sites.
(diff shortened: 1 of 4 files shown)


mokagio:
Closing to keep the PRs list tidy. See last commit in #861 . TL;DR updating `Podfile` was good for _development_. But in order to have a ready-to-publish pod (which is what the validate `.podspec` CI step checks) we also needed to update the `podspec`. The `podspec` has to point to a published version, see https://guides.cocoapods.org/syntax/podspec.html#dependency, hence we also had to publish a new WordPressKit version https://github.com/wordpress-mobile/WordPressKit-iOS/pull/823

Finally, this will all become redundant as soon as we implement https://github.com/woocommerce/woocommerce-ios/issues/14329

PR #867, opened 2024-12-27
[Woo Brand Update] Add more customization options to the prologue

hichamboushaba:
Needed for woocommerce/woomobile-private#385

This PR adds the following changes:

- Removes an outdated logic of blur that causes blur to be shown even when not configured (more details in the comments section).
- Update logic to set both the background color and image at the same time.
- Allow setting a scaling mode for the background image.
- Allow customizing the style of the tertiary button in the prologue.
- Fix an issue with the screen sizing (more details in the comments section).

#### To be tested with https://github.com/woocommerce/woocommerce-ios/pull/14773

---

### DIFF
--- a/WordPressAuthenticator/NUX/Button/NUXStackedButtonsViewController.swift
+++ b/WordPressAuthenticator/NUX/Button/NUXStackedButtonsViewController.swift
@@ -150,13 +150,18 @@ private extension NUXStackedButtonsViewController {
         buttons = []
         topStackView?.arrangedSubviews.forEach({ $0.removeFromSuperview() })
         bottomStackView?.arrangedSubviews.forEach({ $0.removeFromSuperview() })
+        topStackView?.isHidden = true
+        bottomStackView?.isHidden = true
+
         for config in buttonConfigs {
             let button = NUXButton()
             switch config.stackView {
             case .top:
                 topStackView?.addArrangedSubview(button)
+                topStackView?.isHidden = false
             case .bottom:
                 bottomStackView?.addArrangedSubview(button)
+                bottomStackView?.isHidden = false
             }
             button.configure(withConfig: config.config, and: config.style)
             buttons.append(button)
--- a/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
+++ b/WordPressAuthenticator/Signin/LoginPrologueViewController.swift
@@ -66,12 +66,10 @@ class LoginPrologueViewController: LoginViewController {
             topContainerView.pinSubviewToAllEdges(topContainerChildViewController.view)
         }
 
+        setupViewBackground()
         createButtonViewController()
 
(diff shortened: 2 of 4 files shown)


WordPressAuthenticator/NUX/Button/NUXStackedButtonsViewController.swift
```diff
@@ -150,13 +150,18 @@ private extension NUXStackedButtonsViewController {
         buttons = []
         topStackView?.arrangedSubviews.forEach({ $0.removeFromSuperview() })
         bottomStackView?.arrangedSubviews.forEach({ $0.removeFromSuperview() })
+        topStackView?.isHidden = true
+        bottomStackView?.isHidden = true
```
hichamboushaba, line 154: Before this, both stack views were shown even when they are empty, and for some reason this causes the button section to take more space than needed when the top layout doesn't dictate a specific sizing. 
I'm not entirely sure about the cause here, but my basic understanding is that when the stackview is empty and shown, it doesn't report an intrinsic size, so AutoLayout doesn't know how much space to give it when the top view doesn't report a size either.

Anyway, regardless of the explanation, I think hiding the empty view is what we should do, and given this fixes the issue, it's just better.

(I tried setting a higher priority for the content hugging constraint for the button container, but it didn't have an effect).

WordPressAuthenticator/Signin/LoginPrologueViewController.swift
```diff
@@ -386,26 +380,27 @@ class LoginPrologueViewController: LoginViewController {
...
 
-        /// If host apps provide a background color for the prologue buttons:
-        /// 1. Hide the blur effect
-        /// 2. Set the background color of the view controller to prologueViewBackgroundColor
-        let prologueViewBackgroundColor = WordPressAuthenticator.shared.unifiedStyle?.prologueViewBackgroundColor ?? .clear
-
-        guard prologueViewBackgroundColor.cgColor == buttonsBackgroundColor.cgColor else {
-            buttonBlurEffectView.effect = UIBlurEffect(style: blurEffect)
-            return
-        }
```
hichamboushaba, line 404: This logic is old, we had it when we didn't allow client apps to control the blur effect, later, this ability was [added](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/c0344754e9bcee40a8243976f80ead56a8de4c59). Also, this old logic was causing me an issue, since with the new design, we need to have the buttons background as `.clear`, which means the condition applies here, and it causes the blur to be shown even when it's not configured, removing the logic fixes the issue.

itsmeichigo (approved):
Tested as part of https://github.com/woocommerce/woocommerce-ios/pull/14773 and confirmed that the new changes work correctly in iPhone SE, iPhone 16 Plus and iPad Mini simulators ✅.

